PR #253, opened 2021-02-25
Update Directory.Build.props

Youssef1313:


twsouthwick (approved):
Thanks @Youssef1313 . Looks like this is needed for targeting anything less than .NET 5 (which I have a PR to move a library to .NET Standard).

PR #256, opened 2021-02-25
Minor cleanup for analyzers

Youssef1313:


src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/AllowHtmlAttributeAnalyzer.cs
```diff
@@ -22,15 +22,16 @@ public class AllowHtmlAttributeAnalyzer : DiagnosticAnalyzer
 
         public override void Initialize(AnalysisContext context)
         {
-            if (context is null)
-            {
-                throw new ArgumentNullException(nameof(context));
-            }
```
Youssef1313, line 28: Roslyn guarantees this is a non-null.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/AllowHtmlAttributeAnalyzer.cs
```diff
@@ -22,15 +22,16 @@ public class AllowHtmlAttributeAnalyzer : DiagnosticAnalyzer
...
-
             context.EnableConcurrentExecution();
             context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze);
-
-            context.RegisterSyntaxNodeAction(AnalyzeAttribute, SyntaxKind.Attribute);
+            context.RegisterCompilationStartAction(context =>
+            {
+                var allowHtmlType = context.Compilation.GetTypeByMetadataName(AllowHtmlAttributeName);
```
Youssef1313, line 29: From my understanding, the analyzer can only report a diagnostic if this type exists. So I'm checking that first before registering callbacks.

This should improve the performance of this analyzer by preventing unneeded callbacks to be made.

It's also worth trying to update the analyzer to not rely on syntax in a follow-up. Which is both more performant and more reliable.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/HtmlHelperAnalyzer.cs
```diff
@@ -41,14 +36,11 @@ public override void Initialize(AnalysisContext context)
...
-        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get { return ImmutableArray.Create(Rule); } }
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
 
         private void AnalyzeIdentifiers(SyntaxNodeAnalysisContext context)
         {
-            if (context.Node is not IdentifierNameSyntax identifier)
-            {
-                return;
-            }
+            var identifier = (IdentifierNameSyntax)context.Node;
```
Youssef1313, line 43: The cast will always succeed because RegisterSyntaxNodeAction is passed SyntaxKind.IdentifierName

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/AllowHtmlAttributeAnalyzer.cs
mjrousos: The Upgrade Assistant runs source updaters towards the end of the upgrade process. By the time this is run, the System.Web.* references will have already been removed so we probably won't be able to resolve the type to `System.Web.Mvc.AllowHtmlAttribute`. Instead, we have to look for syntax that probably refers to that type (because the simple name matches and it doesn't resolve to any other type).

mjrousos (changes requested):
Looks good. Only concern is that we should make sure the analyzers keep working at the syntax level in case symbols aren't available.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/AllowHtmlAttributeAnalyzer.cs
```diff
@@ -62,11 +55,15 @@ private void AnalyzeAttribute(SyntaxNodeAnalysisContext context, INamedTypeSymbo
 
             // If the identifier resolves to an actual symbol that isn't System.Web.Mvc.AllowHtmlAttribute, then bail out
             var attrNameSymbol = context.SemanticModel.GetTypeInfo(attribute.Name).Type;
-            if (SymbolEqualityComparer.Default.Equals(attrNameSymbol, allowHtmlType))
+            if (attrNameSymbol is INamedTypeSymbol symbol
+                && attrNameSymbol is not IErrorTypeSymbol
+                && !symbol.ToDisplayString(NullableFlowState.NotNull).Equals(AllowHtmlAttributeName, StringComparison.Ordinal))
```
Youssef1313: @mjrousos Thanks for noting this. I reverted my changes in this area to the original code. However, I'm a bit confused how will the attribute syntax be resolved to its fully qualified name at this point?

mjrousos: It's unlikely that we'll get an `INamedTypeSymbol` here. The scenarios in which that would happen are:

1. The `[AllowHtml]` attribute isn't actually referring to System.Web.Mvc.AllowHtmlAttribute but to some other unfortunately named attribute that we've mistaken for that one (maybe it's Foo.OtherAssembly.AllowHtmlAttribute). In that case, the symbol will likely resolve and we'll see that the attribute isn't the one we're looking for and not create a diagnostic.
2. Someone is using these analyzers and code fix providers on their own without running the upgrade assistant tool. That's perfectly fine and could result in us being able to resolve the AllowHtmlAttribute type symbol if the user is addressing that issue before they've addressed UA0001.

To try our best to cover all cases, we'll give a diagnostic for any use of [AllowHtml] where we can't resolve the syntax to a type symbol or where we can and see that it corresponds to System.Web.Mvc.AllowHtmlAttribute.

mjrousos (approved):

Issue #252, opened 2021-02-25
When i make use of Central Package Management, it cannot detect the package versions in the csproj.


ErwinBakels:
When i make use of Central Package Management, it cannot detect the package versions in the csproj.

![image](https://user-images.githubusercontent.com/25903888/109198862-0312c800-779f-11eb-85e7-461cb144d3ab.png)

twsouthwick:
Good catch! Are you using Directory.Packages.props to define the version?

ErwinBakels:
I use in Directory.Build.props the following

  
  <PropertyGroup Condition="'$(MSBuildProjectExtension)' == '.csproj'">
    <EnableCentralPackageVersions>true</EnableCentralPackageVersions>
    <EnablePackageVersionOverride>false</EnablePackageVersionOverride>
    <RestoreProjectStyle>PackageReference</RestoreProjectStyle>

<CentralPackagesFile>$(RootDir)PackageVersions.props</CentralPackagesFile>
  </PropertyGroup>


and in my PackageVersions.props

<Project>

  
  <ItemGroup>
    <PackageReference Update="Microsoft.NET.Test.Sdk" Version="16.7.0" />
    <PackageReference Update="Moq" Version="4.14.5" />
    <PackageReference Update="MSTest.TestAdapter" Version="2.1.2" />
    <PackageReference Update="MSTest.TestFramework" Version="2.1.2" />
    <PackageReference Update="SpecFlow" Version="3.0.220" />
    <PackageReference Update="SpecFlow.MsTest" Version="3.0.220" />
    <PackageReference Update="SpecFlow.Tools.MsBuild.Generation"
Version="3.0.220" />
  </ItemGroup>





Kind regards,

*Erwin Bakels*
Software Developer/Consultant


*Note: This email and its content is confidential and intended only for the
addressee. If you are not the addressee of this e-mail, please inform the
sender immediately by means of a reply e-mail and remove the received
e-mail from your systems. If you are not an addressee, you may not peruse,
copy, distribute, disclose or otherwise use the content.*


Op do 25 feb. 2021 om 19:54 schreef Taylor Southwick <
notifications@github.com>:

twsouthwick:
This will be a bit of work to enable, so we'll keep this open while also adding a preliminary check to notify people that this isn't supported at the moment.

sunandabalu:
- [ ] Read the props file for versions probably a medium workitem
 - [ ] Updating versions in the props file would be a large workitem

[...]

PR #279, opened 2021-02-26
Add a way to verify upgrade readiness

twsouthwick:
This adds an interface IUpgradeReadyCheck that will be run at initialization to allow easy checks to block known problematic configurations. An initial check to block on CentralPackageVersion is added while we investigate adding support for this.

See #252


mjrousos (approved):
LGTM

PR #282, opened 2021-02-26
Add option to provide entry-point and make non-interactive mode visible

sunandabalu:


src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/UpgradeOptions.cs
```diff
@@ -16,6 +16,8 @@ public class UpgradeOptions
 
         public bool SkipBackup { get; set; }
 
+        public string EntryPoint { get; set; } = string.Empty;
```
twsouthwick, line 19: Should this be marked as `string?`?

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Solution/EntrypointSelectionStep.cs
```diff
@@ -79,6 +85,29 @@ protected UpgradeStepInitializeResult InitializeImpl(IUpgradeContext context)
...
+
+                    return new UpgradeStepInitializeResult(UpgradeStepStatus.Complete, "Selected user's choice of entry point project.", BuildBreakRisk.None);
+                }
+                else
+                {
+                    Logger.LogInformation("Invalid Entry-Point Project specified : {Project}", _entryPoint);
+                }
+            }
+
+            if (context.EntryPoint is null && _isNonInteractiveMode)
```
twsouthwick, line 105: Instead of a flag like this, why not have `IUserInput` say whether it is interactive, ie `IUserInput.IsInteractive`?

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Solution/EntrypointSelectionStep.cs
```diff
@@ -13,13 +14,18 @@ public class EntrypointSelectionStep : UpgradeStep
     {
         private readonly IPackageRestorer _restorer;
         private readonly IUserInput _userInput;
+        private readonly bool _isNonInteractiveMode;
+        private readonly string _entryPoint;
 
         public EntrypointSelectionStep(
+            UpgradeOptions options,
```
twsouthwick, line 21: I'm not a fan of passing in `UpgradeOptions` to things - this should be viewed more as an object for the console parsing. I know we're currently using it in a few steps, so maybe it's ok. At the moment, there are options that wouldn't matter potentially if hosted elsewhere, though

twsouthwick (approved):
A couple of comments to think about. I'm approving as I'll be out for the next week, and none of the issues I've raised are necessarily blockers.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Solution/EntrypointSelectionStep.cs
sunandabalu: This is what I tried first, but it seems unnecessary in case of interactive mode as this is relevant only in the non-interactive mode.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Solution/EntrypointSelectionStep.cs
sunandabalu: We do need to pass the entry-point arg along, so it made sense to pass UpgradeOptions along similar to BackupStep

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/UpgradeOptions.cs
sunandabalu: fixed

PR #283, opened 2021-02-26
Add a sample demonstrating how to make a custom upgrade step

mjrousos:


samples/UpgradeStepSample/UpgradeStepSample/AuthorsPropertyUpgradeStep.cs
```diff
@@ -0,0 +1,213 @@
...
+        }
+
+        // ***** ***** *****
+        // Required overrides
+        // ***** ***** *****
+
+        /// <summary>
+        /// Gets a unique ID string used for referring to the upgrade step from other steps.
+        /// </summary>
+        public override string Id => typeof(AuthorsPropertyUpgradeStep).FullName!;
```
twsouthwick, line 53: should we make this be a virtual method and default to GetType().FullName!;?

samples/UpgradeStepSample/UpgradeStepSample/AuthorsPropertyUpgradeStep.cs
```diff
@@ -0,0 +1,213 @@
...
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            if (_options?.Authors is null)
+            {
+                throw new InvalidOperationException("AuthorsPropertyUpgradeStep should not be applied without a valid authors option set.");
+            }
+
+            // The IProjectFile abstraction exposes the csproj file.
```
twsouthwick, line 146: does it matter to be csproj? Or just project file? (thinking of the vb users here)

samples/UpgradeStepSample/UpgradeStepSample/SampleExtensionServiceProvider.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// service collection to register services in and the extension's configuration file.</param>
+        /// <returns>The service collection updated with services Upgrade Assistant should use.</returns>
+        public IServiceCollection AddServices(ExtensionServiceConfiguration serviceConfiguration)
+        {
+            if (serviceConfiguration is null)
+            {
+                throw new ArgumentNullException(nameof(serviceConfiguration));
+            }
+
+            return serviceConfiguration.ServiceCollection.AddScoped<UpgradeStep, AuthorsPropertyUpgradeStep>();
```
twsouthwick, line 36: Can we add a method of `AddStep<>`?  Just to ensure steps are added with the right scope? As far as scope, do we need to to be scoped or would transient be fine?

samples/UpgradeStepSample/UpgradeStepSample/UpgradeStepSample.csproj
```diff
@@ -0,0 +1,20 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
```
twsouthwick, line 4: These should be netstandard2.0 in case we decide to host in vs

twsouthwick (approved):
A couple of comments, but nothing big.

samples/UpgradeStepSample/UpgradeStepSample/AuthorsPropertyUpgradeStep.cs
mjrousos: Yea, that makes sense. I'll do that.

samples/UpgradeStepSample/UpgradeStepSample/AuthorsPropertyUpgradeStep.cs
mjrousos: Good point. Fixed.

samples/UpgradeStepSample/UpgradeStepSample/UpgradeStepSample.csproj
mjrousos: That's a good point, but it will be a somewhat larger change to make this work on NetStandard because Microsoft.DotNet.UpgradeAssistant.Extensions loads extensions with AssemblyLoadContext. I'll merge this as-is for now and assign myself a follow-up issue to make all the upgrade steps netstandard.

samples/UpgradeStepSample/UpgradeStepSample/SampleExtensionServiceProvider.cs
mjrousos: Good idea. The scope question is interesting. Transient should be fine, I think. I think the reason behind them originally having a scoped lifetime was that the upgrade steps may store state (between init and apply) and I wanted to make sure that different services resolving a step within that scenario would get the same step. So there's theoretically some use in the scope lifetime. In practice, though, I think we only resolve them once and then just pass them around, so transient is probably fine.

PR #290, opened 2021-03-02
Update README.md to add FAQ section

sunandabalu:


README.md
```diff
@@ -149,6 +149,25 @@ Concepts referred to in this repository which may have unclear meaning are expla
 ## Roadmap
 Take a look at the high level overview of the roadmap for this tool and the journey to upgrade your apps from .NET Framework to .NET 5 and beyond in the [roadmap](docs/roadmap.md).
 
+## FAQ
+
+**401 Unauthorized using dotnet tool install**: <br/>
```
mjrousos, line 154: The error could also be `Failed to retrieve information about 'upgrade-assistant' from remote source` if the source is missing rather than just authenticated (though that's probably less common).

mjrousos (approved):
I just made the same change (but in a different spot in the docs)! It may be worth taking both changes.

sunandabalu:
This is covered by #291

PR #317, opened 2021-03-05
Allow sub-steps to be conditional on project components

mjrousos:
This enables config updaters, analyzers, and code fix providers to use an attribute (`[ApplicableComponents]`) to specify the types of projects they apply to. Sub-steps based on these types will only show up as applicable if the project being upgraded has the indicated components.


.editorconfig
```diff
@@ -278,7 +278,3 @@ dotnet_diagnostic.CA1711.severity = none
 
 # CA1716: Using a reserved keyword as the name of a namespace makes it harder for consumers in other languages to use the namespace
 dotnet_diagnostic.CA1716.severity = none
-
-# SA1200: Using directives must be placed within namespace
-dotnet_diagnostic.SA1200.severity = none
```
mjrousos, line 283: This was a duplicate

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Configuration/ConfigUpdaterSubStep.cs
```diff
@@ -27,7 +27,18 @@ public ConfigUpdaterSubStep(UpgradeStep parentStep, IConfigUpdater configUpdater
             _configUpdater = configUpdater ?? throw new ArgumentNullException(nameof(configUpdater));
         }
 
-        protected override bool IsApplicableImpl(IUpgradeContext context) => context?.CurrentProject is not null && (_parentStep?.ConfigFiles.Any() ?? false);
```
KSchlobohm, line 30: We no longer need the ConfigFiles piece of this conditional?

mjrousos: The parent step (ConfigUpdaterStep) checks that there are config files and won't apply if there aren't any, so we don't need to check it here in the sub-steps also. I removed this because it was causing issues because I moved around when IsApplicable is called a bit and now the ConfigFiles aren't available when this is invoked.

tests/components/Microsoft.DotNet.UpgradeAssistant.Tests/UpgraderTests.cs
```diff
@@ -51,7 +51,7 @@ public async Task UpgraderStepsEnumeration()
             // in the expected order
             using var context = mock.Mock<IUpgradeContext>().Object;
 
-            var steps = upgrader.GetStepsForContext(context);
+            var steps = upgrader.AllSteps;
```
sunandabalu, line 54: We don't need conditional steps in tests?

sunandabalu (approved):

tests/components/Microsoft.DotNet.UpgradeAssistant.Tests/UpgraderTests.cs
mjrousos: We could leave this; I pulled it out because the test doesn't have an steps that don't apply so it was just an extra call that didn't do anything.

src/components/Microsoft.DotNet.UpgradeAssistant/UpgraderManager.cs
```diff
@@ -40,11 +40,9 @@ public async Task<IEnumerable<UpgradeStep>> InitializeAsync(IUpgradeContext cont
                 await _restorer.RestorePackagesAsync(context, context.EntryPoint, token).ConfigureAwait(false);
             }
 
-            return GetStepsForContext(context);
+            return AllSteps;
```
KSchlobohm, line 43: Why return `AllSteps` and then filter in `ConsoleUpgrade.cs` by applicable? Why not keep the filter here?

mjrousos: I debated this. I thought it was better to filter in ConsoleUpgrade for a couple reasons:

1. ConsoleUpgrade has should filter anyhow to make sure that it's also filtering out sub-steps that don't apply.
2. Even though we don't show inapplicable steps, I feel like that's a UI/UX decision and there could conceivably be a front-end someday that wants to list things like that grayed out or something. So I figured it made more sense to pass the whole list to the front-end and let it make the decision about how to display things.

KSchlobohm: Cool!

KSchlobohm (approved):
Looks good!

PR #318, opened 2021-03-06
Install try-convert if not available

sunandabalu:
This change checks if try-convert tool is already present if not will install at the start of upgrade-assistant.

### DIFF
--- a/src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.ProjectFormat/TryConvertToolStartup.cs
+++ b/src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.ProjectFormat/TryConvertToolStartup.cs
@@ -1,8 +1,6 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-using System.Linq;
-using System.Runtime.Loader;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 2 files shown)


src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.ProjectFormat/TryConvertToolStartup.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            if (!_tryConvertTool.IsAvailable)
+            {
+                _logger.LogInformation("Try-Convert not found. This tool depends on the Try-Convert CLI tool, installing now to the path : {path}", _tryConvertTool.Path);
+
+                return _runner.RunProcessAsync(new ProcessInfo
+                {
+                    Command = "dotnet",
+                    Arguments = " tool install -g try-convert"
+                }, token).ConfigureAwait(true).GetAwaiter().GetResult();
```
mjrousos, line 41: Rather than use GetResult here, it would be better to just do this directly in `StartupAsync` in an async way. Also, when you await it there, it's better to use `ConfigureAwait(false)` since this is in library code.

mjrousos (changes requested):

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.ProjectFormat/TryConvertToolStartup.cs
sunandabalu: Changed.

sunandabalu:
Closing this, since #322 addresses with installing a local tool instead of global.

PR #322, opened 2021-03-08
Include try-convert as part of build

twsouthwick:
This change adds a target that will download try-convert at build time and include it in the package output.


twsouthwick:
An alternative to #318

sunandabalu (approved):
LGTM, other than we need to manage versions for try-convert we need in the props file, which is Ok since try-convert does not get that many revisions.

twsouthwick:
We can always update it to use versions.props if that gets hooked up

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/ToolsInstaller.props
```diff
@@ -0,0 +1,16 @@
+<Project>
+  <PropertyGroup>
+    <!-- By default, NuGet will not include files starting with '.' or ending with '.nupkg'. Since we are including try-convert via global tools, some of the files fit into this. -->
```
mjrousos, line 3: Nit-pick, I think you mean 'local tools' here rather than 'global tools'

mjrousos (approved):
Looks good!

ToolsInstaller.targets
```diff
@@ -0,0 +1,24 @@
...
+    <NoDefaultExcludes>true</NoDefaultExcludes>
+    <TryConvertVersion>0.7.212201</TryConvertVersion>
+    <ToolsDirectory>$(MSBuildProjectDirectory)\tools\</ToolsDirectory>
+  </PropertyGroup>
+
+  <ItemGroup Condition="$(InstallTools)">
+    <None Include="$(ToolsDirectory)**\*">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <Content Include="$(ToolsDirectory)**\*" />
```
mjrousos, line 17: Why `None` and `Content` both?

Directory.Build.targets
```diff
@@ -11,13 +12,4 @@
     <Compile Include="$(MSBuildThisFileDirectory)/shared/**/*.cs" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)'!='netstandard2.0'">
-    <!-- Exclude MSBuild runtime assets from both src and test projects
-         as they shouldn't be present in this solution's output paths.
-         Instead, these dependencies should be loaded from the selected
-         MSBuild's location. -->
-    <PackageReference Include="Microsoft.Build" ExcludeAssets="runtime" />
-    <PackageReference Include="Microsoft.Build.Framework" ExcludeAssets="runtime" />
```
mjrousos, line 20: This is probably ok, but it would be worth double-checking that these libraries don't show up in our artifacts directories after the change. There was an issue previously where some projects were pulling these in transitively (likely fixed now) and it resulted in the binaries being present at runtime, which is not what we want. Microsoft.Build.Locator should be the only MSBuild library in the CLI's bin path.

ToolsInstaller.targets
```diff
@@ -0,0 +1,24 @@
...
+  </ItemGroup>
+  
+  <Target Name="RestoreTools" BeforeTargets="Build" Condition="$(InstallTools) AND !Exists($(ToolsDirectory))">
+    <Exec Command="dotnet tool install try-convert --version $(TryConvertVersion) --tool-path $(ToolsDirectory)" />
```
mjrousos, line 21: Is it possible to use `dotnet tool restore` instead (with a checked-in tools manifest)? If that works, it would mean that our dev machines would auto-update the local try-convert version when we rev'd it in the manifest.

If that doesn't work, though, this is probably good enough. We can just manually delete try-convert when it needs updated.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.ProjectFormat/TryConvertTool.cs
```diff
@@ -35,7 +35,7 @@ public class TryConvertTool : ITryConvertTool
                 throw new ArgumentNullException(nameof(tryConvertOptionsAccessor));
             }
 
-            Path = Environment.ExpandEnvironmentVariables(tryConvertOptionsAccessor.Value.TryConvertPath);
+            Path = tryConvertOptionsAccessor.Value.TryConvertPath;
```
mjrousos, line 38: I'd leave the `Environment.ExpandEnvironmentVariables`. It won't be needed 99% of the time, but we're pulling the path from a user-modifiable config file so who knows what will end up there.

mjrousos (approved):
Looks good. A couple comments inline.

ToolsInstaller.targets
twsouthwick: I tried that originally and there doesn't seem a way to restore to a tool-path

PR #324, opened 2021-03-09
Add package map entries for configuration and EF

### LINKED ISSUE #300: Upgraded web projects should not have a dependency on System.Configuration
Configuration is handled differently in ASP.NET Core
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-5.0

image

twsouthwick:
Fixes #300


mjrousos (approved):
LGTM

mjrousos:
Looks like both C# integration tests had these references and end up a little more upgraded now.

```
Assert.Equal() Failure\r\n                                 ↓ (pos 629)\r\nExpected: ···nce Include="System.Configuration" />\r\n    <Reference Include···\r\nActual:   ···nce Include="System.IdentityModel" />\r\n    <Reference Include···\r\n      
```

PR #325, opened 2021-03-09
Remove  MSBuild from being included in projects

twsouthwick:
I think this is a remnant from earlier iterations, but doesn't appear needed anymore.

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -11,13 +11,4 @@
     <Compile Include="$(MSBuildThisFileDirectory)/shared/**/*.cs" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)'!='netstandard2.0'">
-    <!-- Exclude MSBuild runtime assets from both src and test projects
-         as they shouldn't be present in this solution's output paths.
-         Instead, these dependencies should be loaded from the selected
-         MSBuild's location. -->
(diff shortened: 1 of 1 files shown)


Directory.Build.targets
```diff
@@ -11,13 +11,4 @@
     <Compile Include="$(MSBuildThisFileDirectory)/shared/**/*.cs" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)'!='netstandard2.0'">
-    <!-- Exclude MSBuild runtime assets from both src and test projects
-         as they shouldn't be present in this solution's output paths.
-         Instead, these dependencies should be loaded from the selected
-         MSBuild's location. -->
-    <PackageReference Include="Microsoft.Build" ExcludeAssets="runtime" />
-    <PackageReference Include="Microsoft.Build.Framework" ExcludeAssets="runtime" />
```
mjrousos, line 20: We added this previously because some projects were transitively referencing MSBuild packages and it was resulting in MSBuild binaries showing up in the CLI's output path. I agree that this is probably not needed now, but please double-check that no MSBuild binaries other than `Microsoft.Build.Locator` show up under artifacts.

mjrousos (approved):
One comment, otherwise this looks good.

twsouthwick:
Hmmm seems needed actually

PR #326, opened 2021-03-09
Enable logging for tests

twsouthwick:
This change updates the tests to log output to test ITestOutputHelper XUnit provides. This allows test results to show all the logging output from the tests.

In order to do this, instead of just allowing configuration of services, we pass a delegate in that allows modifying the IHostBuilder itself.

### DIFF
--- a/src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Program.cs
+++ b/src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Program.cs
@@ -13,8 +13,8 @@ using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.DotNet.UpgradeAssistant.Extensions;
-using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.Hosting;
 using Serilog;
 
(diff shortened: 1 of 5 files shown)


tests/tool/Integration.Tests/E2ETest.cs
```diff
@@ -34,27 +41,23 @@ public async Task UpgradeTest(string scenarioName, string language, string input
...
 
-                // Run upgrade
-                await UpgradeRunner.UpgradeAsync(Path.Combine(workingDir, inputFileName), entrypoint, Console.Out, 300).ConfigureAwait(false);
-                CleanupBuildArtifacts(workingDir);
+            // Run upgrade
+            await UpgradeRunner.UpgradeAsync(Path.Combine(workingDir, inputFileName), entrypoint, _output, 300).ConfigureAwait(false);
 
-                await AssertDirectoriesEqualAsync(Path.Combine(scenarioDir, UpgradedProjectSubDir), workingDir).ConfigureAwait(false);
-            }
-            finally
```
mjrousos, line 52: Was the try/finally just removed because UpgradeAsync won't throw or is there another reason?

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Program.cs
```diff
@@ -118,25 +125,17 @@ private static Task RunCommandAsync(UpgradeOptions options, Action<HostBuilderCo
 
                     services.AddSingleton(new InputOutputStreams(Console.In, Console.Out));
                     services.AddSingleton<CommandProvider>();
-                    services.AddSingleton(logSettings);
+                    services.TryAddSingleton(new LogSettings(true));
```
mjrousos, line 128: Should this go in the integration test's UpgradeRunner so that all necessary services are registered there?

mjrousos (approved):
Looks good. A couple non-blocking questions inline.

tests/tool/Integration.Tests/E2ETest.cs
twsouthwick: I didn't want the test directories deleted if it failed. This way you can go to the directory and see what failed

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Program.cs
twsouthwick: There's a command that requires it so this will add a default one if needed

PR #327, opened 2021-03-09
Add flag to exclude MSBuild runtime libraries

twsouthwick:


mjrousos (approved):
Looks good! As before, let's make sure MSBuild binaries don't show up in artifacts (to confirm that `<ExcludeMSBuildRuntime>true</ExcludeMSBuildRuntime>` has been applied to the right projects).

Issue #286, opened 2021-03-01
Upgrade Assistant should automatically replace BinaryFormatter.UnsafeDeserialize usage


mjrousos:
[`BinaryFormatter.UnsafeDeserialize`](https://apisof.net/catalog/0fa9da7e-02eb-f21a-6ef9-f235ec58ea4e) doesn't exist for .NET 5.0. However, `BinaryFormatter.Deserialize` is always the recommended alternative, so this should be an easy update to make automatically with a code fix updater.

PR #339, opened 2021-03-10
Replace BinaryFormatter.UnsafeDeserialization

KSchlobohm:
1. Adds an analyzer
2. Adds a code fixer
3. Adds console integration test scenario for c#
4. Adds unit test for new analyzer and code fixer

Addresses  #286


src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/BinaryFormatterUnsafeDeserializeAnalyzer.cs
```diff
@@ -0,0 +1,103 @@
...
+                _ => null
+            };
+
+            // Return if the accessed identifier wasn't from a simple member access expression, object creation expression, or identifier
+            if (accessedIdentifier is null)
+            {
+                return;
+            }
+
+            // If the accessed identifier resolves to a type symbol other than System.Web.HttpContext, then bail out
```
mjrousos, line 75: Copy/paste error in this comment ("System.Web.HttpContext")

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/BinaryFormatterUnsafeDeserializeFixer.cs
```diff
@@ -0,0 +1,94 @@
...
+            // Register a code action that will invoke the fix.
+            context.RegisterCodeFix(
+                CodeAction.Create(
+                    CodeFixResources.BinaryFormatterUnsafeDeserializeTitle,
+                    cancellationToken => ReplaceUnsafeDeserializationAsync(context.Document, node, cancellationToken),
+                    nameof(CodeFixResources.BinaryFormatterUnsafeDeserializeTitle)),
+                context.Diagnostics);
+        }
+
+        private static async Task<Solution> ReplaceUnsafeDeserializationAsync(Document document, SyntaxNode node, CancellationToken cancellationToken)
```
mjrousos, line 53: This can be simplified a bit. Code fix provider actions can return either Solution or Document. In this case, you only need to change the single document, so returning a document will be simpler.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/BinaryFormatterUnsafeDeserializeFixer.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            var project = document.Project;
+            var slnEditor = new SolutionEditor(project.Solution);
+            var docEditor = await slnEditor.GetDocumentEditorAsync(document.Id, cancellationToken).ConfigureAwait(false);
+            var docRoot = (CompilationUnitSyntax)docEditor.OriginalRoot;
+
+            if (!(node is MemberAccessExpressionSyntax))
```
mjrousos, line 60: If you change this to `if (node is not MemberAccessExpressionSyntax memberExpression)`  (or something similar) than you will have a non-null `MemberAccessExpressionSyntax` variable called memberExpression that you can use after this, so you can skip line 66 and not have to bother with !.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/BinaryFormatterUnsafeDeserializeFixer.cs
```diff
@@ -0,0 +1,94 @@
...
+            {
+                // stop processing - the code fixer only fixes member invocations of UnsafeDeserialize
+                return document.Project.Solution;
+            }
+
+            var memberExpression = node as MemberAccessExpressionSyntax;
+
+            if (memberExpression!.Expression is IdentifierNameSyntax)
```
mjrousos, line 68: This can be simplified. I think if you just do `var replacementSyntax = memberExpression.WithName("Deserialize")` it will be a lot more concise.

tests/tool/Integration.Tests/Integration.Tests.csproj
```diff
@@ -25,6 +25,9 @@
     <ProjectReference Include="..\..\..\src\steps\Microsoft.DotNet.UpgradeAssistant.Steps.Packages\Microsoft.DotNet.UpgradeAssistant.Steps.Packages.csproj" />
     <ProjectReference Include="..\..\Microsoft.DotNet.UpgradeAssistant.TestHelpers\Microsoft.DotNet.UpgradeAssistant.TestHelpers.csproj" />
   </ItemGroup>
+  <ItemGroup>
+    <Folder Include="IntegrationScenarios\WpfSample\vb\" />
```
mjrousos, line 29: I think this is unnecessary.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/BinaryFormatterUnsafeDeserializeAnalyzer.cs
```diff
@@ -0,0 +1,103 @@
...
+        {
+            if (context is null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze);
+
+            context.RegisterSyntaxNodeAction(AnalyzeMemberAccessExpressions, SyntaxKind.SimpleMemberAccessExpression);
```
mjrousos, line 41: Unfortunately, I think you'll need to analyze invocation expressions instead of member access expressions. UnsafeDeserialize takes two arguments whereas Deserialize only takes one, so you're going to have to change the invocation's arguments.

In most cases when UnsafeDeserialize is used, the second argument is null and it's ok to omit it when calling Deserialize instead. So, the change here is probably to only produce diagnostics for calls to UnsafeDeserialize that have a NullLiteralExpression for the second argument. Then update the code fix provider to remove the second (null) argument.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/BinaryFormatterUnsafeDeserializeFixer.cs
mjrousos: Though, as mentioned above, you'll also need to change the invocation expression's argument list to remove the second argument.

mjrousos (changes requested):
Thanks for adding this! I added some comments to the analyzer and code fix provider.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/BinaryFormatterUnsafeDeserializeFixer.cs
```diff
@@ -0,0 +1,94 @@
...
+            {
```
twsouthwick, line 69: Many of these actions should be able to be done in a language agnostic way via the DocumentEditor and the SyntaxGenerator it produces

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/BinaryFormatterUnsafeDeserializeAnalyzer.cs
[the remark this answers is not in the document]
twsouthwick: See RegisterMemberAccess and SystemDeploymentAnalyzer for examples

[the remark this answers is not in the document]
mjrousos: This is another one that needs to work at the syntax level, though. After the TFM changes, `BinaryFormatter.UnsafeDeserialize` won't resolve to a symbol.

twsouthwick: We should raise the diagnostic regardless, but only have the codefixer do anything if it is not null

mjrousos: Agreed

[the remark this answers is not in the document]
KSchlobohm: Thanks @twsouthwick I hadn't considered that approach.

Thanks @mjrousos I see the helper invokes `RegisterOperationAction` and `RegisterSymbolAction` which are both executed at the completion of semantic analysis. I think we need to keep the `RegisterSyntaxNodeAction` here so I think this feedback is resolved.

KSchlobohm: Updated the unit tests. As-is, the analyzer will surface the diagnostic. This recommendation will be applied to  the code fixer.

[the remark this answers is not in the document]
twsouthwick: Why does this **need** to run at the syntax level? This is one where they can change to a supported API before. The only ones we **need** to run at syntax level are ones where we are not able to do that.

If we decide we **need** to run at the syntax level, we should abstract the code so it's not duplicated

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/BinaryFormatterUnsafeDeserializeAnalyzer.cs
```diff
@@ -0,0 +1,103 @@
...
+            {
+                return;
+            }
+
+            // If the accessed identifier resolves to a type symbol other than BinaryFormatter, then bail out
+            // since it means the user is calling some other similarly named API.
+            var accessedSymbol = context.SemanticModel.GetSymbolInfo(accessedIdentifier).Symbol;
+            if (accessedSymbol is ILocalSymbol localSymbol)
+            {
+                if (!TargetTypeSymbolName.Equals($"{localSymbol.Type.ContainingNamespace}.{localSymbol.Type.Name}", StringComparison.Ordinal))
```
twsouthwick, line 80: better to compare the namespace and type name separately - minimizes allocations in analyzers help with compilation performance (although here it may not be a big deal as at this point it has already filtered out a ton of nodes)

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/BinaryFormatterUnsafeDeserializeCodeFixer.cs
```diff
@@ -0,0 +1,100 @@
...
+            }
+
+            var invocationExpression = node.Parent as InvocationExpressionSyntax;
+            if (invocationExpression is null)
+            {
+                return;
+            }
+
+            var lastArg = invocationExpression.ArgumentList.Arguments.Last();
+            if (!"null".Equals(lastArg.GetText().ToString(), System.StringComparison.Ordinal))
```
twsouthwick, line 51: You should be able to check if it is a `null` parameter at the token level

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/BinaryFormatterUnsafeDeserializeCodeFixer.cs
```diff
@@ -0,0 +1,100 @@
...
+                return document;
+            }
+
+            var project = document.Project;
+            var slnEditor = new SolutionEditor(project.Solution);
+
+            var docEditor = await slnEditor.GetDocumentEditorAsync(document.Id, cancellationToken).ConfigureAwait(false);
+            var docRoot = docEditor.OriginalRoot;
+
+            var replacementExpression = node.Parent.ToFullString();
```
twsouthwick, line 79: why handle this as a string? Seems like you should could do this at the syntax level

twsouthwick (approved):
A couple of comments, but nothing to block this on. I'd rather not add a new integration test, though

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/BinaryFormatterUnsafeDeserializeCodeFixer.cs
mjrousos: Agreed, it seems like getting the syntax's expression as a MemberAccessExpressionSyntax and using WithName, and then using WithArgumentList or something like that would help to do this more directly.

mjrousos (approved):

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/BinaryFormatterUnsafeDeserializeAnalyzer.cs
```diff
@@ -0,0 +1,109 @@
...
+
+            // If the accessed identifier resolves to a type symbol other than BinaryFormatter, then bail out
+            // since it means the user is calling some other similarly named API.
+            var accessedSymbol = context.SemanticModel.GetSymbolInfo(accessedIdentifier).Symbol;
+            if (accessedSymbol is ILocalSymbol localSymbol)
+            {
+                // refactored from $"{localSymbol.Type.ContainingNamespace}.{localSymbol.Type.Name}" for performance reasons
+                // avoiding the memory pressure from creating an interpolated string due to the volume of times a syntax analyzer is executed
+                if (!TargetTypeSymbolNamespace.Equals(localSymbol.Type.ContainingNamespace.ToString(), StringComparison.Ordinal)
+                    && !TargetTypeSymbolName.Equals(localSymbol.Type.Name, StringComparison.Ordinal))
```
mjrousos, line 84: || instead of &&, I think. We return if either part of the name is wrong.

tests/tool/Integration.Tests/IntegrationScenarios/ConsoleSample/csharp/Upgraded/ConsoleApp1/Program.cs
```diff
@@ -0,0 +1,121 @@
+﻿using System;
```
mjrousos, line 1: I don't think we need an integration test for this. Unit testing should be sufficient.

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/assets/TestClasses/UA0012.cs
```diff
@@ -0,0 +1,96 @@
...
+        {
+            // Open the file containing the data that you want to deserialize.
+            var fs = new FileStream(ADDRESSES_FILE, FileMode.Open);
+            try
+            {
+                var formatter = new BinaryFormatter();
+
+                // Deserialize the hashtable from the file and
+                // assign the reference to the local variable.
+                var addresses = (Dictionary<string, string>)formatter.UnsafeDeserialize(fs, null);
```
mjrousos, line 34: I think you hit the major test cases, but if you want to be even more complete, you could make sure that SomeOtherNamespace.BinaryFormatter.UnsafeDeserialize doesn't get flagged by the analyzer (though that does seem like an uncommon scenario).

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/TestHelper.cs
```diff
@@ -121,7 +126,7 @@ public static async Task<Document> FixSourceAsync(string documentPath, string di
                     }
                 }
             }
-            while (diagnosticFixed);
+            while (diagnosticFixed && fixAttempts < MAX_TRIES);
```
mjrousos, line 129: We may want to just throw if fixAttempts gets too high to make sure tests fail.

PR #340, opened 2021-03-11
Add IComponentIdentifier and ProjectComponents.AspNetCore

### LINKED ISSUE #337: Identify ASP.NET Core as a component

twsouthwick:
This change extracts the component identification out of the project
implementation, adds tests, and adds support for a new
ProjectComponent.AspNetCore flag.

Fixes #337

### DIFF
--- a/src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ProjectComponents.cs
+++ b/src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ProjectComponents.cs
@@ -10,9 +10,10 @@ namespace Microsoft.DotNet.UpgradeAssistant
     {
         None = 0,
         WindowsDesktop = 1,
-        Web = 1 << 1,
+        AspNet = 1 << 1,
         WinRT = 1 << 2,
         Wpf = 1 << 3,
         WinForms = 1 << 4,
+        AspNetCore = 1 << 5,
     }
 }
--- a/src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IComponentIdentifier.cs
+++ b/src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IComponentIdentifier.cs
@@ -1,0 +1,10 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Microsoft.DotNet.UpgradeAssistant
+{
+    public interface IComponentIdentifier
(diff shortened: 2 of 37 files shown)


sunandabalu (approved):

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ProjectComponents.cs
```diff
@@ -14,5 +14,6 @@ public enum ProjectComponents
         WinRT = 1 << 2,
         Wpf = 1 << 3,
         WinForms = 1 << 4,
+        AspNetCore = 1 << 5,
```
mjrousos, line 17: Separating web and AspNetCore complicates things because it introduces the first case where a project's components will change over the course of the upgrade process. Is there a benefit to having them separate instead of having a single component that applies to projects that conceptually target web scenarios (regardless of the specific framework used)? 

If we do separate them, we'll need a few other changes -

1. We should probably rename 'Web' to something more specific so that it's clear we're talking about classic ASP.NET. Maybe `AspNet`?
2. This split will require changes to some of our upgrade steps.
    1. Templates used by the template inserter can specify components that projects have to have for the templates to apply. [Our default ones](https://github.com/dotnet/upgrade-assistant/blob/main/src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Templates/Templates/CSharpWebAppTemplates/TemplateConfig.json#L35) list `Web` but should be AspNetCore now, instead.
    2. Similarly, most of our analyzers and all of our code fix providers have attributes saying they're only for use on `Web` projects. They should switch to `AspNetCore`, too. 
        1. This one's a little tricky because we might want to use the `AspNetCore` analyzers during our analysis stage even though the project will still be `AspNet` at that point. But the analysis stage still needs a bit of design anyhow, so I'm not sure whether we'd run all the analyzers there or not (and I'm sure we could special case in `AspNetCore` ones if we needed to). I'm beginning to wonder if the analysis step might not just be running all the upgrade steps (non-interactively) on a temporary copy of the project and then reporting back what got changed.

mjrousos (changes requested):

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ProjectComponents.cs
twsouthwick: I've removed web and updated accordingly

mjrousos (approved):

PR #341, opened 2021-03-11
Update try-convert target to honor version

twsouthwick:
This change honors the version of try-convert for the dev loop as well
as only downloading it once for whichever projects may need it.


mjrousos (approved):
Good changes; thanks!

PR #346, opened 2021-03-11
Add check so ensure file can be loaded.

twsouthwick:
For instance, if you don't the PCL workload installed, the integration test will fail (currently hang). Because this is required to pass or other checks fail and throw exceptions, this change changes the checks to fail at the first one.

### DIFF
--- a/src/components/Microsoft.DotNet.UpgradeAssistant/Checks/CanLoadProjectFile.cs
+++ b/src/components/Microsoft.DotNet.UpgradeAssistant/Checks/CanLoadProjectFile.cs
@@ -1,0 +1,44 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.DotNet.UpgradeAssistant.Checks
+{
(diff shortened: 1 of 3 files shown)


src/components/Microsoft.DotNet.UpgradeAssistant/Checks/CanLoadProjectFile.cs
```diff
@@ -0,0 +1,44 @@
...
+            {
+                // Just need to access some values in the file to see if it can be properly loaded.
+                _ = project.GetFile().GetPropertyValue("SomeValue");
+                return Task.FromResult(true);
+            }
+#pragma warning disable CA1031 // Do not catch general exception types
+            catch (Exception e)
+#pragma warning restore CA1031 // Do not catch general exception types
+            {
+                _logger.LogError("Project {Name} can not be loaded: {Message}", project.FilePath, e.Message);
```
mjrousos, line 39: We can use `_logger.LogError(e, "Project {Name} can not be loaded", project.FilePath)` to have the logger automatically include the exception information in the log.

mjrousos (approved):

src/components/Microsoft.DotNet.UpgradeAssistant/Checks/CanLoadProjectFile.cs
twsouthwick: But then the stack trace gets in the way for the end user. I'd rather keep it at just the message for now.

PR #351, opened 2021-03-12
Only consider ordered projects in CurrentProjectSelection

### LINKED ISSUE #284: I am stuck here unable to select next project
[13:58:49 INF] Initializing upgrade step Select project to upgrade

Upgrade Steps

Entrypoint: D:\Projects\Neural-Network\NeuralNetwork\NeuralNetwork\NeuralNetwork.csproj

1. [Next step] Select project to upgrade
2. Finalize Solution

Choose a command:
   1. Apply next step (Select project to upgrade)
   2. Skip next step (Select project to upgrade)
   3. See more step details
   4. Configure logging
   5. Exit
> 1
[13:59:06 INF] Applying upgrade step Select project to upgrade
Here is the recommended order to upgrade. Select enter to follow this list, or input the project you want to start with.
   [Completed] NeuralNetwork
>
[13:59:35 ERR] Unknown selection: ''
> 4
[13:59:48 ERR] Unknown selection: '4'
>
[14:00:08 ERR] Unknown selection: ''
> 1
[14:00:13 ERR] Unknown selection: '1'
> quit
[14:00:21 ERR] Unknown selection: 'quit'
> 2
[14:00:31 ERR] Unknown selection: '2'
>
[14:00:37 ERR] Unknown selection: ''
> NeuralNetwork
[14:00:58 ERR] Unknown selection: 'NeuralNetwork'
>
[14:01:05 ERR] Unknown selection: ''
>
[14:01:21 ERR] Unknown selection: ''
> 2
[14:01:59 ERR] Unknown selection: '2'
> 1
[14:02:01 ERR] Unknown selection: '1'
> 2
[14:02:03 ERR] Unknown selection: '2'
> 3
[14:02:04 ERR] Unknown selection: '3'
> 4
[14:02:05 ERR] Unknown selection: '4'
> 5
[14:02:07 ERR] Unknown selection: '5'
>

mjrousos:
This fixes a bug where `CurrentProjectSelectionStep` would ask the user to pick a project but all of the projects would be complete and unselectable (since `CurrentProjectSelectionStep` looks at *all* projects to decide if upgrade is done or not but only shows the user *ordered* projects to pick from).

Fixes #284

### DIFF
--- a/src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Solution/CurrentProjectSelectionStep.cs
+++ b/src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Solution/CurrentProjectSelectionStep.cs
@@ -49,6 +49,12 @@ namespace Microsoft.DotNet.UpgradeAssistant.Steps.Solution
         // This upgrade step is meant to be run fresh every time a new project needs selected
         protected override bool ShouldReset(IUpgradeContext context) => context?.CurrentProject is null && Status == UpgradeStepStatus.Complete;
 
+        public override UpgradeStepInitializeResult Reset()
+        {
+            _orderedProjects = null;
+            return base.Reset();
+        }
+
         protected override async Task<UpgradeStepInitializeResult> InitializeImplAsync(IUpgradeContext context, CancellationToken token)
         {
             if (context is null)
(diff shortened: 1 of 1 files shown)


src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Solution/CurrentProjectSelectionStep.cs
```diff
@@ -17,6 +17,7 @@ public class CurrentProjectSelectionStep : UpgradeStep
         private readonly IUserInput _input;
         private readonly ITargetFrameworkMonikerComparer _tfmComparer;
         private readonly ITargetTFMSelector _tfmSelector;
+        private IProject[]? _orderedProjects;
```
twsouthwick, line 20: Should we set this to null after applying? Or on reset?

twsouthwick (approved):

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Solution/CurrentProjectSelectionStep.cs
mjrousos: That's an interesting question. On principle, I think you're right that this should probably be set to null on reset (since reset 'undoes' initialization). In practice, I don't think it matters since the project order shouldn't change.

Still, it's probably better to be correct in case anything changes in the future.

PR #366, opened 2021-03-19
Showing try-convert error for vbproj unit test projects

KSchlobohm:
Because the error was previously hidden the upgrade-assistant was treating this step as success.

The displayed error message will be similar to:
`[try-convert] '..\VbUnitTest1Converted.vbproj' has invalid Project Type Guids for test projects and is not supported.`

### DIFF
--- a/src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.ProjectFormat/TryConvertTool.cs
+++ b/src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.ProjectFormat/TryConvertTool.cs
@@ -21,7 +21,8 @@ namespace Microsoft.DotNet.UpgradeAssistant.Steps.ProjectFormat
         private static readonly string[] ErrorMessages = new[]
         {
             "This project has custom imports that are not accepted by try-convert",
-            "is an unsupported project type. Not all project type guids are supported."
+            "is an unsupported project type. Not all project type guids are supported.",
+            "has invalid Project Type Guids for test projects and is not supported."
         };
 
         private readonly IProcessRunner _runner;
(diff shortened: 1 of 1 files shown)


src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.ProjectFormat/TryConvertTool.cs
```diff
@@ -21,7 +21,8 @@ public class TryConvertTool : ITryConvertTool
         private static readonly string[] ErrorMessages = new[]
         {
             "This project has custom imports that are not accepted by try-convert",
-            "is an unsupported project type. Not all project type guids are supported."
+            "is an unsupported project type. Not all project type guids are supported.",
+            "has invalid Project Type Guids for test projects and is not supported."
```
mjrousos, line 25: These error messages come from [here](https://github.com/dotnet/try-convert/blob/main/src/MSBuild.Abstractions/MSBuildConversionWorkspace.cs#L346). If we want to be proactive, we could also add these to avoid needing to add them later:

* is a coded UI test. Coded UI tests are deprecated and not convertable to .NET Core.
* is an unsupported MSTest test type. Only Unit Tests are supported.
* does not have a supported OutputType.

mjrousos (approved):
Looks good

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.ProjectFormat/TryConvertTool.cs
```diff
@@ -21,7 +21,8 @@ public class TryConvertTool : ITryConvertTool
         private static readonly string[] ErrorMessages = new[]
         {
             "This project has custom imports that are not accepted by try-convert",
-            "is an unsupported project type. Not all project type guids are supported."
+            "is an unsupported project type. Not all project type guids are supported.",
```
twsouthwick, line 24: This pattern is ok for now but is very much culture dependent (does try-convert localize?). We should get try-convert to return error codes or some other mechanism we can use instead of checking for error messages.

twsouthwick (approved):
We should revisit error handling around try-convert in general to not check for magic strings. But not a blocker for now.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.ProjectFormat/TryConvertTool.cs
mjrousos: Agreed. But I think try-convert doesn't return different error codes and these messages aren't localized, so I expect this is fine for now and we can add a work item to update try-convert/upgrade-assistant to handle this better.

PR #370, opened 2021-03-19
Consolidate IProject.FilePath and .Directory

twsouthwick:
We don't need separate properties for these. This change replaces these two properties with a single `IProject.FileInfo` property.

### DIFF
--- a/src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.cs
+++ b/src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.cs
@@ -21,13 +21,13 @@ namespace Microsoft.DotNet.UpgradeAssistant.MSBuild
 
         public MSBuildWorkspaceUpgradeContext Context { get; }
 
-        public string FilePath { get; }
+        public FileInfo FileInfo { get; }
 
-        public string Directory => Path.GetDirectoryName(FilePath)!;
+        public string Directory => FileInfo.Directory!.FullName;
 
-        public MSBuildProject(MSBuildWorkspaceUpgradeContext context, IComponentIdentifier componentIdentifier, string path, ILogger logger)
+        public MSBuildProject(MSBuildWorkspaceUpgradeContext context, IComponentIdentifier componentIdentifier, FileInfo file, ILogger logger)
         {
-            FilePath = path;
+            FileInfo = file;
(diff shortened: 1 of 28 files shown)


src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.cs
```diff
@@ -21,13 +21,13 @@ internal partial class MSBuildProject : IProject
 
         public MSBuildWorkspaceUpgradeContext Context { get; }
 
-        public string FilePath { get; }
+        public FileInfo FileInfo { get; }
 
-        public string Directory => Path.GetDirectoryName(FilePath)!;
+        public string Directory => FileInfo.Directory!.FullName;
```
mjrousos, line 26: FileInfo.DirectoryName instead, perhaps?

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.ConfigUpdaters/AppSettingsConfigUpdater.cs
```diff
@@ -51,7 +51,7 @@ public async Task<bool> ApplyAsync(IUpgradeContext context, ImmutableArray<Confi
             var project = context.CurrentProject.Required();
 
             // Determine where appsettings.json should live
-            var appSettingsPath = Path.Combine(project.Directory ?? string.Empty, AppSettingsJsonFileName);
+            var appSettingsPath = Path.Combine(project.FileInfo.DirectoryName, AppSettingsJsonFileName);
```
mjrousos, line 54: DirectoryName can be null, I think, so we still need the ?? here and in the other similar places it was removed.

mjrousos (approved):
LGTM; left a couple nit-picky comments

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.ConfigUpdaters/AppSettingsConfigUpdater.cs
twsouthwick: There was no nullable error, so I assumed it was ok... we're targeting .net standard 2.0 which doesn't have those marked

PR #371, opened 2021-03-22
using float range and tryParse

KSchlobohm:
Addresses a bug with upgrade assistant being unable to operate on a C# Unit Test project. The version string, by default, is 16.* for the "Microsoft.NET.Test.Sdk" package.

### DIFF
--- a/src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/NuGetReference.cs
+++ b/src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/NuGetReference.cs
@@ -7,7 +7,7 @@ namespace Microsoft.DotNet.UpgradeAssistant
 {
     public record NuGetReference(string Name, string Version)
     {
-        public bool HasWildcardVersion => Version.Contains("*");
+        public bool HasWildcardVersion => Version.EndsWith("*", StringComparison.OrdinalIgnoreCase);
 
         public override string ToString()
         {
--- a/src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/NuGetReference.cs
+++ b/src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/NuGetReference.cs
@@ -7,7 +7,7 @@ namespace Microsoft.DotNet.UpgradeAssistant
 {
     public record NuGetReference(string Name, string Version)
     {
-        public bool HasWildcardVersion => Version.Equals("*", StringComparison.OrdinalIgnoreCase);
+        public bool HasWildcardVersion => Version.Contains("*");
 
         public override string ToString()
         {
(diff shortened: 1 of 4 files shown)


src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/NuGetReference.cs
```diff
@@ -7,7 +7,7 @@ namespace Microsoft.DotNet.UpgradeAssistant
 {
     public record NuGetReference(string Name, string Version)
     {
-        public bool HasWildcardVersion => Version.Equals("*", StringComparison.OrdinalIgnoreCase);
+        public bool HasWildcardVersion => Version.Contains("*");
```
twsouthwick, line 10: Can you add a comment as the forms wildcard version can take?

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/NuGetExtensions.cs
```diff
@@ -11,10 +11,23 @@ internal static class NuGetExtensions
         {
             if (nugetRef.HasWildcardVersion)
             {
+                if (FloatRange.TryParse(nugetRef.Version, out var range))
```
twsouthwick, line 14: Could you add some unit tests for these? We're looking to fill in some test gaps as much as possible

KSchlobohm: added

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/NuGetReference.cs
KSchlobohm: added

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages.Tests/NuGetExtensionsTests.cs
```diff
@@ -0,0 +1,41 @@
...
+            NuGetReference? reference = default;
+
+            Assert.Throws<ArgumentNullException>(() => reference!.GetNuGetVersion());
+        }
+
+        [InlineData("6.0.*")]
+        [InlineData("4.*")]
+        [InlineData("*")]
+        [Theory]
+        public void GetNuGetVersionCanHandleFloatingVersions(string version)
```
twsouthwick, line 20: What about passing in here the expected version? I'd rather be explicit here in the test about expected in and out rather than logic to create the expected

twsouthwick (approved):

PR #383, opened 2021-03-25
Allow multiple tfms in a project

### LINKED ISSUE #379: Support multi targeted dependencies
My legacy ASP.NET project has a dependency on a library project that is multi-targeting:
`<TargetFrameworks>netstandard2.1;net47</TargetFrameworks>`

The tool fails with:
_Expected project to have exactly one TargetFramework property. Found 0 TargetFramework properties for project ..._

twsouthwick:
This change expands `IProject.TFM` to return an `IEnumerable<TargetFrameworkMoniker>` so it can work with projects that have multiple tfms.

Fixes #379


src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.cs
```diff
@@ -197,34 +189,29 @@ IEnumerable<NuGetReference> GetTransitiveDependencies()
...
         {
             get
             {
                 if (IsSdk)
                 {
                     // Currently only supporting non-multi-targeting scenarios
-                    return new TargetFrameworkMoniker(GetSingleTFMProperty("TargetFramework"));
+                    return GetTfms("TargetFramework")
+#pragma warning disable CA1507 // Use nameof to express symbol names
+                        .Concat(GetTfms("TargetFrameworks"))
```
mjrousos, line 201: You could pull "TargetFramework" and "TargetFrameworks" out into constants to avoid needing the #pragma (and simultaneously remove the hard-coded strings which is arguably also nice).

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.cs
```diff
@@ -197,34 +189,29 @@ IEnumerable<NuGetReference> GetTransitiveDependencies()
...
             ProjectRoot.GetAllReferences().Select(r => r.AsReference()).ToList();
 
-        public TargetFrameworkMoniker TFM
+        public IEnumerable<TargetFrameworkMoniker> TargetFrameworks
         {
             get
             {
                 if (IsSdk)
                 {
                     // Currently only supporting non-multi-targeting scenarios
```
mjrousos, line 198: This comment can go away now

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.cs
```diff
@@ -197,34 +189,29 @@ IEnumerable<NuGetReference> GetTransitiveDependencies()
...
-
-                    if (propCount != 1)
-                    {
-                        _logger.LogError("Expected project to have exactly one {PropertyName} property. Found {TargetFrameworkCount} TargetFramework properties for project {Project}.", propertyName, propCount, FilePath);
-                        throw new UpgradeException($"Expected project to have exactly one {propertyName} property. Found {propCount} TargetFramework properties for project {FilePath}.");
-                    }
-
-                    return targetFrameworkProperties.Single().Value;
-                }
+                IEnumerable<string> GetTfms(string propertyName)
```
mjrousos, line 211: Some users have tried running the tool on project files without target framework properties before. We still want to fail nicely in that scenario. I haven't looked yet to see what happens if this returns an empty enumerable, but it may be worth logging an error here if GetTfms is going to return empty.

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildTargetTFMSelector.cs
```diff
@@ -68,36 +69,53 @@ public TargetFrameworkMoniker SelectTFM(IProject project)
...
-                try
+                foreach (var depTFM in dep.TargetFrameworks)
                 {
-                    if (_tfmComparer.IsCompatible(tfm, dep.TFM))
+                    try
                     {
-                        continue;
+                        if (_tfmComparer.IsCompatible(tfm, depTFM))
+                        {
+                            continue;
```
mjrousos, line 78: This isn't quite right. If the dependency has any TFMs that are compatible with our proposed TFM we should continue out of the *outer* loop (not just the inner one) since that's sufficient for the proposed TFM to make sense with the dependency.

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildTargetTFMSelector.cs
```diff
@@ -68,36 +69,53 @@ public TargetFrameworkMoniker SelectTFM(IProject project)
...
+                        }
+
+                        if (depTFM.IsNetCore || depTFM.IsNetStandard)
```
mjrousos, line 81: This should only happen if none of the depTFMs are compatible with tfm.

Probably the logic here now should be:

```
foreach (var dep in project.ProjectReferences
{
    if (dep.TargetFrameworks.Any(IsCompatible)) continue;
    if (dep.TargetFrameworks.Any(IsNetCore || IsNetStandard)) tfm = the netcore/netstandard one
}
```

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageLoader.cs
```diff
@@ -55,7 +55,8 @@ public async Task<bool> DoesPackageSupportTargetFrameworkAsync(NuGetReference pa
             }
 
             var packageFrameworks = await GetTargetFrameworksAsync(packageArchive, token).ConfigureAwait(false);
-            return packageFrameworks.Any(f => DefaultCompatibilityProvider.Instance.IsCompatible(NuGetFramework.Parse(targetFramework.Name), f));
+
+            return packageFrameworks.Any(f => targetFrameworks.All(targetFramework => DefaultCompatibilityProvider.Instance.IsCompatible(NuGetFramework.Parse(targetFramework.Name), f)));
```
mjrousos, line 59: This isn't right. This requires that the package support a single TFM that works with all of the project's TFMs. Instead, we need to make sure that for all the project's TFMs, there's at least one package TFM that will work for it.

For example, if a project targets net472;net5 and a NuGet package has target frameworks net461 and netcoreapp3.1, then this method should return true (since that project can use that package) but I think currently it will return false since none of the package frameworks work for all of the project frameworks.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IPackageLoader.cs
```diff
@@ -11,7 +11,7 @@ public interface IPackageLoader
     {
         IEnumerable<string> PackageSources { get; }
 
-        Task<bool> DoesPackageSupportTargetFrameworkAsync(NuGetReference packageReference, string cachePath, TargetFrameworkMoniker targetFramework, CancellationToken token);
+        Task<bool> DoesPackageSupportTargetFrameworkAsync(NuGetReference packageReference, string cachePath, IEnumerable<TargetFrameworkMoniker> targetFrameworks, CancellationToken token);
```
mjrousos, line 14: Nit-pick, we should change the name to DoesPackageSupportTargetFramework**s**Async since it takes multiple target frameworks now.

mjrousos (changes requested):

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.cs
paulomorgado: Side question, is everything mention [here](https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build) supported by the assistant?

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.cs
sunandabalu: +1 this is def one of those areas where we can avoid #pragma

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.cs
sunandabalu: This is a scenario that some of the 1P customers have hit, project files with no target framework properties.

twsouthwick: @paulomorgado Are you referring to the directory.build.props? I believe it should be as we're using MSBuild to evaluate stuff, but it'd be worth checking. I've created https://github.com/dotnet/upgrade-assistant/issues/387 to track that

twsouthwick: I've updated the TFM ready check to handle that

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildTargetTFMSelector.cs
twsouthwick: Cleaned this up and added some tests. Switched to use `TargetFrameworkComparer.Compare` and it simplifies the code

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildTargetTFMSelector.cs
twsouthwick: Rewrote and added tests

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageLoader.cs
twsouthwick: Good catch.

twsouthwick: This is a hard one for a unit test due to the nuget package dependency :(

twsouthwick:
@mjrousos I've responded to your feedback. Can you review again?

mjrousos (approved):
Looks good; one last comment on TFM selection which we already discussed over Teams.

Issue #376, opened 2021-03-23
Step 2 should not add Microsoft.AspNetCore.Mvc.NewtonsoftJson package


KSchlobohm:
When upgrading the `Microsoft.AspNetCore.Mvc.NewtonsoftJson` should not be added before setting the TFM and moving to ASP.NET Core. For the duration of step 2 a project is expected to target net48

PR #390, opened 2021-03-30
Implemented IsApplicable for PackageReferenceAnalyzers

KSchlobohm:
Addresses issue #376

Most analyzers are always applicable. The Newtonsoft Package Reference Analyzer is only applicable post TFM change.


twsouthwick:
What scenario does this fix? If it's not needed, it was just bailing out early. Why add this method instead?

KSchlobohm:
> What scenario does this fix?
This bug was opened when the analyzer was being added to Web Library projects. The code has since addressed that issue by adding the `project.OutputType == ProjectOutputType.Exe` condition to the analyzer. The other scenario where the code runs is prior to the TFM change for a Web head project. At that point the project is SDK style, using package reference and targeting net48. The solution is in a broken state at that point but this is a step that should be delayed until after changing the TFM.

> If it's not needed, it was just bailing out early. Why add this method instead?
Adding this method separates the two responsibilities the method was performing and follows the pattern established in other parts of the code `IConfigUpdater.cs`, `UpgradeStep.cs`, and to an extent `ApplicableComponentsAttribute`. These other parts of the code are performing a "filter" check prior to performing the operation. Filtering things is a recurring theme as part of the support for the VB epic and I predict this will go further as we continue to evolve support for more permutations of languages, project types, and targets.

> Seems like we shouldn't be saying the component is AspNetCore at this point - I'd investigate why that's happening. If we don't say it's AspNetCore, then it will just ignnore this.
The oddity is the head project, The code adds this ProjectComponent when the project uses SDK style and the target is "Microsoft.NET.Sdk.Web"

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages.Tests/Analyzers/NewtonsoftReferenceAnalyzerTests.cs
```diff
@@ -0,0 +1,115 @@
...
+        public async Task AnalyzerIsNotApplicableToNetFrameworkProjects()
+        {
+            // Arrange
+            using var mock = AutoMock.GetLoose();
+            var comparer = mock.Mock<ITargetFrameworkMonikerComparer>();
+            comparer.Setup(comparer => comparer.Compare(It.IsAny<TargetFrameworkMoniker>(), It.IsAny<TargetFrameworkMoniker>()))
+                .Returns(-1);
+
+            var analyzer = mock.Create<NewtonsoftReferenceAnalyzer>();
+            using var tokenSource = new CancellationTokenSource();
```
twsouthwick, line 56: you can just pass in `default` or `CancellationToken.None` (same thing)

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages.Tests/Microsoft.DotNet.UpgradeAssistant.Steps.Packages.Tests.csproj
```diff
@@ -18,5 +18,6 @@
   <ItemGroup>
     <PackageReference Include="NuGet.Versioning" />
     <ProjectReference Include="..\..\..\src\steps\Microsoft.DotNet.UpgradeAssistant.Steps.Packages\Microsoft.DotNet.UpgradeAssistant.Steps.Packages.csproj" />
+    <ProjectReference Include="..\..\components\Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests\Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests.csproj" />
```
twsouthwick, line 21: You shouldn't need to add this

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages.Tests/MSBuildStepTestCollection.cs
```diff
@@ -0,0 +1,20 @@
...
+using Xunit;
+
+namespace Microsoft.DotNet.UpgradeAssistant.Steps.Packages.Tests
+{
+    /// <summary>
+    /// Class (which is never instantiated) for collecting package step tests that
+    /// need test fixtures to register MSBuild and make NuGet packages available.
+    /// </summary>
+    [CollectionDefinition(Name)]
+    public class MSBuildStepTestCollection : ICollectionFixture<MSBuildRegistrationFixture>
```
twsouthwick, line 14: Why did you need to add this?

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Analyzers/NewtonsoftReferenceAnalyzer.cs
```diff
@@ -9,42 +9,70 @@
...
             }
 
-            if (state is null)
-            {
-                throw new ArgumentNullException(nameof(state));
-            }
-
             // This reference only needs added to ASP.NET Core exes
-            if (!(project.Components.HasFlag(ProjectComponents.AspNetCore) && project.OutputType == ProjectOutputType.Exe))
+            if (!(project.Components.HasFlag(ProjectComponents.AspNetCore)
```
twsouthwick, line 46: Can't this just go in the `AnalyzeAsync` method? Why do we need the `IsApplicable`?

twsouthwick (changes requested):

twsouthwick:
The difference between this and the other checks that have a `IsApplicable` is that they *do* something. These analyzers populate a list of what will be done. None of them show up as substeps or anything. That may be something we'd want to do, but we could just as easily take a look to see if there are any changes we need to perform via the state object.

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages.Tests/MSBuildStepTestCollection.cs
KSchlobohm: Added because the `IsApplicable()` logic checks for a [transitive dependency](https://github.com/dotnet/upgrade-assistant/blob/65a12f4e5efa98d052efe0c7ab533e117bc9e3d0/src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Analyzers/TransitiveDependencyExtensions.cs#L15) which ultimately calls `NuGetFramework.Parse(tfm.Name);`

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages.Tests/Microsoft.DotNet.UpgradeAssistant.Steps.Packages.Tests.csproj
KSchlobohm: thanks, corrected this to `TestHelpers.csproj`

mjrousos:
@twsouthwick makes a good point that the other cases where we have the IsApplicable/Apply pattern, the apply is actually making the change. In those cases, IsApplicable is needed to determine whether the step (whatever it is) makes sense to show in the UI as something that will be applied (though the actual update, if any, happens later).

That model might make sense for package analyzers if we end up splitting them into their own sub-steps in the future, but it probably isn't necessary yet. As @twsouthwick points out, the logic for 'should this apply' is basically the same as the 'figure out what should change' method.

mjrousos (commented):
I added some comments inline. I agree that it may make sense to not add `IsApplicable`, at least for the time being.

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages.Tests/MSBuildStepTestCollection.cs
twsouthwick: Ah. Hold on for my open PR to merge then it won't

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Analyzers/NewtonsoftReferenceAnalyzer.cs
```diff
@@ -9,42 +9,70 @@
...
+                && project.OutputType == ProjectOutputType.Exe
+                && !project.TargetFrameworks.Any(tfm => _tfmComparer.Compare(tfm, new TargetFrameworkMoniker("netcoreapp3.0")) < 0)))
```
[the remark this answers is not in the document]
twsouthwick: Technically, we'd want to ensure it's at least .net core 3.0, and not .net core 2.1. That being said, `.IsNetCore` is probably sufficient since 2.1 is not supported anymore

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages.Tests/MSBuildStepTestCollection.cs
twsouthwick: It's merged now, so you shouldn't need any of the msbuild related stuff you added

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Analyzers/TransitiveDependencyExtensions.cs
```diff
@@ -65,7 +65,13 @@ private static bool ContainsDependency(this IProject project, TargetFrameworkMon
 
         private static IEnumerable<LockFileTargetLibrary> GetAllDependencies(this IProject project, TargetFrameworkMoniker tfm)
         {
+            if (string.IsNullOrEmpty(project.LockFilePath))
```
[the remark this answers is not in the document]
twsouthwick: If you merge from main, you'll get a more robust check behind `project.NuGetReferences.GetAllDependencies()`

tests/components/Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests/MSBuildStepTestCollection.cs
```diff
@@ -13,6 +13,8 @@ namespace Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests
     [CollectionDefinition(Name)]
     public class MSBuildStepTestCollection : ICollectionFixture<MSBuildRegistrationFixture>
     {
+        // by design this class must be implemented in every test project that uses the MSBuildRegistrationFixture
+        // https://github.com/xunit/xunit/issues/409
```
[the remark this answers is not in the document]
twsouthwick: A good reason to minimize how much we need msbuild/nuget :) Almost got everything behind the `.MSBuild` project

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Analyzers/TransitiveDependencyExtensions.cs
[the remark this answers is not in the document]
KSchlobohm: I'll remove this. It was only added because the project.LockFilePath was null when unit testing and I couldn't easily mock the response from this method.

twsouthwick (approved):
One small, nit, but otherwise LGTM

PR #412, opened 2021-04-05
Restore packages if not already restored when nuget packages are retrieved

### LINKED ISSUE #402: Projects aren't always restored when retrieving dependencies
In order to identify transitive dependencies, we require that projects are restored as we get that info from the lock file. If not, we throw. We have attempted to add in restore steps to ensure that occurs, but the better fix would be to do it for the caller if it is not done. The error end users see is:

```
[11:17:03 ERR] Project is not restored: [Path].csproj
```

Naively, this would be as simple as updating to the following signature:

```csharp
Task<INuGetReference> GetNuGetReferences(CancellationToken token);
```
And call `IPackageRestorer.RestoreAsync` if the lockfile is not present.  However, this has a cascading effect on methods as it then requires callers to be async. This expands out to needing to make the following async:

- `UpgradeStep.IsApplicableImpl`
- `IComponentIdentifier.GetComponents`
- `ITargetTFMSelector.SelectTfm`

This ends up being quite viral. This should probably still be done, but I wanted to create an issue and get feedback before making the changes.

twsouthwick:
This change ended up being fairly viral due to the additional async requirement. In order to restore packages when necessary (and only when necessary), we end up with a number of new async methods.

Fixes #402


src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/ConsoleUpgrade.cs
```diff
@@ -145,7 +145,7 @@ private void ShowUpgradeSteps(IEnumerable<UpgradeStep> steps, IUpgradeContext co
                 _io.Output.WriteLine(step.Title);
                 nextStepFound = nextStepFound || step.Status != UpgradeStepStatus.Complete;
 
-                ShowUpgradeSteps(step.SubSteps, context, currentStep, offset + 1);
+                await ShowUpgradeStepsAsync(step.SubSteps, context, token, currentStep, offset + 1);
```
KSchlobohm, line 148: Should this have a ConfigureAwait()?

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/ConsoleUpgrade.cs
```diff
@@ -131,7 +131,7 @@ private void ShowUpgradeSteps(IEnumerable<UpgradeStep> steps, IUpgradeContext co
                 }
             }
 
-            foreach (var step in steps.Where(s => s.IsApplicable(context)).ToList())
+            await foreach (var step in steps.ToAsyncEnumerable().WhereAwait(async s => await s.IsApplicableAsync(context, token)))
```
KSchlobohm, line 134: Should this have ConfigureAwait()?

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/ComponentIdentifier.cs
```diff
@@ -3,13 +3,15 @@
...
+using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
 
 namespace Microsoft.DotNet.UpgradeAssistant.MSBuild
 {
     public class ComponentIdentifier : IComponentIdentifier
     {
-        public ProjectComponents GetComponents(IProject project)
+        public async ValueTask<ProjectComponents> GetComponents(IProject project, CancellationToken token)
```
KSchlobohm, line 14: Just wanted to confirm.. 

Is the intent here to use ValueTask to describe an asynchronous return result that we expect to be synchronous most of the time?

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.NuGetPackages.cs
```diff
@@ -63,7 +75,7 @@ private bool ContainsDependency(TargetFrameworkMoniker tfm, Func<PackageDependen
 
         private IEnumerable<LockFileTargetLibrary> GetAllDependencies(TargetFrameworkMoniker tfm)
         {
-            if (LockFilePath is null)
+            if (!IsRestored)
             {
                 throw new InvalidOperationException("Project needs to be restored before continuing");
```
KSchlobohm, line 80: why throw here instead of `await _restorer.RestorePackagesAsync(Context, this, token).ConfigureAwait(false);`

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildWorkspaceUpgradeContext.cs
```diff
@@ -94,7 +97,7 @@ public IProject GetOrAddProject(FileInfo path)
                 return cached;
             }
 
-            var project = new MSBuildProject(this, _componentIdentifier, path, _logger);
+            var project = new MSBuildProject(this, _componentIdentifier, _restorer, path, _logger);
```
KSchlobohm, line 100: this is odd, it looks like a hidden dependency.

Is the thought process here that the MSBuildWorkspace and MSBuildProject types are always used together in pairs?

It could make it easier to test if this was a constructor dependency.

src/components/Microsoft.DotNet.UpgradeAssistant/UpgraderManager.cs
```diff
@@ -101,7 +96,7 @@ public async Task<IEnumerable<UpgradeStep>> InitializeAsync(IUpgradeContext cont
             // 2. Recurse into sub-steps (if needed)
             // 3. Return the step if it's not completed, or
             //    continue iterating with the next step if it is.
-            foreach (var step in steps)
+            await foreach (var step in steps)
```
KSchlobohm, line 99: ConfigureAwait here?

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Analyzers/UpgradeAssistantReferenceAnalyzer.cs
```diff
@@ -39,7 +39,8 @@ public async Task<PackageAnalysisState> AnalyzeAsync(IProject project, PackageAn
                 throw new ArgumentNullException(nameof(state));
             }
 
-            var packageReferences = project.Required().NuGetReferences.PackageReferences.Where(r => !state.PackagesToRemove.Contains(r));
+            var references = await project.Required().GetNuGetReferencesAsync(token).ConfigureAwait(false);
```
KSchlobohm, line 42: what does project.Required() do that we didn't do in a dozen other places?

KSchlobohm: And why do I have to know about it before I should access certain methods.

KSchlobohm (changes requested):
Found a few places I'd like to confirm regarding ConfigureAwait

KSchlobohm:
+1 for changelog.md

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/ConsoleUpgrade.cs
twsouthwick: We've disabled it in this project since we know it doesn't have a synchronization context

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/ConsoleUpgrade.cs
twsouthwick: same as above

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/ComponentIdentifier.cs
twsouthwick: yup

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.NuGetPackages.cs
twsouthwick: It should always be restored here

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildWorkspaceUpgradeContext.cs
twsouthwick: yeah, that would be nice

src/components/Microsoft.DotNet.UpgradeAssistant/UpgraderManager.cs
twsouthwick: done

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.NuGetPackages.cs
KSchlobohm: I like the change to file an issue as an error message. seems like a pattern we should have in other places.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source/CodeFixerStep.cs
```diff
@@ -68,7 +68,7 @@ protected override bool IsApplicableImpl(IUpgradeContext context)
 
             // Check the code fix provider for an [ApplicableComponents] attribute
             // If one exists, the step only applies if the project has the indicated components
-            return _fixProvider.GetType().AppliesToProject(context.CurrentProject);
+            return await _fixProvider.GetType().AppliesToProject(context.CurrentProject, token).ConfigureAwait(false);
```
KSchlobohm, line 71: Add Async name to method `AppliesToProject`

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source/SourceUpdaterStep.cs
```diff
@@ -116,7 +116,8 @@ public async Task GetDiagnosticsAsync(CancellationToken token)
             Logger.LogTrace("Running ASP.NET Core upgrade analyzers on {ProjectName}", project.Name);
 
             // Compile with upgrade analyzers enabled
-            var applicableAnalyzers = GetApplicableAnalyzers(_allAnalyzers, Project);
+            var applicableAnalyzers = await GetApplicableAnalyzers(_allAnalyzers, Project).ToListAsync(token).ConfigureAwait(false);
```
KSchlobohm, line 119: AsyncEnumerable should be a method named async

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.NuGetPackages.cs
```diff
@@ -14,7 +16,84 @@ namespace Microsoft.DotNet.UpgradeAssistant.MSBuild
...
+        {
+            get
+            {
+                var packagesConfig = GetPackagesConfigPath();
+
+                if (packagesConfig is null)
+                {
+                    var packages = ProjectRoot.GetAllPackageReferences();
+
+                    return packages.Select(p => p.AsNuGetReference()).ToList();
```
KSchlobohm, line 62: looks like this didn't change but it made me wonder if we should actually force the enumeration here.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IComponentIdentifier.cs
```diff
@@ -1,10 +1,13 @@
...
 
+using System.Threading;
+using System.Threading.Tasks;
+
 namespace Microsoft.DotNet.UpgradeAssistant
 {
     public interface IComponentIdentifier
     {
-        ProjectComponents GetComponents(IProject project);
+        ValueTask<ProjectComponents> GetComponents(IProject project, CancellationToken token);
```
KSchlobohm, line 11: method name should be `Async`

KSchlobohm (approved):
A few more Async changes. The rest LGTM

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source/SourceUpdaterStep.cs
twsouthwick: The pattern for async enumerables is to not include the async.

Issue #378, opened 2021-03-23
Do not add _ViewImports.cshtml to VB projects


KSchlobohm:
Do not add _ViewImports.cshtml to the vbproj web app

KSchlobohm:
#255

Issue #255, opened 2021-02-25
UpgradeSteps should be filtered based on project components


mjrousos:
Currently, Upgrade Assistant displays all upgrade steps and sub-steps for every project it upgrades. Many of the steps only make sense for certain app models, though (web, winforms, etc.).

Ideally, upgrade steps and sub-steps could optionally indicate which [components](https://github.com/dotnet/upgrade-assistant/blob/main/src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ProjectComponents.cs) they apply to and the tool would only show steps that apply to the project being upgraded.

PR #417, opened 2021-04-06
Filter features by language

KSchlobohm:
Addresses bugs
#255
#378 

Also hides the source code fixer step when there are no source code fixers available

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 - UpgradeSteps should be filtered based on project components [#255](https://github.com/dotnet/upgrade-assistant/issues/255)
-- Do not add _Imports.cshtml to VB projects [#378](https://github.com/dotnet/upgrade-assistant/issues/378)
+- Do not add _ViewImports.cshtml to VB projects [#378](https://github.com/dotnet/upgrade-assistant/issues/378)
 
 ## Version 0.2.220602 - 2021-04-06 ([Link](https://www.nuget.org/packages/upgrade-assistant/0.2.220602))
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added support for WPF and Unit Test projects written with Visual Basic [#403](https://github.com/dotnet/upgrade-assistant/pull/403)
 
 ### Fixed
-- Fixed issue surfacing from floating package references. [#371](https://github.com/dotnet/upgrade-assistant/pull/371)
-- Microsoft.AspNetCore.Mvc.NewtonsoftJson package should no longer be added to .NET Framework projects [#290](https://github.com/dotnet/upgrade-assistant/pull/390)
-- Fixed issue to ignore disabled NuGet sources. [#396](https://github.com/dotnet/upgrade-assistant/issues/396)
+- Allows projects to be read if they have multiple TFMs. The tool still can't upgrade them, but it won't block upgrading dependent projects [#379](https://github.com/dotnet/upgrade-assistant/issues/379)
+- Fixed issue surfacing from floating package references [#371](https://github.com/dotnet/upgrade-assistant/pull/371)
+- Microsoft.AspNetCore.Mvc.NewtonsoftJson package should no longer be added to .NET Framework projects [#376](https://github.com/dotnet/upgrade-assistant/issues/376)
+- Fixed issue to ignore disabled NuGet sources [#396](https://github.com/dotnet/upgrade-assistant/issues/396)
+- Ensured restore is run when required [#402](https://github.com/dotnet/upgrade-assistant/issues/402)
 
 ## Version 0.2.217201 - 2021-03-23  ([Link](https://www.nuget.org/packages/upgrade-assistant/0.2.217201))
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to the .NET Upgrade Assistant will be documented in this fil
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## Current
+
+### Fixed
+- UpgradeSteps should be filtered based on project components [#255](https://github.com/dotnet/upgrade-assistant/issues/255)
+- Do not add _Imports.cshtml to VB projects [#378](https://github.com/dotnet/upgrade-assistant/issues/378)
+
+## Version 0.2.220602 - 2021-04-06 ([Link](https://www.nuget.org/packages/upgrade-assistant/0.2.220602))
 ### Added
 - Added support for WPF and Unit Test projects written with Visual Basic [#403](https://github.com/dotnet/upgrade-assistant/pull/403)
 
--- a/src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ApplicableComponentsAttribute.cs
+++ b/src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ApplicableComponentsAttribute.cs
@@ -1,10 +1,9 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
+using System;
 
 namespace Microsoft.DotNet.UpgradeAssistant
 {
-    using System;
-
     /// <summary>
     /// An attribute for marking code fix providers that are only applicable when
     /// the project being upgraded contains certain components.
--- a/src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ApplicableComponentsAttribute.cs
+++ b/src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ApplicableComponentsAttribute.cs
@@ -1,10 +1,10 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-using System;
-
 namespace Microsoft.DotNet.UpgradeAssistant
 {
+    using System;
+
     /// <summary>
     /// An attribute for marking code fix providers that are only applicable when
     /// the project being upgraded contains certain components.
--- a/src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ApplicableLanguageAttribute.cs
+++ b/src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ApplicableLanguageAttribute.cs
@@ -13,18 +13,18 @@ namespace Microsoft.DotNet.UpgradeAssistant
     public sealed class ApplicableLanguageAttribute : Attribute
     {
         /// <summary>
-        /// Gets the language that a project must contain for the attributed feature to apply.
+        /// Gets a list of valid languages that a project could contain for the attributed feature to apply.
         /// </summary>
-        public Language Language { get; }
+        public Language[] Languages { get; }
 
         /// <summary>
         /// Initializes a new instance of the <see cref="ApplicableLanguageAttribute"/> class.
         /// </summary>
-        /// <param name="language">The language that a project must contain for the attributed
-        /// code fix provider to apply to it.</param>
-        public ApplicableLanguageAttribute(Language language)
+        /// <param name="languages">If a projec contains any of the the languages then the attributed
+        /// code fix provider will apply to it.</param>
+        public ApplicableLanguageAttribute(params Language[] languages)
         {
-            Language = language;
+            Languages = languages;
         }
     }
 }
--- a/src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ApplicableLanguageAttribute.cs
+++ b/src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ApplicableLanguageAttribute.cs
@@ -1,0 +1,30 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Microsoft.DotNet.UpgradeAssistant
+{
+    using System;
+
+    /// <summary>
+    /// An attribute for marking code fix providers that are only applicable when
+    /// the project being upgraded contains certain language.
+    /// </summary>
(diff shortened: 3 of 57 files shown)


sunandabalu (approved):
LGTM

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ApplicableComponentsAttribute.cs
```diff
@@ -1,10 +1,10 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-using System;
-
 namespace Microsoft.DotNet.UpgradeAssistant
 {
+    using System;
```
mjrousos, line 6: Convention for UA has been to have `using` statements outside namespaces.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ApplicableLanguageAttribute.cs
```diff
@@ -0,0 +1,30 @@
...
+
+namespace Microsoft.DotNet.UpgradeAssistant
+{
+    using System;
+
+    /// <summary>
+    /// An attribute for marking code fix providers that are only applicable when
+    /// the project being upgraded contains certain language.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false, Inherited = false)]
```
mjrousos, line 12: It may make sense to allow multiple in case something in the future works on, say, C# and F# but not VB or something weird like that.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ProjectExtensions.cs
```diff
@@ -58,7 +59,110 @@ public static async ValueTask<bool> AppliesToProjectAsync(this Type type, IProje
...
+            }
+
+            // if the attribute is not applied then this type defaults to opt into language support
+            return true;
+        }
+
+        /// <summary>
+        /// Checks to see if the type is a codefixer that should be filtered out of the user's view.
+        /// </summary>
+        /// <param name="type">a type that represents an upgrade-assistant feature decorated with <see cref="ExportCodeFixProviderAttribute"/>.</param>
```
mjrousos, line 99: Nit-pick: Other doc comments that we have capitalize the first letter of params, returns, etc. sentence style. Also, what's decorated in this case is a CodeFixProvider, not an upgrade-assistant specific feature.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ProjectExtensions.cs
```diff
@@ -58,7 +59,110 @@ public static async ValueTask<bool> AppliesToProjectAsync(this Type type, IProje
...
+        /// <param name="project">the project currently being upgraded.</param>
+        /// <returns>true if the codefixer should be hidden from the user. Will return false for non-codefixer objects.</returns>
+        private static bool DoesCodeFixerFilterThisFeature(Type type, IProject project)
+        {
+            var analyzerAttr = type.CustomAttributes.FirstOrDefault(a => a.AttributeType.FullName.Equals(typeof(ExportCodeFixProviderAttribute).FullName, StringComparison.Ordinal));
+            if (analyzerAttr is not null)
+            {
+                var applicableLanguage = analyzerAttr.ConstructorArguments.First().Value as string;
```
mjrousos, line 107: This only checks the attribute's first argument, but the attribute also takes a `params string[]` of "additional languages". We should check whether *any* of the languages match the project's language.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ProjectExtensions.cs
```diff
@@ -58,7 +59,110 @@ public static async ValueTask<bool> AppliesToProjectAsync(this Type type, IProje
...
+
+        /// <summary>
+        /// Checks to see if the type has an ApplicableLanguageAttribute that should filter the feature out of the user's view.
+        /// </summary>
+        /// <param name="type">a type that represents an upgrade-assistant feature decorated with <see cref="ApplicableLanguageAttribute"/>.</param>
+        /// <param name="project">the project currently being upgraded.</param>
+        /// <returns>true if the object should be hidden from the user. Will return false for objects not defning the language they support.</returns>
+        private static bool DoesApplicableLanguageFilterThisFeature(Type type, IProject project)
+        {
+            var applicableLangAttr = type.CustomAttributes.FirstOrDefault(a => a.AttributeType.FullName.Equals(typeof(ApplicableLanguageAttribute).FullName, StringComparison.Ordinal));
```
mjrousos, line 141: We might want to allow multiple ApplicableLanguage attributes, but I'm not sure how much of a priority that is.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source/SourceUpdaterStep.cs
```diff
@@ -69,7 +69,24 @@ public SourceUpdaterStep(IEnumerable<DiagnosticAnalyzer> analyzers, IEnumerable<
...
+            {
+                return false;
+            }
+
+            var applicableSubsteps = 0;
+            foreach (var substep in SubSteps)
+            {
+                if (await substep.IsApplicableAsync(context, token).ConfigureAwait(false))
+                {
+                    applicableSubsteps++;
```
mjrousos, line 84: Nit-pick - you could just return true here. No need to count the applicable sub-steps if we're only interested in knowing if >0 of them are true.

mjrousos (changes requested):
Looks good except for a concern about handling the additionalLanguages parameter on code fixers. Also, it would be nice to add some unit tests for these cases.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ProjectExtensions.cs
KSchlobohm: At the moment, if something applies to multiple languages then it wouldn't use the ApplicableLanguage filter. The default is opt-in to all languages.

Decided that if we need to filter something that applies to VB and C# but not F# that would be a design decision, and unit tests, for later. LMK if you think it's worth considering as part of this change.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ApplicableLanguageAttribute.cs
KSchlobohm: At the moment, if something applies to multiple languages then it wouldn't use the ApplicableLanguage filter. The default is opt-in to all languages.

Decided that if we need to filter something that applies to VB and C# but not F# that would be a design decision, and unit tests, for later. LMK if you think it's worth considering as part of this change.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ApplicableComponentsAttribute.cs
KSchlobohm: sounds good

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ProjectExtensions.cs
KSchlobohm: Once we added support for multiple languages to codefixers it made sense to just do the appicableLanguageAttr too

mjrousos: This one feels different to me than ApplicableLanguageAttr. The ability to specify multiple languages on a code fix provider is already present, so I think we should respect that additional languages that are specified. In fact, I'm not sure if it's possible to create a code fix provider *without* specifying language, so a multi-language code fix would have to include a list of languages (unlike other components that could just leave `[ApplicableLanguage]` off in those cases.

mjrousos (commented):
Looks good except that I think we still need to support multiple languages in the specific case of CodeFixProviders with multiple languages specified.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ProjectExtensions.cs
```diff
@@ -58,6 +66,145 @@ public static async ValueTask<bool> AppliesToProjectAsync(this Type type, IProje
...
+            var firstLanguage = analyzerAttr.ConstructorArguments.First().Value as string;
+            if (firstLanguage is null)
+            {
+                // this object is not the type we thought it was, the filter does not apply
+                return false;
+            }
+
+            supportedLanguages.Add(firstLanguage);
+
+            var additionalLanguages = analyzerAttr.ConstructorArguments.Last().Value as IEnumerable<System.Reflection.CustomAttributeTypedArgument>;
```
KSchlobohm, line 128: @mjrousos i think this is the part you're looking for. The constructor for codefixer actually stores the "additional languages" as a 2nd param

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ProjectExtensions.cs
KSchlobohm: agreed

CHANGELOG.md
```diff
@@ -3,6 +3,12 @@ All notable changes to the .NET Upgrade Assistant will be documented in this fil
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
+## Current
+
+### Fixed
+- UpgradeSteps should be filtered based on project components [#255](https://github.com/dotnet/upgrade-assistant/issues/255)
+- Do not add _Imports.cshtml to VB projects [#378](https://github.com/dotnet/upgrade-assistant/issues/378)
```
mjrousos, line 10: `_Imports.cshtml` -> `_ViewImports.cshtml`

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ProjectExtensions.cs
```diff
@@ -58,6 +66,145 @@ public static async ValueTask<bool> AppliesToProjectAsync(this Type type, IProje
...
+        {
+            var analyzerAttr = type.CustomAttributes.FirstOrDefault(a => a.AttributeType.FullName.Equals(typeof(ExportCodeFixProviderAttribute).FullName, StringComparison.Ordinal));
+            if (analyzerAttr is null)
+            {
+                // this object is not the type we thought it was, the filter does not apply
+                return false;
+            }
+
+            var supportedLanguages = new List<string>();
+            var firstLanguage = analyzerAttr.ConstructorArguments.First().Value as string;
```
mjrousos, line 119: This is probably unnecessary, but using FirstOrDefault might be safer here since I think it's preferable to have the attribute ignored than to throw an exception.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ProjectExtensions.cs
```diff
@@ -58,6 +66,145 @@ public static async ValueTask<bool> AppliesToProjectAsync(this Type type, IProje
...
+
+            supportedLanguages.Add(firstLanguage);
+
+            var additionalLanguages = analyzerAttr.ConstructorArguments.Last().Value as IEnumerable<System.Reflection.CustomAttributeTypedArgument>;
+            if (additionalLanguages is not null)
+            {
+                supportedLanguages.AddRange(additionalLanguages.Select(l => l.Value.ToString()));
+            }
+
+            foreach (var applicableLanguage in supportedLanguages)
```
mjrousos, line 134: Nit-pick: This could be more succinct with supportedLanuages.Any syntax. It will work fine as it is, though, so either way is fine with me.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ProjectExtensions.cs
```diff
@@ -58,6 +66,145 @@ public static async ValueTask<bool> AppliesToProjectAsync(this Type type, IProje
...
+            {
+                if (applicableLanguage is not null && !string.IsNullOrWhiteSpace(applicableLanguage))
```
mjrousos, line 136: The first check is redundant since the second one also checks null.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ProjectExtensions.cs
mjrousos, line 128: Again, LastOrDefault might be better here.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ProjectExtensions.cs
```diff
@@ -58,6 +66,145 @@ public static async ValueTask<bool> AppliesToProjectAsync(this Type type, IProje
...
+            foreach (var applicableLangInt in languageNumbers.Select(l => l.Value as int?))
+            {
+                if (!applicableLangInt.HasValue)
+                {
+                    // this object is not the type we thought it was, the filter does not apply
+                    return false;
+                }
+
+                var applicableLanguage = (Language)applicableLangInt;
+                if (project.Language == Language.CSharp && applicableLanguage == Language.CSharp)
```
mjrousos, line 192: Maybe just check `project.Language == applicableLanguage` instead of having three separate conditions?

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ProjectExtensions.cs
```diff
@@ -58,6 +66,145 @@ public static async ValueTask<bool> AppliesToProjectAsync(this Type type, IProje
...
+        private static bool DoesApplicableLanguageFilterThisFeature(Type type, IProject project)
+        {
+            var applicableLangAttr = type.CustomAttributes.FirstOrDefault(a => a.AttributeType.FullName.Equals(typeof(ApplicableLanguageAttribute).FullName, StringComparison.Ordinal));
+            if (applicableLangAttr is null)
+            {
+                // this object is not the type we thought it was, the filter does not apply
+                return false;
+            }
+
+            var languageNumbers = applicableLangAttr.ConstructorArguments.First().Value as IEnumerable<System.Reflection.CustomAttributeTypedArgument>;
```
mjrousos, line 176: FirstOrDefault, perhaps

mjrousos (approved):
Looks good; a few minor comments inline.

PR #420, opened 2021-04-08
Change queue to unblock PR and CI builds

sunandabalu:
As per suggestion from the eng services team, moving to a preview queue if possible could relieve the original vs2019 queue. This should also help us unblock our PR and CI builds.


KSchlobohm (approved):
That'd be great news

PR #421, opened 2021-04-09
Mock nuget for integration tests

### LINKED ISSUE #416: Mock package manager versions for integration tests
Currently, integration tests will fail if there is a newer version of a package available. This causes spurious failures that will cause builds to start failing.

KSchlobohm:
Implements an idea @twsouthwick had started to resolve #416

* Adds an Interceptor that wraps *PackageLoader.cs*. Known packages will always return the value defined in *ExpectedPackageVersions.json*
* Adds Autofac to the cli
* Removes the option that was used to set the Analyzer package to a constant version


tests/tool/Integration.Tests/InterceptingKnownPackageLoader.cs
```diff
@@ -0,0 +1,69 @@
...
+            if (_packages.TryGetValue(reference.Name, out var known))
+            {
+                return new NuGetReference[] { known };
+            }
+
+            var latest = await _other.GetNewerVersionsAsync(reference, latestMinorAndBuildOnly, token).ConfigureAwait(false);
+
+            if (latest is not null)
+            {
+                _logger.LogError("Unexpected check for newer version: {Name}, {Version}", reference.Name, reference.Version);
```
KSchlobohm, line 63: I think @twsouthwick had some specific feedback about this condition. Using **draft** PR to discuss.

tests/tool/Integration.Tests/ExpectedPackageVersions.json
```diff
@@ -0,0 +1,4 @@
+{
+  "Microsoft.AspNetCore.Mvc.NewtonsoftJson": "5.0.5",
```
KSchlobohm, line 2: It should only change when we're ready to update the tests!

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 ### Fixed
 - UpgradeSteps should be filtered based on project components [#255](https://github.com/dotnet/upgrade-assistant/issues/255)
 - Do not add _ViewImports.cshtml to VB projects [#378](https://github.com/dotnet/upgrade-assistant/issues/378)
+- Now mocking package manager versions for integration tests [#416](https://github.com/dotnet/upgrade-assistant/issues/416)
```
twsouthwick, line 11: Is this user facing?

Directory.Packages.props
```diff
@@ -3,6 +3,7 @@
   <ItemGroup>
     <PackageVersion Include="Automapper" Version="10.1.1" />
     <PackageVersion Include="Autofac.Extras.Moq" Version="6.0.0" />
+    <PackageVersion Include="Autofac.Extensions.DependencyInjection" Version="7.1.0" />
```
twsouthwick, line 6: Also bring in Autofac manually (there's a newer version than this will bring in)

tests/tool/Integration.Tests/InterceptingKnownPackageLoader.cs
twsouthwick: Mostly that I think we'd want to fail fast if the version is not known. ie we should pin all packages to a specific version. Ideally, the test would output a list when this wasn't the case that we could copy/paste into the json doc you have

Directory.Packages.props
KSchlobohm: added directly

tests/tool/Integration.Tests/InterceptingKnownPackageLoader.cs
KSchlobohm: We now fail the integration test and will report the packages that were not pinned

CHANGELOG.md
sunandabalu: Agreed, this does not fall into user facing category and probably does not need to be called out in the changelog.

twsouthwick (commented):
LGTM, except I'd rather we don't use static variables for the data

tests/tool/Integration.Tests/E2ETest.cs
```diff
@@ -50,14 +51,16 @@ public async Task UpgradeTest(string scenarioPath, string inputFileName, string
...
 
             // Run upgrade
-            var result = await UpgradeRunner.UpgradeAsync(Path.Combine(workingDir, inputFileName), entrypoint, _output, TimeSpan.FromMinutes(5)).ConfigureAwait(false);
+            var result = await upgradeRunner.UpgradeAsync(Path.Combine(workingDir, inputFileName), entrypoint, _output, TimeSpan.FromMinutes(5)).ConfigureAwait(false);
 
             Assert.Equal(ErrorCodes.Success, result);
 
             CleanupBuildArtifacts(workingDir);
 
+            AssertOnlyKnownPackagesWereReferenced(upgradeRunner.UnknownPackages ,workingDir);
```
twsouthwick, line 63: nit: weird comma placement

tests/tool/Integration.Tests/ExpectedPackageVersions.json
```diff
@@ -0,0 +1,9 @@
+{
+  "Hyak.Common": "1.2.2",
+  "MahApps.Metro": "2.4.4",
```
twsouthwick, line 3: This was causing build breaks as it was just updated. Happy to see this change!

twsouthwick (approved):

tests/tool/Integration.Tests/E2ETest.cs
KSchlobohm: yeah, heh, that's weird. thanks!

Issue #293, opened 2021-03-02
Put backups and logs in a folder or location not tracked by Git


mtaylorfsmb:
I suspect many people are going to run this tool on projects in Git/Github. If someone runs the tool and gets things working and then commits to Git then the backup projects and log files will be included using the default settings of the tool. 

It would probably be nice if the tool defaulted to placing the log file and backups someplace that would not get committed to Git by default. Maybe a `backup` folder or something that could be included in a `.gitignore` file would be nice. Even better would be if the tool looked for a `.gitignore` and added the location to it automatically or something.

sunandabalu:
#435 is a good design proposal for address backup folder locations, closing this in favor of that.

PR #422, opened 2021-04-09
Razor upgrade step (part 1)

mjrousos:
This is the first of two PRs to add a new Razor update step for upgrading cshtml files.

This PR includes:

1. The top-level `RazorUpdaterStep`
2. The `RazorUpdaterSubStep` which will be automatically added for each `IUpdater<RazorCodeDocument>` in the DI container
3. Unit tests for `RazorUpdaterStep`
4. `IUpdater` which implementers will use to define the actual update behavior for RazorCodeDocuments. This was created by generalizing the existing ConfigFileUpdater to be more generally useful.

A second PR next week will include important pieces to complete the feature:

1. An implementation of `IUpdater<RazorCodeDocument>` to update C# source in Razor documents using DI-registered analyzers and code fix providers.
2. Unit tests for `RazorUpdaterSubStep`
3. Unit tests for the implementation of `IUpdater<RazorCodeDocument>`


src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorUpdaterStepExtensions.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.DotNet.UpgradeAssistant.Steps.Razor;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.DotNet.UpgradeAssistant
```
KSchlobohm, line 7: is this namespace part of a pattern we've been using and I just didn't notice?

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorUpdaterStep.cs
```diff
@@ -0,0 +1,148 @@
...
+        /// determines whether it makes sense to even initialize the Razor updaters. So, it
+        /// will return false if there are no cshtml files in the current project or if there
+        /// are no Razor updaters available.
+        /// </summary>
+        /// <param name="context">The context to evaluate.</param>
+        /// <param name="token">A token that can be used to cancel execution.</param>
+        /// <returns>True if the Razor updater step might apply, false otherwise.</returns>
+        protected override Task<bool> IsApplicableImplAsync(IUpgradeContext context, CancellationToken token) =>
+            Task.FromResult(context?.CurrentProject is not null
+            && SubSteps.Any()
```
KSchlobohm, line 82: Can we ask the Razor substeps if they apply (similar to how we ask IUpdater<ConfigFile> substeps if they apply)

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorUpdaterStep.cs
```diff
@@ -0,0 +1,148 @@
...
+            _razorEngine = null;
+            _razorDocuments = null;
+            return base.Reset();
+        }
+
+        protected override Task<UpgradeStepApplyResult> ApplyImplAsync(IUpgradeContext context, CancellationToken token)
+        {
+            // Updates are made in sub-steps, so no changes need made in this apply step.
+            // Just check that sub-steps executed correctly.
+            var incompleteSubSteps = SubSteps.Count(s => !s.IsDone);
```
KSchlobohm, line 124: Do we need to filter substeps by applicable?

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorUpdaterStep.cs
```diff
@@ -0,0 +1,148 @@
...
+            && GetRazorFileSystem(context.CurrentProject.Required()).EnumerateItems("/").Any());
+
+        protected override async Task<UpgradeStepInitializeResult> InitializeImplAsync(IUpgradeContext context, CancellationToken token)
+        {
+            if (context is null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            _razorEngine = RazorProjectEngine.Create(RazorConfiguration.Default, GetRazorFileSystem(context.CurrentProject.Required()), builder =>
```
KSchlobohm, line 92: Do we ask is applicable before initializing a step? Seems like we shouldn't need `GetRazorFileSystem` if we can reuse `_razorEngine`

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorUpdaterStep.cs
```diff
@@ -0,0 +1,148 @@
...
+            {
+                // Add useful "MVC View-isms" like _ViewImports support, page models, etc.
+                // https://github.com/dotnet/aspnetcore/blob/main/src/Razor/Microsoft.AspNetCore.Mvc.Razor.Extensions/src/RazorExtensions.cs
+                RazorExtensions.Register(builder);
+            });
+
+            ProcessRazorDocuments();
```
KSchlobohm, line 99: I think this says we have to process razor documents foreach substep which would be a perf hit. I can't think of a reasonable workaround but it seems like something we should be fully aware of before adding.

tests/Microsoft.DotNet.UpgradeAssistant.TestHelpers/UpgradeStepExtensions.cs
```diff
@@ -18,5 +20,17 @@ public static void SetStatus(this UpgradeStep step, UpgradeStepStatus status)
 
             method.Invoke(step, new object[] { status });
         }
+
+        public static void ClearRazorDocuments(this RazorUpdaterStep step)
+        {
+            var field = typeof(RazorUpdaterStep).GetField("_razorDocuments", BindingFlags.NonPublic | BindingFlags.Instance);
```
KSchlobohm, line 26: maybe makes sense to use a folder path already included in the .gitignore file
https://github.com/github/gitignore/blob/master/VisualStudio.gitignore

e.g. "/bld"

#293

KSchlobohm:
Does this PR associate to a changelog item?

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorUpdaterStep.cs
mjrousos: Good idea. I'll make that update.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorUpdaterStep.cs
mjrousos: I don't think so. It probably wouldn't hurt, but inapplicable sub-steps should go straight to a completed state when they're initialized.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorUpdaterStep.cs
mjrousos: I have `GetRazorFileSystem` for use in `IsApplicable` (so that we don't have to setup the Razor engine there before we've determined if the step even applies).

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorUpdaterStep.cs
mjrousos: Yes, I was thinking about this too. I think a better design will be for sub-steps to indicate which files they changed (if any) and only re-process those documents. I'll see what that looks like.

tests/Microsoft.DotNet.UpgradeAssistant.TestHelpers/UpgradeStepExtensions.cs
mjrousos: I'm afraid I'm not following. This method is just a helper to clear the internal store of processed Razor documents for test purposes (to make sure they're regenerated at the right time).

mjrousos:
This is a partial fix for #367, but it will also need the follow-up PR coming later this week.

tests/Microsoft.DotNet.UpgradeAssistant.TestHelpers/UpgradeStepExtensions.cs
KSchlobohm: I think I misunderstood. Are the razor documents written to disk at this point?

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorUpdaterStep.cs
```diff
@@ -0,0 +1,148 @@
...
+        public override IEnumerable<string> DependsOn { get; } = new[]
+        {
+            // Project should be backed up before changing source
+            WellKnownStepIds.BackupStepId,
+
+            // Template files should be added prior to changing source (since some code fixers will change added templates)
+            WellKnownStepIds.TemplateInserterStepId,
+
+            // Project should have correct TFM
+            WellKnownStepIds.SetTFMStepId,
```
twsouthwick, line 40: We've been including all these, but are they necessary? What does this step actually require to complete?

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorUpdaterStepExtensions.cs
twsouthwick: The namespace of the service registrations have been at a higher level so that the Program.cs where we tie stuff together does not need to have a bunch of usings

tests/Microsoft.DotNet.UpgradeAssistant.TestHelpers/UpgradeStepExtensions.cs
```diff
@@ -18,5 +20,17 @@ public static void SetStatus(this UpgradeStep step, UpgradeStepStatus status)
 
             method.Invoke(step, new object[] { status });
         }
+
+        public static void ClearRazorDocuments(this RazorUpdaterStep step)
```
twsouthwick, line 24: Does this need to be in the testhelpers library? That will bring in the razorupdater dependency there.

twsouthwick (approved):
A couple questions, but LGTM

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorUpdaterStep.cs
mjrousos: I've tried to include the ones that we actually do want run first but it's certainly possible this list isn't right. My reasoning for these was:

1. Backup should run first since we'll be changing files.
2. Template inserter should run first since some of the source updaters depend on there being a startup.cs they can modify.
3. Set TFM should run first because most Razor updaters (applying existing code fixes to cshtml and replacing @Helper sections) will be updating be add .NET Core-specific code. Eventually we may want Razor updaters that can run before and after TFM change. That seems a very low priority at this point, though.

tests/Microsoft.DotNet.UpgradeAssistant.TestHelpers/UpgradeStepExtensions.cs
mjrousos: I can put it in the Razor step tests which might be cleaner.

tests/Microsoft.DotNet.UpgradeAssistant.TestHelpers/UpgradeStepExtensions.cs
mjrousos: The test razor files are present before the test runs. The tests I have so far just make sure they're read correctly. Nothing is changed yet (that's coming in the next PR :))

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorUpdaterStep.cs
twsouthwick: Hmm I see that reasoning. It just feels like it's extra, but that makes sense.

sunandabalu (approved):
LGTM. Nothing more to add on top of other comments here. :)

KSchlobohm (approved):

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorUpdaterStep.cs
mjrousos: I've fixed this. Updaters now return a list of files they changed and the sub-step only calls for those to be re-processed.

tests/Microsoft.DotNet.UpgradeAssistant.TestHelpers/UpgradeStepExtensions.cs
mjrousos: Fixed

PR #425, opened 2021-04-12
Add globbing for entrypoint so arbitrary entrypoints can be combined

twsouthwick:


Directory.Packages.props
```diff
@@ -21,6 +21,7 @@
     <PackageVersion Include="Microsoft.Extensions.Configuration.Json" Version="5.0.0" />
     <PackageVersion Include="Microsoft.Extensions.DependencyInjection" Version="5.0.0" />
     <PackageVersion Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="5.0.0" />
+    <PackageVersion Include="Microsoft.Extensions.FileSystemGlobbing" Version="5.0.0" />
```
mjrousos, line 24: I didn't know this package existed! Good to know about.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Solution/EntrypointResolver.cs
```diff
@@ -0,0 +1,48 @@
...
+            }
+
+            var matcher = new Matcher(StringComparison.OrdinalIgnoreCase);
+
+            matcher.AddIncludePatterns(names);
+
+            var count = 0;
+            var firstProject = default(IProject);
+
+            foreach (var project in projects)
```
mjrousos, line 31: It took me a few minutes to figure out what's going on here. This method could benefit from some comments.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Solution/EntrypointResolver.cs
```diff
@@ -0,0 +1,48 @@
...
+                firstProject = count == 0 ? project : null;
+
+                if (matcher.Match(project.FileInfo.Name).HasMatches)
+                {
+                    count++;
+                    yield return project;
+                }
+            }
+
+            if (names.Count == 0 && firstProject is not null)
```
mjrousos, line 42: If `names.Count == 0`, wouldn't that mean nothing will match? If so, `firstProject` will just be the last element of `projects` (since `count` will never increment). It seems like just yield returning `projects.LastOrDefault()` when `name.Count` is 0 would be a more readable way of getting that behavior. 

I'm not sure that's the behavior we want, though. If nothing matches, why arbitrarily pick the last project of the solution as the entry point? Shouldn't we return an empty enumerable and prompt the user to select an entry point in that case similar to the behavior we have today?

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Solution/EntrypointSelectionStep.cs
```diff
@@ -90,57 +96,36 @@ protected UpgradeStepInitializeResult InitializeImpl(IUpgradeContext context)
...
-
-#pragma warning disable CA1508 // Avoid dead conditional code
-            if (context.EntryPoint is null && _isNonInteractiveMode)
-#pragma warning restore CA1508 // Avoid dead conditional code
+            else
             {
-                throw new UpgradeException("Entry Point needs to be provided when more than 1 projects present in a non-interative mode. There are 2 ways of providing an entry-point :\n" +
-                    "a) Execute upgrade-assistant in interactive mode up until select entry point step and re-run upgrade-assistant in non-interactive mode \n" +
-                    "b) Execute upgrade-assistant in non-interactive mode with valid value for --entry-point");
+                return new UpgradeStepInitializeResult(UpgradeStepStatus.Failed, "No entrypoint was selected. Solutions require an entrypoint to proceed.", BuildBreakRisk.None);
```
mjrousos, line 114: Shouldn't we return Incomplete in this scenario so that the user can execute this step to select an entry point? As this change reads now, `EntrypointSelectionStep`'s `InitializeImplAsync` never returns incomplete - only complete or failed, so the user *has* to specify -e. They won't get an option of choosing an entrypoint interactively. Is that intentional?

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Solution/FinalizeSolutionStep.cs
```diff
@@ -30,23 +29,15 @@ public FinalizeSolutionStep(ILogger<FinalizeSolutionStep> logger)
         protected override Task<UpgradeStepApplyResult> ApplyImplAsync(IUpgradeContext context, CancellationToken token)
         {
             context.IsComplete = true;
-            context.SetEntryPoint(null);
```
mjrousos, line 33: I think we still need to clear entry points in this step, otherwise whatever entry point the user chose during this run will be persisted to the state file and they won't be prompted to pick a different entry point next time they run the tool on that solution even though at this point they're done upgrading from that particular entry point.

I think this regresses things so that if a user runs on a sln without -e, even after completing the upgrade, they won't be able to run on that same solution and pick a different entry point.

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.ProjectFormat.Tests - Copy/EntrypointResolverTests.cs
```diff
@@ -0,0 +1,53 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
mjrousos, line 1: This directory name has an errant " - Copy" in it.

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.ProjectFormat.Tests - Copy/EntrypointResolverTests.cs
```diff
@@ -0,0 +1,53 @@
...
+{
+    public class EntrypointResolverTests
+    {
+        [InlineData(new string[] { }, new string[] { }, new string[] { })]
+        [InlineData(new[] { "file1" }, new[] { "*" }, new[] { "file1" })]
+        [InlineData(new[] { "file1" }, new[] { "f" }, new string[] { })]
+        [InlineData(new[] { "file1" }, new[] { "i*" }, new string[] { })]
+        [InlineData(new[] { "file1" }, new[] { "i*", "f*" }, new[] { "file1" })]
+        [InlineData(new[] { "file1", "other" }, new[] { "i*", "f*" }, new[] { "file1" })]
+        [InlineData(new[] { "file1", "other" }, new[] { "o*", "f*" }, new[] { "file1", "other" })]
```
mjrousos, line 21: I would add a test case where `fileNames` is not empty but `globs` is.

mjrousos (changes requested):

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Solution/EntrypointResolver.cs
twsouthwick: It was picking this first project. I cannot remember why I did that.... I remember at the time it made sense, but I've removed it as it doesn't seem to make sense.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Solution/EntrypointSelectionStep.cs
twsouthwick: Yup!

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Solution/FinalizeSolutionStep.cs
twsouthwick: Reverted back, but to updated to check if any endpoints are set

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.ProjectFormat.Tests - Copy/EntrypointResolverTests.cs
twsouthwick: Good catch

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Solution/EntrypointSelectionStep.cs
sunandabalu: This line should retain the exception message from before.

README.md
```diff
@@ -86,14 +86,14 @@ Options:
...
+  --extension <extension>                      Specifies a .NET Upgrade Assistant
                                                extension package to include. This could
                                                be an ExtensionManifest.json file, a
                                                directory containing an
                                                ExtensionManifest.json file, or a zip
                                                archive containing an extension. This
                                                option can be specified multiple times.
-  --entry-point <entry-point>                  Provides the entry-point project to start
+  -e, --entry-point <entry-point>              Provides the entry-point project to start
                                                the upgrade process.
```
sunandabalu, line 97: Probably makes sense to add a line on globbing pattern that users can use? like ```upgrade-assistant <solution> -e *``` will run all projects in the solution.

mjrousos (approved):
LGTM

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Solution/EntrypointSelectionStep.cs
twsouthwick: Reverted that

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Program.cs
```diff
@@ -210,8 +211,8 @@ private static void ConfigureUpgradeCommand(Command command)
 
             command.AddArgument(new Argument<FileInfo>("project") { Arity = ArgumentArity.ExactlyOne }.ExistingOnly());
             command.AddOption(new Option<bool>(new[] { "--skip-backup" }, "Disables backing up the project. This is not recommended unless the project is in source control since this tool will make large changes to both the project and source files."));
-            command.AddOption(new Option<string[]>(new[] { "--extension", "-e" }, "Specifies a .NET Upgrade Assistant extension package to include. This could be an ExtensionManifest.json file, a directory containing an ExtensionManifest.json file, or a zip archive containing an extension. This option can be specified multiple times."));
-            command.AddOption(new Option<string>(new[] { "--entry-point" }, "Provides the entry-point project to start the upgrade process."));
+            command.AddOption(new Option<string[]>(new[] { "--extension" }, "Specifies a .NET Upgrade Assistant extension package to include. This could be an ExtensionManifest.json file, a directory containing an ExtensionManifest.json file, or a zip archive containing an extension. This option can be specified multiple times."));
```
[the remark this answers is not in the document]
twsouthwick: Done

sunandabalu (approved):

PR #426, opened 2021-04-13
Bug fix: When upgrading VB Web App the config step does not complete

KSchlobohm:
* The *ConfigUpdaterSubstep* should check IsApplicable (which calls `AppliesToProjectAsync`)
* The *AppSettingsConfigUpdater* should check to see if adding a key to the dictionary will cause an exception. This could happen if `IsApplicableAsync` us called more than once.

### DIFF
--- a/src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.ConfigUpdaters/AppSettingsConfigUpdater.cs
+++ b/src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.ConfigUpdaters/AppSettingsConfigUpdater.cs
@@ -154,7 +154,8 @@ namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Default.ConfigUpdaters
             // Check for existing appSettings.json files for app settings
             foreach (var setting in appSettings)
             {
-                if (!jsonConfigFiles.Any(s => !string.IsNullOrEmpty(s.Configuration[setting.Key])))
+                if (!jsonConfigFiles.Any(s => !string.IsNullOrEmpty(s.Configuration[setting.Key]))
+                    && !_appSettings.ContainsKey(setting.Key))
                 {
                     _appSettings.Add(setting.Key, setting.Value);
                 }
--- a/src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.ConfigUpdaters/AppSettingsConfigUpdater.cs
+++ b/src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.ConfigUpdaters/AppSettingsConfigUpdater.cs
@@ -154,10 +154,12 @@ namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Default.ConfigUpdaters
             // Check for existing appSettings.json files for app settings
             foreach (var setting in appSettings)
             {
-                if (!jsonConfigFiles.Any(s => !string.IsNullOrEmpty(s.Configuration[setting.Key]))
-                    && !_appSettings.ContainsKey(setting.Key))
+                if (!jsonConfigFiles.Any(s => !string.IsNullOrEmpty(s.Configuration[setting.Key])))
                 {
-                    _appSettings.Add(setting.Key, setting.Value);
+                    if (!_appSettings.ContainsKey(setting.Key))
+                    {
+                        _appSettings.Add(setting.Key, setting.Value);
+                    }
                 }
                 else
                 {
--- a/src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.ConfigUpdaters/AppSettingsConfigUpdater.cs
+++ b/src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.ConfigUpdaters/AppSettingsConfigUpdater.cs
@@ -156,10 +156,7 @@ namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Default.ConfigUpdaters
             {
                 if (!jsonConfigFiles.Any(s => !string.IsNullOrEmpty(s.Configuration[setting.Key])))
                 {
-                    if (!_appSettings.ContainsKey(setting.Key))
-                    {
-                        _appSettings.Add(setting.Key, setting.Value);
-                    }
+                    _appSettings[setting.Key] = setting.Value;
                 }
                 else
                 {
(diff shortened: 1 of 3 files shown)


KSchlobohm:
This issue has not been pushed to NuGet yet and does not have a correlating GH issue on our backlog. Decided not to update the changelog. LMK if the team thinks it should be part of the change.

mjrousos:
That makes sense to me. Fixing something that regressed after the last NuGet release probably doesn't need a line in the changelog.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Configuration/ConfigUpdaterSubStep.cs
```diff
@@ -90,7 +90,8 @@ protected override async Task<UpgradeStepInitializeResult> InitializeImplAsync(I
 
             try
             {
-                return await _configUpdater.IsApplicableAsync(context, _parentStep.ConfigFiles, token).ConfigureAwait(false)
+                return await IsApplicableAsync(context, token).ConfigureAwait(false)
```
mjrousos, line 93: I'm not sure we need to check `IsApplicableAsync` here (despite the fact that we're calling _configUpdater.IsApplicableAsync - I've filed #427 to fix the fact that these two different concepts have the same name).

If a ConfigUpdaterStep isn't applicable, users shouldn't be calling Initialize at all. If they do, then this check here will just cause the step to show up as 'not yet finished' and needing to be applied, which is wrong. If we do want to check `IsApplicable` here, it probably should throw or return Failed if it comes back false.

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.ConfigUpdaters/AppSettingsConfigUpdater.cs
```diff
@@ -156,7 +156,10 @@ public Task<bool> IsApplicableAsync(IUpgradeContext context, ImmutableArray<Conf
             {
                 if (!jsonConfigFiles.Any(s => !string.IsNullOrEmpty(s.Configuration[setting.Key])))
                 {
-                    _appSettings.Add(setting.Key, setting.Value);
+                    if (!_appSettings.ContainsKey(setting.Key))
```
mjrousos, line 159: This seems an ok precaution, but I wouldn't expect IConfigUpdater.IsApplicable to be called more than once in most cases. We may want to see why this is necessary and fix that also.

mjrousos (changes requested):
What is causing the config step to not complete? Is it just that it has inapplicable substeps? If that's the case, I think the correct fix is to fix `ConfigUpdaterStep.IsApplicableImplAsync` to confirm that there are applicable substeps (instead of just checking SubSteps.Any()).

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.ConfigUpdaters/AppSettingsConfigUpdater.cs
KSchlobohm: This wasn't directly necessary, this was something I discovered as a precaution.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Configuration/ConfigUpdaterSubStep.cs
KSchlobohm: A substep (WebNamespaceConfigUpdater) that was hidden from the user was being asked if it was complete. Per our discussion I've implemented a filter in the `ConfigUpdaterStep.cs` that supports reset.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Configuration/ConfigUpdaterStep.cs
```diff
@@ -90,6 +94,26 @@ protected override async Task<UpgradeStepInitializeResult> InitializeImplAsync(I
...
+                    applicableSubSteps.Add(substep);
+                }
+            }
+
+            SubSteps = applicableSubSteps;
+        }
+
+        public override UpgradeStepInitializeResult Reset()
+        {
+            SubSteps = _allSteps;
```
twsouthwick, line 113: Should we do `_allSteps = null` here?

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Configuration/ConfigUpdaterStep.cs
```diff
@@ -52,7 +54,7 @@ public ConfigUpdaterStep(IEnumerable<IConfigUpdater> configUpdaters, ConfigUpdat
             }
 
             _configFilePaths = (configUpdaterOptions.ConfigFilePaths ?? Enumerable.Empty<string>()).ToArray();
-            SubSteps = configUpdaters.Select(u => new ConfigUpdaterSubStep(this, u, logger)).ToList();
+            SubSteps = _allSteps = configUpdaters.Select(u => new ConfigUpdaterSubStep(this, u, logger)).ToList();
```
twsouthwick, line 57: It seems like `SubSteps` may be better as a virtual property so we could just override it for stuff like this

twsouthwick (approved):

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.ConfigUpdaters/AppSettingsConfigUpdater.cs
sunandabalu: you could just do _appSettings[setting.Key] = setting.Value;  here, it will do the same.

KSchlobohm: i like that, simpler to read.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Configuration/ConfigUpdaterStep.cs
mjrousos: That's a good idea.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Configuration/ConfigUpdaterStep.cs
mjrousos: No, because _allSteps is set in the ctor which won't be re-invoked when initializing the step for use with the next project. Reset should undo exactly whatever Initialize does.

mjrousos (approved):
Looks good, though I like @twsouthwick's idea of making SubSteps virtual. It might be a cleaner change.

sunandabalu (approved):

Issue #353, opened 2021-03-16
About MSBuild and Visual Studio dependency


Nadare2012:
The prerequisites say that the tool uses MSBuild (which can be installed with Visual Studio) and there is no word that Visual Studio is required. However, the tool looks for the path to Visual Studio and throws an exception if not found. If a tool needs Visual Studio to find MSBuild, can the --msbuild-path option be added, like in try-convert, to remove Visual Studio from dependencies?

twsouthwick:
We should probably update that we require VS. We could try adding an option to say where to look for it, but we need it to help load projects that are being converted. Many old projects require targets files that are installed with VS and fail if we don't have those targets.

sunandabalu:
@twsouthwick will update the wiki with clear instructions for VS / MSBuild

sunandabalu:
Nothing more at this moment to be updated here, closing this.

PR #428, opened 2021-04-13
Update README for prereqs

twsouthwick:
Part of #353


sunandabalu (approved):

Nadare2012:
If you try to use the tool (to convert default net48 console app, for example) without Visual Studio, it will fail with the error "Could not find a Visual Studio installation to use to update." So, the "This may include installing Visual Studio 2019" part is incorrect, currently Visual Studio must be installed.

PR #438, opened 2021-04-14
Replace ITargetTFMSelector with a filter-based pattern

twsouthwick:
This change doesn't change too much for callers, but changes how a TFM is selected. Previously, one class held all the knowledge about how to select a TFM for a project. This swaps that out for a filter chain that can attempt to change the final result. Each of the filters can do whatever checks they may want, but the ITargetFrameworkSelectorFilter.Builder that is passed through will only update the TFM if it is a higher or more constrained TFM.


twsouthwick:
I'd like y'all feedback on this. As I'm looking at approaching the LTS/Current/Preview change, I wanted to get this cleaned up first. I haven't updated the tests to account for this, but am interested on structure/naming/pattern feedback.

src/components/Microsoft.DotNet.UpgradeAssistant/TargetFramework/WindowsSdkTargetFrameworkSelectorFilter.cs
```diff
@@ -0,0 +1,46 @@
...
+        }
+
+        private static bool TryGetMoniker(ITargetFrameworkSelectorFilterState updater, [MaybeNullWhen(false)] out string tfm)
+        {
+            // Projects with Windows Desktop components or a WinExe output type should use a -windows suffix
+            if (updater.Components.HasFlag(ProjectComponents.WindowsDesktop) || updater.Project.OutputType == ProjectOutputType.WinExe)
+            {
+                tfm = $"{updater.AppBase}{WindowsSuffix}";
+
+                if (updater.Components.HasFlag(ProjectComponents.WinRT))
```
KSchlobohm, line 33: What does this project component identify? The name directed me to https://en.wikipedia.org/wiki/Windows_RT

src/components/Microsoft.DotNet.UpgradeAssistant/TargetFramework/TargetFrameworkSelector.cs
```diff
@@ -0,0 +1,104 @@
...
+        {
+            _comparer = comparer;
+            _selectors = selectors;
+
+            _currentTFMBase = selectorOptions?.Value.CurrentTFMBase ?? DefaultCurrentTFMBase;
+            _ltsTFMBase = selectorOptions?.Value.LTSTFMBase ?? DefaultLTSTFMBase;
+            _upgradeTarget = options?.UpgradeTarget ?? throw new ArgumentNullException(nameof(options));
+        }
+
+        public async ValueTask<TargetFrameworkMoniker> SelectTargetFrameworkAsync(IProject project, CancellationToken token)
```
KSchlobohm, line 35: Not directly tied to this change, but why do we call this method twice?

We invoke from `SetTFMStep.cs` during Initialize and during Apply

src/components/Microsoft.DotNet.UpgradeAssistant/TargetFramework/TargetFrameworkSelector.cs
```diff
@@ -0,0 +1,104 @@
...
+
+            foreach (var selector in _selectors)
+            {
+                selector.Process(updater);
+            }
+
+            return updater.Current;
+        }
+
+        private static TargetFrameworkMoniker GetNetStandardTFM(IProject project)
```
KSchlobohm, line 59: Looks more like `GetDefaultTargetFrameworkMoniker`

src/components/Microsoft.DotNet.UpgradeAssistant/TargetFramework/TargetFrameworkSelector.cs
```diff
@@ -0,0 +1,104 @@
...
+            var appBaseTfm = new TargetFrameworkMoniker(appBase);
+
+            var updater = new FilterState(_comparer, project, current, appBaseTfm)
+            {
+                Components = await project.GetComponentsAsync(token).ConfigureAwait(false),
+            };
+
+            foreach (var selector in _selectors)
+            {
+                selector.Process(updater);
```
KSchlobohm, line 53: I see now that this is more of the decorator pattern with conditionals. https://www.dofactory.com/net/decorator-design-pattern

We may want to order the decorators

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ITargetFrameworkSelectorFilter.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Microsoft.DotNet.UpgradeAssistant
+{
+    public interface ITargetFrameworkSelectorFilter
+    {
+        void Process(ITargetFrameworkSelectorFilterState tfm);
```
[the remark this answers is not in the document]
twsouthwick: I actually had taken a look at the best pattern it fit and found this one: https://www.tutorialspoint.com/design_pattern/filter_pattern.htm

I'm not sure what you mean by the `IsApplicable`

src/components/Microsoft.DotNet.UpgradeAssistant/TargetFramework/WindowsSdkTargetFrameworkSelectorFilter.cs
twsouthwick: It refers to WinRT components, which is a subset of COM used by UWP applications.

src/components/Microsoft.DotNet.UpgradeAssistant/TargetFramework/TargetFrameworkSelector.cs
twsouthwick: In initialize to see if we have set it to something that we expect, and the apply to to get what to set it to. We could cache it for later, but apparently don't

src/components/Microsoft.DotNet.UpgradeAssistant/TargetFramework/SatisifiesProjectDependenciesTargetFrameworkSelectorFilter.cs
```diff
@@ -0,0 +1,47 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.DotNet.UpgradeAssistant.TargetFramework
+{
+    public class SatisifiesProjectDependenciesTargetFrameworkSelectorFilter : ITargetFrameworkSelectorFilter
```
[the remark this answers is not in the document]
twsouthwick: Kind of. I've added a comment, but the effect is that we want to ensure that it's at least as high as the minimum of each of its dependencies.

src/components/Microsoft.DotNet.UpgradeAssistant/TargetFramework/TargetFrameworkSelector.cs
twsouthwick: There's no ordering involved here - we ensure that you can never downgrade a TFM in a filter.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ITargetFrameworkSelectorFilter.cs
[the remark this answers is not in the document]
KSchlobohm: I meant, that's the first name that popped in my head as a replacement for "Process". The code is doing a decision and where applicable calling another class.

src/components/Microsoft.DotNet.UpgradeAssistant/TargetFramework/TargetFrameworkSelector.cs
KSchlobohm: Yeah, that makes sense. I still think it could apply though.

One thing the order could do is put the SatisfiesProjectDependencies step at last and skip a bunch of NuGet calls if you're already targetting net5.0?

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ITargetFrameworkSelectorFilter.cs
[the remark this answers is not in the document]
twsouthwick: Got it. The `Is` prefix usually implies returning a decision - this does not really do that. It's processing the filter, which can chose to update the TFM.

src/components/Microsoft.DotNet.UpgradeAssistant/TargetFramework/TargetFrameworkSelector.cs
twsouthwick: What NuGet calls are you seeing? It's just looking at the TFM the dependencies are currently targeting.

src/components/Microsoft.DotNet.UpgradeAssistant/TargetFramework/MaybeNullWhenAttribute.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace System.Diagnostics.CodeAnalysis
+{
+    [AttributeUsage(AttributeTargets.Parameter, Inherited = false)]
+    internal sealed class MaybeNullWhenAttribute : Attribute
```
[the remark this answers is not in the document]
twsouthwick: Yup - I'll consolidate them so that there's not  duplication

src/components/Microsoft.DotNet.UpgradeAssistant/TargetFramework/WindowsSdkTargetFrameworkSelectorFilter.cs
```diff
@@ -0,0 +1,46 @@
...
+                {
+                    // TODO: Default to this version to ensure everything is supported.
+                    tfm += "10.0.19041.0";
```
mjrousos, line 36: I thought of another scenario where these filters won't layer together properly. 

Suppose the same scenario as before (WinExe project with a 5.0 AppBase and 6.0 dependency). If the filters are applied in the opposite order I suggested before, there's still a problem (just a different one):

1. First, the WindowsSdk filter would change Current to 5.0-Windows.
2. But then when the min dependency filter runs it will either change Current to 6.0 or (more likely) leave it as 5.0-Windows (depending on how 5.0-Windows and 6.0 compare). What we *want* is 6.0-Windows.

Ordering issues are creeping in because different filters are changing the TFM in different kinds of ways. We either need to track suffixes like this somewhere in the state so that filters really can execute on the different "parts" of a TFM independently (standard/netcore, AppBase, suffixes) or we need to make sure the filters run in a known good order.

mjrousos (changes requested):
I really like this approach but I think the current code has some unintentional ordering issues with the filters. Comments inline.

src/components/Microsoft.DotNet.UpgradeAssistant/TargetFramework/TargetFrameworkSelector.cs
KSchlobohm: Sorry, I was still using my old understanding of `SatisifiesProjectDependenciesTargetFrameworkSelectorFilter` which does not make NuGet calls.

KSchlobohm: Does this not have the opportunity to `break` the loop if the tfm is already as net5.0?

src/components/Microsoft.DotNet.UpgradeAssistant/TargetFramework/WindowsSdkTargetFrameworkSelectorFilter.cs
KSchlobohm: I think we could solve that by changing the way `TryUpdate` executes. We aren't looking to compare between 5.0-Windows and 6.0. We're looking to merge the two requirements together.

Can we say that merging 5.0-Windows and 6.0 always produces the same result?

src/components/Microsoft.DotNet.UpgradeAssistant/TargetFramework/TargetFrameworkSelector.cs
mjrousos: We don't want to break the loop because there's no way to know what future selectors might do. Even if we're targeting net5.0, it's possible that a future selector (min dependency selector, for example) could require a TFM higher than our app base!

src/components/Microsoft.DotNet.UpgradeAssistant/TargetFramework/WindowsSdkTargetFrameworkSelectorFilter.cs
```diff
@@ -0,0 +1,46 @@
...
+                tfm.TryUpdate(new TargetFrameworkMoniker(result));
+            }
+        }
+
+        private static bool TryGetMoniker(ITargetFrameworkSelectorFilterState updater, [MaybeNullWhen(false)] out string tfm)
+        {
+            // Projects with Windows Desktop components or a WinExe output type should use a -windows suffix
+            if (updater.Components.HasFlag(ProjectComponents.WindowsDesktop) || updater.Project.OutputType == ProjectOutputType.WinExe)
+            {
+                tfm = $"{updater.AppBase}{WindowsSuffix}";
```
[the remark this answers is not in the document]
twsouthwick: Good catch! I like that final idea to hide it behind TryUpdate

src/components/Microsoft.DotNet.UpgradeAssistant/TargetFramework/WindowsSdkTargetFrameworkSelectorFilter.cs
twsouthwick: Good point. I'm going to look at updating the TryUpdate to be smart about platforms

src/components/Microsoft.DotNet.UpgradeAssistant/TargetFramework/WindowsSdkTargetFrameworkSelectorFilter.cs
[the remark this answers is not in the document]
twsouthwick: I've added logic for that in TargetFrameworkMonikerComparer

src/components/Microsoft.DotNet.UpgradeAssistant/TargetFramework/WindowsSdkTargetFrameworkSelectorFilter.cs
twsouthwick: I've added logic for that in TargetFrameworkMonikerComparer

paulomorgado:
Nit picking...

`ITargetTFMSelector` violates the [framework design guidelines naming capitalization conventions](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions#capitalization-rules-for-identifiers):

> The PascalCasing convention, used for all identifiers except parameter names, capitalizes the first character of each word (including acronyms over two letters in length), as shown in the following examples:
> 
> `PropertyDescriptor` `HtmlTag`
> 
> A special case is made for two-letter acronyms in which both letters are capitalized, as shown in the following identifier:
> 
> `IOStream`

It should be `ITargetTfmSelector`.

tests/components/Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests/TargetFrameworkIdentifierTests.cs
```diff
@@ -80,5 +82,32 @@ public void TfmCompare(string target, string tfm, int expected)
             static TargetFrameworkMoniker? Create(string? input)
                 => input is null ? null : new TargetFrameworkMoniker(input);
         }
+
+        [InlineData("net6.0-windows", "net5.0-windows", "net6.0")]
+        [InlineData("net6.0-windows", "net6.0", "net5.0-windows")]
+        [InlineData("net6.0-windows", "net5.0", "net6.0-windows")]
+        [InlineData("net6.0-windows10.0.5", "net5.0", "net6.0-windows10.0.5")]
+        [InlineData("net6.0-windows10.0.5", "net5.0-windows10.0.5", "net6.0")]
+        [InlineData("net6.0-windows10.1.5", "net5.0-windows10.1.5", "net6.0-windows10.0.5")]
```
sunandabalu, line 91: hmm, Why would the expected be net6.0-windows10.1.5? Should it not be net6.0-windows10.0.5, coz that is the highest version (combination of name and version) there might not be a 10.1.5 for 6.0, is it not?

twsouthwick: When merging, you want to take the highest framework version (here that's net6.0) and the highest platform version (windows10.1.5)

src/components/Microsoft.DotNet.UpgradeAssistant/TargetFramework/TargetFrameworkSelector.cs
```diff
@@ -0,0 +1,100 @@
...
+
+            public IProject Project { get; }
+
+            public ProjectComponents Components { get; init; }
+
+            public bool TryUpdate(TargetFrameworkMoniker tfm)
+            {
+                Current = _comparer.Merge(Current, tfm);
+
+                return Current != tfm;
```
mjrousos, line 96: Should we check Current != tfm before merging? If Current starts as net5.0 and tfm is net5.0-windows, then after merging they'll be the same, but an update did happen.

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/NuGetTargetFrameworkMonikerComparer.cs
```diff
@@ -67,6 +72,85 @@ public int Compare(TargetFrameworkMoniker? x, TargetFrameworkMoniker? y)
...
+                throw new ArgumentNullException(nameof(tfm2));
+            }
+
+            var nugetTfm1 = NuGetFramework.Parse(tfm1.Name);
+            var nugetTfm2 = NuGetFramework.Parse(tfm2.Name);
+
+            // We can only combine if the platform is the same or at least one is null.
+            if (!IsPlatformSameOrNull(nugetTfm1, nugetTfm2))
+            {
+                throw new InvalidOperationException("Cannot merge two target frameworks with different platforms");
```
mjrousos, line 93: This is a bit picky because this isn't likely to come up often, but if a filter did try to apply a TFM with a different platform from the current one, we wouldn't want it to crash. We'd want the merge to fail and alert the user that the TFM we're using might not be right. So, it may be better to have this method return a bool instead of throwing in this case.

mjrousos (approved):
LGTM. A couple minor comments in-line.

src/components/Microsoft.DotNet.UpgradeAssistant/TargetFramework/TargetFrameworkSelector.cs
twsouthwick: Good catch!

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/NuGetTargetFrameworkMonikerComparer.cs
twsouthwick: Good idea. Add a log message instead

tests/components/Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests/TargetFrameworkIdentifierTests.cs
```diff
@@ -80,5 +82,32 @@ public void TfmCompare(string target, string tfm, int expected)
             static TargetFrameworkMoniker? Create(string? input)
                 => input is null ? null : new TargetFrameworkMoniker(input);
         }
+
+        [InlineData("net6.0-windows", "net5.0-windows", "net6.0")]
+        [InlineData("net6.0-windows", "net6.0", "net5.0-windows")]
+        [InlineData("net6.0-windows", "net5.0", "net6.0-windows")]
+        [InlineData("net6.0-windows10.0.5", "net5.0", "net6.0-windows10.0.5")]
```
KSchlobohm, line 89: Error	NETSDK1140	10.0.5.0 is not a valid TargetPlatformVersion for Windows. Valid versions include:
10.0.19041.0
10.0.18362.0
10.0.17763.0
8.0
7.0

Issue #435, opened 2021-04-14
Persisting knowledge about non-default backup location


oteinone:
## Summary

As #293 states, it is probably a bad idea to back up the solution inside the git folder. The ability to select another location for the backup helps but unfortunately the backup location is not persisted. 

## Motivation and goals

- Being able to store backups outside project/solution folder is good
- Having to do backups multiple times is bad

## Detailed design

The location could be persisted in `.upgrade-assistant` file and `BackupStep.InitializeImplAsync` could query for this location instead of the default location if it has been set.

sunandabalu:
@oteinone This makes sense, feel free to make a PR we will review it.

oteinone:
@sunandabalu Added PR, please let me know if something needs improving :)

twsouthwick:
This is merged in now. Thanks for the PR!

PR #447, opened 2021-04-20
Backup path is now persisted in upgrade file

oteinone:
Path in upgrade file is checked for existing backups

- [ x ] You've read the [Contributor Guide](../CONTRIBUTING.md) and [Code of Conduct](../CODE-OF-CONDUCT.md).
- [ x ] You've included unit or integration tests for your change, where applicable.
- [ x ] You've included inline docs for your change, where applicable.
- [ x ] There's an open issue for the PR that you are making. If you'd like to propose a new feature or change, please open an issue to discuss the change or find an existing issue.

**Persisting backup location**
Custom backup location is stored in .upgrade-assistant-file

**PR Description**
- Backup location (whether default or custom) is stored in .upgrade-assistant file
- If backup step is initialized again, this location will be checked for existing backup
- Currently if no backup is found, the backup step asks to fill in the backup location again
- I did not add any tests because FileUpgradeStateFactory is not included in test coverage

Addresses #435


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/upgrade-assistant?pullRequest=447) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/upgrade-assistant?pullRequest=447) before we can accept your contribution.<br/><br/>:x: oteinone [sign now](https://cla.dotnetfoundation.org/dotnet/upgrade-assistant?pullRequest=447)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/upgrade-assistant?pullRequest=447) it.</sub>

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IUpgradeContext.cs
```diff
@@ -27,6 +27,8 @@ public interface IUpgradeContext : IDisposable
 
         bool UpdateSolution(Solution updatedSolution);
 
+        string? BackupLocation { get; set; }
```
twsouthwick, line 30: Instead of an explicit property here, could you make it a property bag?

@mjrousos I know we've talked about this kind of thing in the past. Wanted to make sure you saw this conversation.

twsouthwick (changes requested):

sunandabalu:
Integration tests are failing in the PR build 
Error Message:
```
The contents of "BeanTraderClient\BeanTraderClient.csproj" do not match.\r\nAssert.Equal() Failure\r\n                                 ↓ (pos 1731)\r\nExpected: ···Metro" Version="2.4.4" />\r\n    <PackageReference Include="Nit···\r\nActual:   ···Metro" Version="2.4.5" />\r\n    <PackageReference Include="Nit···\r\n                                 ↑ (pos 1731)\r\nExpected: True\r\nActual:   False
```
Stacktrace:
```
at Integration.Tests.E2ETest.AssertDirectoriesEqualAsync(String expectedDir, String actualDir) in /_/tests/tool/Integration.Tests/E2ETest.cs:line 103
   at Integration.Tests.E2ETest.UpgradeTest(String scenarioPath, String inputFileName, String entrypoint) in /_/tests/tool/Integration.Tests/E2ETest.cs:line 61
--- End of stack trace from previous location ---
```

twsouthwick:
@oteinone Feel free to ignore that integration test - if you merge from main it should be fixed

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IUpgradeContext.cs
oteinone: @twsouthwick: Hopefully I understand correctly that with a property bag you mean a Dictionary construct. I  think that if we use a dictionary, the properties there should be used only from one context which would hold the property key. Otherwise the benefits gained would be outweighed by magic strings. BackupLocation is currently like that but if another step would want to use that, we would have to refactor this - and refactoring would be annoying because searching by reference would not work.

I was also thinking what kind of items would go into the property bag? Persisted items, settable items or something else? In both categories there will be items which are not strings and might require abstraction or generics.

I also noticed that IsComplete is seemingly designed to be persisted in file but it is never set. 

I'll happily make any changes, just verbalising my thoughts here.

mjrousos: FWIW, I'm removing IsComplete in another unrelated PR. It's vestigial and can go away.

twsouthwick: Thanks for the thoughts! The reason I'd prefer more of a property bag construct is that it allows future steps to be able to add properties they care about without needing to add properties to the context and doesn't special case the BackupStep as something the context explicitly knows about.

For now, a prototype like the following could work well:

```csharp
public interface IUpgradeContext
{
  string GetPropertyValue(string propertyName);

  void SetPropertyValue(string propertyName, string value);
}
```

However, if that feels like a larger work item for you, we can punt that to a future time to get this feature in.

twsouthwick: As far as the `IsComplete` it was initially tracked in the state file, but was removed as it complicated some E2E experiences

oteinone: Hi,

I can look at this. I was thinking along the same lines but maybe with
```
public interface IUpgradeContext
{
  string GetPropertyValue(string propertyName);

  void SetPropertyValue(string propertyName, string value, bool persistent);
}
```
So that the context can specify whether the property is worth storing in the file.

twsouthwick: That's a reasonable API. Good idea!

oteinone: @twsouthwick Pushed the requested changes, can you take a look?

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IUpgradeContext.cs
```diff
@@ -30,5 +30,26 @@ public interface IUpgradeContext : IDisposable
...
+        /// <param name="propertyName">Name identifier for the property.</param>
+        /// <param name="value">The property value to store.</param>
+        /// <param name="persistent">Whether to persist the value when the program terminates.</param>
+        void SetPropertyValue(string propertyName, string value, bool persistent);
+
+        /// <summary>
+        /// Returns persistent properties for storage.
+        /// </summary>
+        /// <returns>All defined persistent properties.</returns>
+        Dictionary<string, string> GetPersistentProperties();
```
twsouthwick, line 53: Let's expose this as an `IEnumerable<KeyValuePair<string,string>>`.

With 3 methods here, it may make sense to  expose it as a new interface

```cs
public interface IContextProperties:IEnumerable<KeyValuePair<string,string>>
{
  GetValue(...)
  SetValue(....)
}
```

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Backup/BackupStep.cs
```diff
@@ -13,12 +13,13 @@ namespace Microsoft.DotNet.UpgradeAssistant.Steps.Backup
...
     {
         private const string FlagFileName = "upgrade.backup";
+        private const string BackupPropertyName = "BackupLocation";
 
         private readonly bool _skipBackup;
         private readonly IUserInput _userInput;
 
         private string? _projectDir;
-        private string? _backupPath;
+        private string? _defaultBackupPath;
```
twsouthwick, line 22: Can you remove this and just grab it from the context object?

twsouthwick (changes requested):
I'm liking this. A couple of comments but I like this direction. Thanks!

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Backup/BackupStep.cs
oteinone: Makes sense. I think the same treatment should be applied to _projectDir as well.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IUpgradeContext.cs
oteinone: Ok, the refactoring was a bit bigger than I expected. I separated IContextProperties as a separate singleton service. I relocated the implementation as it has nothing to do with MSBuild now. I decided against directly inheriting from IEnumerable because I wanted to give the method a name to explain why it only returns persistent properties.

Most classes aren't really interested about whether the values are persistent or not so there could (in principle) be another interface for this but I feel this is too much complexity for a very minor advantage.

Also, added unit tests.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Backup/BackupStep.cs
```diff
@@ -132,16 +135,19 @@ protected override async Task<UpgradeStepApplyResult> ApplyImplAsync(IUpgradeCon
 
         public override UpgradeStepInitializeResult Reset()
         {
-            _backupPath = null;
+            _properties.RemovePropertyValue(BackupPropertyName);
```
twsouthwick, line 138: If you want it to persist, you can remove the whole method. This method is (usually) called when you switch to a new project

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Backup/BackupStep.cs
```diff
@@ -37,11 +36,12 @@ public class BackupStep : UpgradeStep
             WellKnownStepIds.NextProjectStepId,
         };
 
-        public BackupStep(UpgradeOptions options, ILogger<BackupStep> logger, IUserInput userInput)
+        public BackupStep(UpgradeOptions options, ILogger<BackupStep> logger, IUserInput userInput, IUpgradeContextProperties properties)
```
twsouthwick, line 39: The properties should still be accessible off the context rather than a service that is injected.

tests/components/Microsoft.DotNet.UpgradeAssistant.Tests/UpdateContextPropertiesTests.cs
```diff
@@ -0,0 +1,62 @@
...
+using Autofac.Extras.Moq;
+using Microsoft.Extensions.Logging.Abstractions;
+using Xunit;
+
+namespace Microsoft.DotNet.UpgradeAssistant.Tests
+{
+    public class UpdateContextPropertiesTests
+    {
+        [Fact]
+        public async Task GetAndSetWork()
```
twsouthwick, line 19: You're getting some build warnings (errors as warnings are turned on) because this should be void.

twsouthwick (changes requested):
A couple of comments. Main one is that instead of injecting as a service, the properties are specific to a context so it should be passed around there. Most likely as a property off the IUpgradeContext so that's the main state being passed around.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IUpgradeContext.cs
twsouthwick: Sounds good to me. The only change would be to not have it a separate service as we want the properties scoped to a context which isn't something tracked by DI (we have a factory that will create it on demand)

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Backup/BackupStep.cs
oteinone: Ok, I misunderstood the purpose.

tests/components/Microsoft.DotNet.UpgradeAssistant.Tests/UpdateContextPropertiesTests.cs
oteinone: Argh, my bad. Would you consider turning off documentation requirements in the project if they are not followed? The warnings for this were hidden by over 200 documentation warnings.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Backup/BackupStep.cs
oteinone: This was a complete brain fart. It should be corrected now in the latest push.

tests/components/Microsoft.DotNet.UpgradeAssistant.Tests/UpdateContextPropertiesTests.cs
mjrousos: There's some sort of weirdness going on with VS and our editorconfig. I see this same thing (real warnings are drowned out by documentation warnings). The thing that's weird about it is that [we do have documentation warnings disabled](https://github.com/dotnet/upgrade-assistant/blob/main/.editorconfig#L262)!

What I usually do is build from the command line (msbuild /t:rebuild) before pushing commits because the documentation warnings don't show up when built outside of VS.

twsouthwick: I've noticed this as well but thought it was a preview thing.

It's probably due to editor config vs ruleset. We may want to just go back to rulesets.....

twsouthwick (approved):
LGTM! Thanks for the PR!

tests/components/Microsoft.DotNet.UpgradeAssistant.Tests/UpdateContextPropertiesTests.cs
oteinone: @mjrousos @twsouthwick I took a peek at editorconfig and to me it seems that rules SA1600 and SA1601 are accidentally written as CA1600 and CA1601.

mjrousos: Good catch, @oteinone! I can't believe I missed that! We've run into issues with preview VS's handling of .editorconfig before, so I figured this was just another instance of that type of problem that would need investigated at some point. I'll get a PR opened; it'll be nice to not have all those warnings!

PR #448, opened 2021-04-20
Use NuGet credential providers when querying for package info with the NuGet SDK

### LINKED ISSUE #359: Use users' credential providers for working with authenticated NuGet sources
Currently, Upgrade-Assistant uses NuGet.config files relevant to the project being upgraded to pull NuGet packages from the right feeds. But if any of the packages come from authenticated feeds, it can fail (with a FatalProtocolException; 401) due to not using available credential providers.

mjrousos:
Fixes #359


src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/NuGetReference.cs
```diff
@@ -7,6 +7,8 @@ namespace Microsoft.DotNet.UpgradeAssistant
 {
     public record NuGetReference(string Name, string Version)
     {
+        public bool IsPrerelease => Version.Contains("-");
```
twsouthwick, line 10: Can you add a comment as to why this is a good check for IsPrerelease? I'm assuming it's per semantic versioning, but it would be good to document the why.

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/NuGetCredentials.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Threading.Tasks;
+using NuGet.Common;
+using NuGet.Configuration;
+using NuGet.Credentials;
+using NuGet.Protocol;
+using NuGet.Protocol.Plugins;
+
+namespace Microsoft.DotNet.UpgradeAssistant.MSBuild
+{
+    public static class NuGetCredentials
```
twsouthwick, line 16: could this be something done via a factory pattern through DI?

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Analyzers/TargetCompatibilityReferenceAnalyzer.cs
```diff
@@ -84,17 +84,30 @@ public async Task<PackageAnalysisState> AnalyzeAsync(IProject project, PackageAn
...
 
             foreach (var newerPackage in latestMinorVersions)
             {
                 if (await _packageLoader.DoesPackageSupportTargetFrameworksAsync(newerPackage, targetFramework, token).ConfigureAwait(false))
                 {
                     _logger.LogDebug("Package {NuGetPackage} will work on {TargetFramework}", newerPackage, targetFramework);
-                    return newerPackage;
+
+                    // Only return a pre-release version if it's the only newer major version that supports the necessary TFM
+                    if (newerPackage.IsPrerelease)
```
twsouthwick, line 96: I'm assuming this is not directly related to the credential provider stuff. I'm fine with it in, just wanted to make sure I'm not missing  aconnection

sunandabalu: yeah looks like not related to the cred provider change but the change itself to consider prerelease looks fine.

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/NuGetCredentials.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        public static void ConfigureCredentialService(bool interactive, ILogger logger)
+        {
+            if (HttpHandlerResourceV3.CredentialService == null)
+            {
+                HttpHandlerResourceV3.CredentialService = new Lazy<ICredentialService>(
+                    () => new CredentialService(
+                        providers: new AsyncLazy<IEnumerable<ICredentialProvider>>(async () => await GetCredentialProvidersAsync(logger).ConfigureAwait(false)),
+                        nonInteractive: !interactive,
+                        handlesDefaultCredentials: PreviewFeatureSettings.DefaultCredentialsAfterCredentialProviders));
```
sunandabalu, line 26: is the defaultcredentials to make it work for non-interactive mode? Or does this feature not work with non-interactive mode?

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/NuGetCredentials.cs
```diff
@@ -0,0 +1,43 @@
...
+            }
+        }
+
+        public static async Task<IEnumerable<ICredentialProvider>> GetCredentialProvidersAsync(ILogger logger)
```
KSchlobohm, line 30: if user input never makes it to the credentials provider then does it matter if the input is interactive? Seems we should hardcode that to always nonInteractive or wire-up credentials and userInput

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Analyzers/TargetCompatibilityReferenceAnalyzer.cs
```diff
@@ -84,17 +84,30 @@ public async Task<PackageAnalysisState> AnalyzeAsync(IProject project, PackageAn
         private async Task<NuGetReference?> GetUpdatedPackageVersionAsync(NuGetReference packageReference, IEnumerable<TargetFrameworkMoniker> targetFramework, CancellationToken token)
         {
             var latestMinorVersions = await _packageLoader.GetNewerVersionsAsync(packageReference, true, token).ConfigureAwait(false);
+            NuGetReference? prereleaseCandidate = null;
 
             foreach (var newerPackage in latestMinorVersions)
```
KSchlobohm, line 89: Does this say we upgrade packages to the lowest version that supports the target TFM?

I thought we upgraded to the latest package always.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Analyzers/TargetCompatibilityReferenceAnalyzer.cs
KSchlobohm: Can we avoid this or make it a feature toggle? Some companies choose not to use pre-release software over concerns about support or viability since pre-release packages can be subject to change.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/NuGetReference.cs
mjrousos: Yes, semver (https://semver.org/spec/v2.0.0.html#spec-item-9) and a comment here is a good idea. I'll add one.

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/NuGetCredentials.cs
mjrousos: `HttpHandlerResourceV3.CredentialService` is, unfortunately, static so it seems like the simplest way to set it is with a static method. What would a DI factory pattern look like here?

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Analyzers/TargetCompatibilityReferenceAnalyzer.cs
mjrousos: Yea, this is unrelated. I just happened to notice while testing out my change that sometimes the latest minor/build version of a package is pre-release and we probably don't want to use those if there are other similar versions that also work on Core that aren't pre-release.

@KSchlobohm, I think this change is ok as it is because it should result in *less* pre-release packages being used. After this change, the only time pre-release packages will be used will be if there are no non-pre-release packages that support .NET Core. I think that's ok, but I see your point about enterprises not wanting to use pre-release. To mitigate that concern, I'll add a warning if we have to use a pre-release package version to get NetCore support.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Analyzers/TargetCompatibilityReferenceAnalyzer.cs
mjrousos: We upgrade to the latest minor/build version of the lowest major version that supports the target TFM. We do that to reduce breaking changes (by using the lowest major version) but maximize bug/security fixes (by using the highest minor version).

Also, after this PR, we don't use pre-release versions unless they're the only ones that support the target TFM.

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/NuGetCredentials.cs
mjrousos: If the user hasn't entered credentials for the credential provider recently, then interactive mode will be required.

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/NuGetCredentials.cs
mjrousos: The caller (`PackageLoader`) is using `UserInput` to determine what to pass for this. Considering this comment and Taylor's suggestion to get NuGetCredentials from DI, it may be better to just have this class be a DI-retrieved service that has the UserInput dependency directly.

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/NuGetCredentials.cs
mjrousos: After looking at the other comments I'm thinking it would be better to have this come from DI. I think I'll make a few small changes to turn it into an `IUpgradeStartup`.

mjrousos:
@twsouthwick @KSchlobohm Thanks for the reviews. Comments should be addressed now. PTAL. Thanks!

sunandabalu (approved):
LGTM after the DI change

twsouthwick (approved):

PR #449, opened 2021-04-20
Expand TargetFrameworkMoniker record to include parts

twsouthwick:
This change replaces a simple record with `Name` to be one that includes
the parts that matter: FrameworkName, Version, Platform, PlatformVersion

This will then be formatted as needed in the property `Name`.

### DIFF
--- a/src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ITargetFrameworkMonikerComparer.cs
+++ b/src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ITargetFrameworkMonikerComparer.cs
@@ -25,5 +25,7 @@ namespace Microsoft.DotNet.UpgradeAssistant
         /// <param name="result">A merged TFM. For example, passing <c>net5.0-windows</c> and <c>net6.0</c> will result in <c>net6.0-windows</c>.</param>
         /// <returns>Whether the merge was successful.</returns>
         bool TryMerge(TargetFrameworkMoniker tfm1, TargetFrameworkMoniker tfm2, [MaybeNullWhen(false)] out TargetFrameworkMoniker result);
+
+        bool TryParse(string input, [MaybeNullWhen(false)] out TargetFrameworkMoniker tfm);
     }
 }
--- a/src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/TargetFrameworkMoniker.cs
+++ b/src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/TargetFrameworkMoniker.cs
@@ -6,7 +6,7 @@ using System.Text;
 
 namespace Microsoft.DotNet.UpgradeAssistant
 {
-    public record TargetFrameworkMoniker(string Framework, Version FrameworkVersion) : IEquatable<TargetFrameworkMoniker>
+    public record TargetFrameworkMoniker
     {
         private const string NetStandard = "netstandard";
         private const string NetFramework = "netframework";
@@ -14,13 +14,13 @@ namespace Microsoft.DotNet.UpgradeAssistant
         private const string NetCoreApp = "netcoreapp";
 
 #pragma warning disable CA1707 // Identifiers should not contain underscores
-        public static readonly TargetFrameworkMoniker NetStandard10 = new TargetFrameworkMoniker(NetStandard, new Version(1, 0)).Normalize();
+        public static readonly TargetFrameworkMoniker NetStandard10 = new(NetStandard, new Version(1, 0));
         public static readonly TargetFrameworkMoniker NetStandard20 = NetStandard10 with { FrameworkVersion = new Version(2, 0) };
         public static readonly TargetFrameworkMoniker NetStandard21 = NetStandard20 with { FrameworkVersion = new Version(2, 1) };
         public static readonly TargetFrameworkMoniker NetCoreApp21 = new(NetCoreApp, new Version(2, 1));
         public static readonly TargetFrameworkMoniker NetCoreApp30 = NetCoreApp21 with { FrameworkVersion = new Version(3, 0) };
         public static readonly TargetFrameworkMoniker NetCoreApp31 = NetCoreApp30 with { FrameworkVersion = new Version(3, 1) };
-        public static readonly TargetFrameworkMoniker Net45 = new TargetFrameworkMoniker(Net, new Version(4, 5)).Normalize();
+        public static readonly TargetFrameworkMoniker Net45 = new(Net, new Version(4, 5));
         public static readonly TargetFrameworkMoniker Net46 = Net45 with { FrameworkVersion = new Version(4, 6) };
         public static readonly TargetFrameworkMoniker Net461 = Net45 with { FrameworkVersion = new Version(4, 6, 1) };
         public static readonly TargetFrameworkMoniker Net462 = Net45 with { FrameworkVersion = new Version(4, 6, 2) };
@@ -41,130 +41,20 @@ namespace Microsoft.DotNet.UpgradeAssistant
         }
 #pragma warning restore CA1034 // Nested types should not be visible
 
-        public string? Platform { get; init; }
+        private string? _platform;
+        private Version? _platformVersion;
+        private string _framework;
+        private Version _frameworkVersion;
 
-        public Version? PlatformVersion { get; init; }
-
-        public string Name => ToString();
-
-        public virtual bool Equals(TargetFrameworkMoniker? other)
+        public TargetFrameworkMoniker(string framework, Version frameworkVersion)
         {
-            if (other is null)
-            {
-                return false;
-            }
-
-            var @this = Normalize();
-            var tfm = other.Normalize();
-
-            return string.Equals(@this.Framework, tfm.Framework, StringComparison.OrdinalIgnoreCase)
-                && Equals(@this.FrameworkVersion, tfm.FrameworkVersion)
-                && string.Equals(@this.Platform, tfm.Platform, StringComparison.OrdinalIgnoreCase)
-                && Equals(@this.PlatformVersion, tfm.PlatformVersion);
+            _framework = framework;
+            _frameworkVersion = frameworkVersion;
         }
 
-        public override int GetHashCode()
-        {
-            var normalized = Normalize();
-            var hashcode = default(HashCode);
-
-            hashcode.Add(normalized.Framework, StringComparer.OrdinalIgnoreCase);
-            hashcode.Add(normalized.FrameworkVersion);
-            hashcode.Add(normalized.Platform, StringComparer.OrdinalIgnoreCase);
-            hashcode.Add(normalized.PlatformVersion);
-
-            return hashcode.ToHashCode();
-        }
-
-        public override string ToString()
-        {
-            var normalized = Normalize();
-            var sb = new StringBuilder();
-
-            sb.Append(normalized.Framework);
-            sb.Append(normalized.FrameworkVersion);
-
-            if (IsFramework)
-            {
-                sb.Replace(".", string.Empty);
-            }
-            else if (IsNet50OrAbove && normalized.Platform is not null)
-            {
-                sb.Append('-');
-                sb.Append(normalized.Platform);
-
-                if (PlatformVersion is not null)
-                {
-                    sb.Append(normalized.PlatformVersion);
-                }
-            }
-
-            return sb.ToString();
-        }
-
-        public bool IsFramework
+        public string Framework
         {
             get
-            {
-                if (Framework.Equals(Net, StringComparison.OrdinalIgnoreCase) && !Is50OrAbove)
-                {
-                    return true;
-                }
-
-                return Framework.ToUpperInvariant().Contains(NetFramework.ToUpperInvariant());
-            }
-        }
-
-        public bool IsNetStandard => Framework.ToUpperInvariant().Contains(NetStandard.ToUpperInvariant());
-
-        private bool IsNet50OrAbove => (IsNetCoreApp || Framework.Equals(Net, StringComparison.OrdinalIgnoreCase)) && Is50OrAbove;
-
-        private bool Is50OrAbove => FrameworkVersion.Major >= 5;
-
-        private bool IsNetCoreApp => Framework.ToUpperInvariant().Contains(NetCoreApp.ToUpperInvariant());
-
-        public bool IsNetCore => IsNetCoreApp || IsNet50OrAbove;
-
-        public bool IsWindows => Platform?.Equals(Platforms.Windows, StringComparison.OrdinalIgnoreCase) ?? false;
-
-        private bool IsNormalized { get; init; }
-
-        /// <summary>
-        /// Normalizes the names and versions used to ensure that comparisons will be accurate.
-        /// </summary>
-        /// <returns>A normalized instance. If the current instance is already normalized, it will be returned without any changes.</returns>
-        public TargetFrameworkMoniker Normalize()
-        {
-            if (IsNormalized)
-            {
-                return this;
-            }
-
-            return this with
-            {
-                Framework = GetNormalizedFramework(),
-                FrameworkVersion = NormalizeVersion(FrameworkVersion) ?? FrameworkVersion,
-                Platform = GetNormalizedPlatform(),
-                PlatformVersion = NormalizeVersion(PlatformVersion),
-                IsNormalized = true,
-            };
-
-            string? GetNormalizedPlatform()
-            {
-                if (IsWindows)
-                {
-                    return Platforms.Windows;
-                }
-
-                if (string.IsNullOrEmpty(Platform))
-                {
-                    return null;
-                }
-
-                return Platform;
-            }
-
-            string GetNormalizedFramework()
             {
                 if (IsFramework || IsNet50OrAbove)
                 {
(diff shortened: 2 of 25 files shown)


src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/TargetFrameworkMoniker.cs
```diff
@@ -2,25 +2,200 @@
...
+#pragma warning disable CA1707 // Identifiers should not contain underscores
+        public static readonly TargetFrameworkMoniker NetStandard10 = new(NetStandard, new Version(1, 0));
+        public static readonly TargetFrameworkMoniker NetStandard20 = NetStandard10 with { FrameworkVersion = new Version(2, 0) };
+        public static readonly TargetFrameworkMoniker NetStandard21 = NetStandard20 with { FrameworkVersion = new Version(2, 1) };
+        public static readonly TargetFrameworkMoniker NetCoreApp21 = new(NetCoreApp, new Version(2, 1));
+        public static readonly TargetFrameworkMoniker NetCoreApp30 = NetCoreApp21 with { FrameworkVersion = new Version(3, 0) };
+        public static readonly TargetFrameworkMoniker NetCoreApp31 = NetCoreApp30 with { FrameworkVersion = new Version(3, 1) };
+        public static readonly TargetFrameworkMoniker Net45 = new(Net, new Version(4, 5));
+        public static readonly TargetFrameworkMoniker Net46 = Net45 with { FrameworkVersion = new Version(4, 6) };
+        public static readonly TargetFrameworkMoniker Net461 = Net45 with { FrameworkVersion = new Version(4, 6, 1) };
```
KSchlobohm, line 25: nit: uses a different pattern than established above
```
NetStandard20 = Netstandard10
NetStandard21 = NetStandard20
```

vs

```
Net46 = Net45
Net461 = Net45
...
Net48 = Net45
```

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/TargetFrameworkMoniker.cs
```diff
@@ -2,25 +2,200 @@
...
 
-        private const string NetStandardNamePrefix = "netstandard";
+#pragma warning disable CA1707 // Identifiers should not contain underscores
+        public static readonly TargetFrameworkMoniker NetStandard10 = new(NetStandard, new Version(1, 0));
+        public static readonly TargetFrameworkMoniker NetStandard20 = NetStandard10 with { FrameworkVersion = new Version(2, 0) };
+        public static readonly TargetFrameworkMoniker NetStandard21 = NetStandard20 with { FrameworkVersion = new Version(2, 1) };
+        public static readonly TargetFrameworkMoniker NetCoreApp21 = new(NetCoreApp, new Version(2, 1));
+        public static readonly TargetFrameworkMoniker NetCoreApp30 = NetCoreApp21 with { FrameworkVersion = new Version(3, 0) };
+        public static readonly TargetFrameworkMoniker NetCoreApp31 = NetCoreApp30 with { FrameworkVersion = new Version(3, 1) };
+        public static readonly TargetFrameworkMoniker Net45 = new(Net, new Version(4, 5));
```
KSchlobohm, line 23: should be `new(NetFramework, new Version(4,5)`?

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/TargetFrameworkMoniker.cs
```diff
@@ -2,25 +2,200 @@
...
-        public override string ToString() => Name;
+        private readonly string? _platform;
+        private readonly Version? _platformVersion;
+        private readonly string _framework;
+        private readonly Version _frameworkVersion;
 
-        public bool IsFramework => !IsNetStandard && !IsNetCore;
+        public TargetFrameworkMoniker(string framework, Version frameworkVersion)
+        {
+            _framework = framework;
```
KSchlobohm, line 54: throw an exception if **framework** is not one of the valid frameworks?

this would prevent code that inadvertently sets the platform as part of the framework.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/TargetFrameworkMoniker.cs
```diff
@@ -2,25 +2,200 @@
...
+                if (_framework.Equals(Net, StringComparison.OrdinalIgnoreCase) && !Is50OrAbove)
+                {
+                    return true;
+                }
+
+                return _framework.ToUpperInvariant().Contains(NetFramework.ToUpperInvariant());
+            }
+        }
+
+        public bool IsNetStandard => _framework.ToUpperInvariant().Contains(NetStandard.ToUpperInvariant());
```
KSchlobohm, line 180: nit: should be `.Equals` now that framework is a distinct field

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/TargetFrameworkMoniker.cs
```diff
@@ -2,25 +2,200 @@
...
+
+        private bool IsNet50OrAbove => (IsNetCoreApp || _framework.Equals(Net, StringComparison.OrdinalIgnoreCase)) && Is50OrAbove;
+
+        private bool Is50OrAbove => _frameworkVersion.Major >= 5;
+
+        private bool IsNetCoreApp => _framework.ToUpperInvariant().Contains(NetCoreApp.ToUpperInvariant());
```
KSchlobohm, line 186: nit: should be `.Equals` now that framework is a distinct field

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/TargetFrameworkMoniker.cs
```diff
@@ -2,25 +2,200 @@
...
+                return _platform;
+            }
+
+            init => _platform = value;
+        }
+
+        public Version? PlatformVersion
+        {
+            get => NormalizeVersion(_platformVersion);
+            init => _platformVersion = value;
```
KSchlobohm, line 112: guard clauses for valid platforms

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/TargetFrameworkMoniker.cs
```diff
@@ -2,25 +2,200 @@
...
+
+                if (string.IsNullOrEmpty(_platform))
+                {
+                    return null;
+                }
+
+                return _platform;
+            }
+
+            init => _platform = value;
```
KSchlobohm, line 106: guard clauses for valid platforms

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/TargetFrameworkMoniker.cs
```diff
@@ -2,25 +2,200 @@
...
+                    return NetStandard;
+                }
+                else
+                {
+                    return Framework;
+                }
+            }
 
-        public bool IsNetCore => Name.StartsWith(NetPrefix, StringComparison.OrdinalIgnoreCase) && Name.Contains(".");
+            init => _framework = value;
```
KSchlobohm, line 80: guard clauses for valid frameworks

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/TargetFrameworkMoniker.cs
KSchlobohm, line 112: guard clauses for valid platform versions?

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/TargetFrameworkMoniker.cs
```diff
@@ -2,25 +2,200 @@
...
+        }
 
-        public bool IsWindows => Name.Contains("windows");
+        public Version FrameworkVersion
+        {
+            get => NormalizeVersion(_frameworkVersion) ?? _frameworkVersion;
+            init => _frameworkVersion = value;
```
KSchlobohm, line 86: guard clauses

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/NuGetTargetFrameworkMonikerComparer.cs
```diff
@@ -112,14 +112,17 @@ public bool TryMerge(TargetFrameworkMoniker tfm1, TargetFrameworkMoniker tfm2, [
                 _ => null,
             };
 
-            var newFramework = (platform, version) switch
+            var frameworkName = (maxFramework.Framework, maxFramework.Version) switch
             {
-                (string p, Version v) => new NuGetFramework(maxFramework.Framework, maxFramework.Version, p, v),
-                (string p, null) => new NuGetFramework(maxFramework.Framework, maxFramework.Version, p, new Version(0, 0, 0, 0)),
-                _ => new NuGetFramework(maxFramework.Framework, maxFramework.Version),
+                (".NETCoreApp", Version v) when v >= TargetFrameworkMoniker.Net50.FrameworkVersion => "net",
```
KSchlobohm, line 117: should use constants from `TargetFrameworkMoniker.cs`

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/NuGetTargetFrameworkMonikerComparer.cs
```diff
@@ -112,14 +112,17 @@ public bool TryMerge(TargetFrameworkMoniker tfm1, TargetFrameworkMoniker tfm2, [
...
+                (string f, _) => f
             };
 
-            result = new TargetFrameworkMoniker(newFramework.GetShortFolderName());
+            result = new TargetFrameworkMoniker(frameworkName, maxFramework.Version)
```
KSchlobohm, line 121: Not related to this change, but I think we missed a scenario in the `GetPlatform` method.

If you merge *net5.0-windows10.0.19041.0* and *net6.0-windows10.0.18362.0* 

I think you should get *net6.0-windows**10.0.19041.0***

src/components/Microsoft.DotNet.UpgradeAssistant/TargetFramework/WindowsSdkTargetFrameworkSelectorFilter.cs
```diff
@@ -35,12 +33,12 @@ private static bool TryGetMoniker(ITargetFrameworkSelectorFilterState updater, [
...
             if (updater.Components.HasFlag(ProjectComponents.WindowsDesktop) || updater.Project.OutputType == ProjectOutputType.WinExe)
             {
-                tfm = $"{current}{WindowsSuffix}";
+                tfm = current with { Platform = TargetFrameworkMoniker.Platforms.Windows };
 
                 if (updater.Components.HasFlag(ProjectComponents.WinRT))
                 {
                     // TODO: Default to this version to ensure everything is supported.
-                    tfm += "10.0.19041.0";
+                    tfm = tfm with { PlatformVersion = new Version(10, 0, 19041, 0) };
```
KSchlobohm, line 41: looks like a good candidate for a readonly static

tests/common/Microsoft.DotNet.UpgradeAssistant.Abstractions.Tests/TargetFrameworkMonikerTests.cs
```diff
@@ -0,0 +1,156 @@
...
+using static Microsoft.DotNet.UpgradeAssistant.TargetFrameworkMonikerParser;
+
+namespace Microsoft.DotNet.UpgradeAssistant.Abstractions.Tests
+{
+    public class TargetFrameworkMonikerTests
+    {
+        public static IEnumerable<object[]> EqualityData()
+        {
+            yield return new object[] { TargetFrameworkMoniker.Net45, TargetFrameworkMoniker.Net45 };
+            yield return new object[] { TargetFrameworkMoniker.Net45, TargetFrameworkMoniker.Net45 with { Framework = "NET" } };
```
KSchlobohm, line 17: why use "NET" here instead of the constant? were you testing case sensitivity?

tests/components/Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests/TargetFrameworkIdentifierTests.cs
```diff
@@ -1,76 +1,79 @@
...
+        [InlineData(NetCoreApp31, Net48, -1)]
+        [InlineData(NetCoreApp31, Net471, -1)]
+        [InlineData(NetCoreApp31, Net47, -1)]
+        [InlineData(NetCoreApp31, Net45, -1)]
+        [InlineData(Net48, NetCoreApp31, -1)]
+        [InlineData(Net471, NetCoreApp31, -1)]
+        [InlineData(Net471, Net50, -1)]
+        [InlineData(Net471, NetStandard20, -1)]
+        [InlineData(Net461, NetStandard20, -1)]
+        [InlineData(Net46, NetStandard20, -1)]
```
KSchlobohm, line 76: Add tests for platform? (might be handled by a different test)

tests/components/Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests/TargetFrameworkIdentifierTests.cs
```diff
@@ -80,23 +83,23 @@ public void TfmCompare(string target, string tfm, int expected)
...
-        [InlineData("net6.0", "net5.0", "net6.0")]
-        [InlineData("net5.0", "net462", "net5.0")]
-        [InlineData("net5.0", "net5.0", "netstandard2.0")]
-        [InlineData("netstandard2.0", "netstandard2.0", "netstandard2.0")]
-        [InlineData("netstandard2.1", "netstandard2.0", "netstandard2.1")]
+        [InlineData(Net60_Windows, Net50_Windows, Net60)]
+        [InlineData(Net60_Windows, Net60, Net50_Windows)]
+        [InlineData(Net60_Windows, Net50, Net60_Windows)]
+        [InlineData(Net60_Windows_10_0_5, Net50, Net60_Windows_10_0_5)]
+        [InlineData(Net60_Windows_10_0_5, Net50_Windows_10_0_5, Net60)]
```
KSchlobohm, line 93: Add test scenario `[InlineData(Net60_Windows_10_0_5, Net50_Windows_10_1_5, Net60_Windows_10_1_5)]`

KSchlobohm (changes requested):
Added some comments inline

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ITargetFrameworkMonikerComparer.cs
```diff
@@ -25,5 +25,7 @@ public interface ITargetFrameworkMonikerComparer : IComparer<TargetFrameworkMoni
         /// <param name="result">A merged TFM. For example, passing <c>net5.0-windows</c> and <c>net6.0</c> will result in <c>net6.0-windows</c>.</param>
         /// <returns>Whether the merge was successful.</returns>
         bool TryMerge(TargetFrameworkMoniker tfm1, TargetFrameworkMoniker tfm2, [MaybeNullWhen(false)] out TargetFrameworkMoniker result);
+
+        bool TryParse(string input, [MaybeNullWhen(false)] out TargetFrameworkMoniker tfm);
```
mjrousos, line 29: This seems like an odd method to have on a comparer interface. It doesn't really have anything to do with comparing and results in the comparer being needed in odd places. It would be great if it could just live on the TargetFrameworkMoniker type, but I know NuGet dependencies make that hard.

How much work would it be to write our own parser that doesn't have a NuGet dependency? It seems like it shouldn't be _that_ hard. If that doesn't seem like a good option, maybe create a dedicated TargetFrameworkMoniker factory interface for this?

mjrousos (approved):
A few questions and comments inline, but looks good overall.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/TargetFrameworkMoniker.cs
twsouthwick: Cool. I've updated it.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/TargetFrameworkMoniker.cs
twsouthwick: Both work. The NetFramework is possible from the NuGet parsing and so it handles that fine. For clarity I'll update it.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/TargetFrameworkMoniker.cs
twsouthwick: We just end up ignoring it in that case. I've added some tests for that now.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/TargetFrameworkMoniker.cs
twsouthwick: This is to potentially match `.NETStandard`, as well as `netstandard` - both show up in NuGet parsing

twsouthwick: I've added a comment to that effect.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/TargetFrameworkMoniker.cs
twsouthwick: I'm not sure if we need that - it also keeps changing and I don't want to have to track that unless we *need* to know a valid option

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/TargetFrameworkMoniker.cs
twsouthwick: What's the need to know if it's a valid framework? It's a more closed set that platforms for sure, but I'd rather be consistent in that choice.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/TargetFrameworkMoniker.cs
twsouthwick: That's an open set as well (like platforms) that I'm not certain of the value of a guard clause there

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/NuGetTargetFrameworkMonikerComparer.cs
twsouthwick: This isn't necessary anymore as those string comparisons are handled in TargetFrameworkMoniker now. I've removed that.

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/NuGetTargetFrameworkMonikerComparer.cs
twsouthwick: I believe it does. Here's the relevant test:

https://github.com/dotnet/upgrade-assistant/blob/cb835e9c4059b0c4f1ffa350e84b905b99ea8486/tests/components/Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests/TargetFrameworkIdentifierTests.cs#L91

tests/common/Microsoft.DotNet.UpgradeAssistant.Abstractions.Tests/TargetFrameworkMonikerTests.cs
twsouthwick: Case sensitivity and framework name normalization. I've add a comment to what the value is.

tests/components/Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests/TargetFrameworkIdentifierTests.cs
twsouthwick: That's not the expected result, though. I'd expect `Net60_Windows_10_1_5` as that's what both provided ones are

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/TargetFrameworkMoniker.cs
```diff
@@ -2,25 +2,200 @@
...
+        public static readonly TargetFrameworkMoniker Net50_Windows = Net50 with { Platform = Platforms.Windows };
+        public static readonly TargetFrameworkMoniker Net50_Linux = Net50 with { Platform = Platforms.Linux };
+        public static readonly TargetFrameworkMoniker Net60 = Net50 with { FrameworkVersion = new Version(6, 0) };
+        public static readonly TargetFrameworkMoniker Net60_Linux = Net60 with { Platform = Platforms.Linux };
+        public static readonly TargetFrameworkMoniker Net60_Windows = Net60 with { Platform = Platforms.Windows };
+#pragma warning restore CA1707 // Identifiers should not contain underscores
 
-        private const string NetPrefix = "net";
+#pragma warning disable CA1034 // Nested types should not be visible
+        public static class Platforms
```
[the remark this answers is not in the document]
twsouthwick: Since it's an open set, I want the parsing of the TFM to drive what platforms are available. I've created a class for the Frameworks to address the asymmetry

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/TargetFrameworkMoniker.cs
```diff
@@ -2,25 +2,200 @@
...
+                {
+                    return NetCoreApp;
+                }
+                else if (IsNetStandard)
+                {
+                    return NetStandard;
+                }
+                else
+                {
+                    return Framework;
```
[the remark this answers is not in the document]
twsouthwick: That's a bug. Fixed.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/TargetFrameworkMoniker.cs
```diff
@@ -2,25 +2,200 @@
...
+                    return true;
+                }
+
+                return _framework.ToUpperInvariant().Contains(NetFramework.ToUpperInvariant());
+            }
+        }
+
+        public bool IsNetStandard => _framework.ToUpperInvariant().Contains(NetStandard.ToUpperInvariant());
+
+        private bool IsNet50OrAbove => (IsNetCoreApp || _framework.Equals(Net, StringComparison.OrdinalIgnoreCase)) && Is50OrAbove;
```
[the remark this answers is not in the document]
twsouthwick: It does and is valid (it was changed at some point during the previews, so it's out there and NuGet still uses it for the name weirdly enough)

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ITargetFrameworkMonikerComparer.cs
twsouthwick: I agree - I'm thinking we can fold a lot of the stuff the comparer was doing into the TargetFrameworkMoniker itself. I'll leave that for a separate PR to address though

twsouthwick: Here's an issue to track this: https://github.com/dotnet/upgrade-assistant/issues/453

tests/components/Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests/TargetFrameworkIdentifierTests.cs
twsouthwick: Swapped them around and added the test

src/components/Microsoft.DotNet.UpgradeAssistant/TargetFramework/WindowsSdkTargetFrameworkSelectorFilter.cs
```diff
@@ -35,12 +33,12 @@ private static bool TryGetMoniker(ITargetFrameworkSelectorFilterState updater, [
...
             if (updater.Components.HasFlag(ProjectComponents.WindowsDesktop) || updater.Project.OutputType == ProjectOutputType.WinExe)
             {
-                tfm = $"{current}{WindowsSuffix}";
+                tfm = current with { Platform = TargetFrameworkMoniker.Platforms.Windows };
 
                 if (updater.Components.HasFlag(ProjectComponents.WinRT))
                 {
                     // TODO: Default to this version to ensure everything is supported.
-                    tfm += "10.0.19041.0";
+                    tfm = tfm with { PlatformVersion = TargetFrameworkMoniker.Net50_Windows_10_0_19041_0.PlatformVersion };
```
KSchlobohm, line 41: i guess this is what i asked for :D but my intent was that your comment declares this is a default. seems like we should have a `PlatformVersion = TargetFrameworkMoniker.Net50_DefaultWindowsPlatform`

KSchlobohm (approved):
looks good to me, added small nit feedback about default var name

src/components/Microsoft.DotNet.UpgradeAssistant/TargetFramework/WindowsSdkTargetFrameworkSelectorFilter.cs
twsouthwick: I'll plan on leaving this as-is as it's not really a "DefaultWindowsPlatform" but a default we're using for this specific usage due to simplicity

PR #455, opened 2021-04-22
Add a RazorUpdater that applies code fixes to Razor docs

### LINKED ISSUE #367: As a user, my cshtml files are updated to use ASP.NET Core types instead of ASP.NET ones, where possible

mjrousos:
Fixes #367


src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/IsExternalInit.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace System.Runtime.CompilerServices
+{
+    /// <summary>
+    /// Workaround to enable init properties and record types while targeting netstandard2.0
+    /// </summary>
+    internal static class IsExternalInit
```
twsouthwick, line 9: If you merge from main, you can follow the pattern I introduced for the nullability attributes to include this file automatically if needed.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/MappedSubText.cs
```diff
@@ -0,0 +1,85 @@
...
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Text;
+
+namespace Microsoft.DotNet.UpgradeAssistant.Steps.Razor
+{
+    public sealed record MappedSubText(SourceText Text, string FilePath, int StartingLine)
+    {
+        public string SourceLocation => $"{FilePath}@{StartingLine}";
+
+        public static async Task<IEnumerable<MappedSubText>> GetMappedSubTextsAsync(Document document, CancellationToken token)
```
twsouthwick, line 20: I'm fine with this, but I'm still trying to figure out when to return a `Task<IEnumerable<>>` vs an `IAsyncEnumerable<>`. How do you think about that?

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/MappedSubText.cs
```diff
@@ -0,0 +1,85 @@
...
+
+            return ret;
+        }
+
+        public bool Equals(MappedSubText other) =>
+            other != null &&
+            Text.ContentEquals(other.Text) &&
+            SourceLocation.Equals(other.SourceLocation, StringComparison.Ordinal);
+
+        public override int GetHashCode()
```
twsouthwick, line 76: You can bring in Microsoft.Bcl.Hashcode to use the System.Hashcode struct here

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorSourceUpdater.cs
```diff
@@ -48,59 +56,332 @@ public async Task<IUpdaterResult> IsApplicableAsync(IUpgradeContext context, Imm
...
+            var diagnosticsByFile = mappedDiagnostics.GroupBy(d => d.Location.GetMappedLineSpan().Path);
+            foreach (var diagnosticsGroup in diagnosticsByFile)
+            {
+                _logger.LogInformation("  {DiagnosticsCount} diagnostics need fixed in {FilePath}", diagnosticsGroup.Count(), diagnosticsGroup.Key);
+            }
+
+            return new FileUpdaterResult(mappedDiagnostics.Any(), diagnosticsByFile.Select(g => g.Key));
+        }
+
+        public async Task<IUpdaterResult> ApplyAsync(IUpgradeContext context, ImmutableArray<RazorCodeDocument> inputs, CancellationToken token)
```
twsouthwick, line 74: Can you split this method up into smaller methods for each step it's doing?

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorSourceUpdater.cs
```diff
@@ -48,59 +56,332 @@ public async Task<IUpdaterResult> IsApplicableAsync(IUpgradeContext context, Imm
...
+            {
+                _logger.LogWarning("Code fix could not be applied for {DiagnosticId}", diagnostic.Id);
+                return document.Project;
+            }
 
-            _logger.LogDebug("Identified {DiagnosticCount} fixable diagnostics in Razor files in project {ProjectName}", _diagnostics.Count(), project.Name);
+            return applyOperation.ChangedSolution.GetProject(document.Project.Id) ?? throw new InvalidOperationException("Updated project not found");
         }
 
         private static IAsyncEnumerable<DiagnosticAnalyzer> GetApplicableAnalyzersAsync(IEnumerable<DiagnosticAnalyzer> analyzers, IProject project) =>
```
twsouthwick, line 387: This class is very large. Can it be broken up into smaller services? I don't see any tests in this PR (were they already merged in?), but the smaller services will make the testing much easier

twsouthwick (changes requested):

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/IsExternalInit.cs
mjrousos: Sounds good. Will do.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/MappedSubText.cs
mjrousos: Awesome. I didn't realize that package existed to get System.Hashcode in netstandard2.0. I'll use it.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/MappedSubText.cs
mjrousos: Yea, that can be a touch decision. In this case, there are two reasons I prefer `Task<IEnumerable>`:

1. The way the enumerable is produced: In this case, all async work is done prior to creating the enumerable items so IAsyncEnumerable wouldn't provide any benefit.
2. The way the enumerable is consumed: The caller in this case is calling the method twice and wants to create lookups from the enumerables. This is a lot simple with IEnumerable than with IAsyncEnumerable.

So, in this scenario, at least, I think `Task<IEnumerable>` is the easy choice since it makes the caller simpler and the actual enumeration part of this API doesn't await anything. If there were awaits while enumerating, I think `IAsyncEnumerable` would make more sense.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorSourceUpdater.cs
mjrousos: Yea, that's a good idea. I'm adding tests in a follow-up PR, but you're right that will be simpler if this is broken up. I'll refactor it a bit.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorSourceUpdater.cs
mjrousos: Good idea.,

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorSourceUpdater.cs
```diff
@@ -38,7 +45,8 @@ public RazorSourceUpdater(IEnumerable<DiagnosticAnalyzer> analyzers, IEnumerable
             _analyzers = analyzers?.OrderBy(a => a.SupportedDiagnostics.First().Id) ?? throw new ArgumentNullException(nameof(analyzers));
             _codeFixProviders = codeFixProviders?.OrderBy(c => c.FixableDiagnosticIds.First()) ?? throw new ArgumentNullException(nameof(codeFixProviders));
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
-            _diagnostics = Enumerable.Empty<Diagnostic>();
+            _chunker = new CharacterChunker();
+            _differ = new Differ();
```
twsouthwick, line 49: This is an implementation of IDiffer that could be passed in via DI to simplify testing.

twsouthwick (approved):
I won't block this on making it smaller as you're adding tests in a later one and that'll help drive how to break things up.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorSourceUpdater.cs
```diff
@@ -48,59 +56,332 @@ public async Task<IUpdaterResult> IsApplicableAsync(IUpgradeContext context, Imm
...
+            var project = GetProjectWithGeneratedCode(context.CurrentProject.Required(), inputs);
+            var originalProject = project;
+            var generatedFilePaths = inputs.Select(GetGeneratedFilePath);
+
+            // Regenerating diagnostics is slow for large projects, but is necessary in between fixing multiple diagnostics
+            // in a single file. To minimize the number of time diagnostics are gathered, fix one diagnostic each
+            // from multiple files before regenerating diagnostics.
+            var diagnostics = await GetDiagnosticsFromTargetFilesAsync(project, context, generatedFilePaths, new LocationAndIDComparer(false), token).ConfigureAwait(false);
+            var diagnosticsCount = diagnostics.Count();
+            while (diagnosticsCount > 0)
```
KSchlobohm, line 90: what-if we can't fix the diagnostic? seems we might try, and fail, then need a max-retry circuit breaker

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorSourceUpdater.cs
```diff
@@ -48,59 +56,332 @@ public async Task<IUpdaterResult> IsApplicableAsync(IUpgradeContext context, Imm
...
+                    if (doc is null)
+                    {
+                        continue;
+                    }
+
+                    // Apply the code fix in the generated source
+                    project = await TryFixDiagnosticAsync(diagnostic, doc).ConfigureAwait(false);
+                }
+
+                diagnostics = await GetDiagnosticsFromTargetFilesAsync(project, context, generatedFilePaths, new LocationAndIDComparer(false), token).ConfigureAwait(false);
```
KSchlobohm, line 105: looks like a do... while candidate

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorSourceUpdater.cs
```diff
@@ -48,59 +56,332 @@ public async Task<IUpdaterResult> IsApplicableAsync(IUpgradeContext context, Imm
...
+                var newDiagnosticCount = diagnostics.Count();
+                if (diagnosticsCount == newDiagnosticCount)
+                {
+                    _logger.LogWarning("Diagnostics could not be fixed as expected. This may be caused by the project being in a bad state (did NuGet packages restore correctly?) or by errors in analyzers or code fix providers related to diagnosticsL {DiagnosticIds}.", string.Join(", ", diagnostics.Select(d => d.Id).Distinct()));
```
KSchlobohm, line 109: `diagnosticsL`

KSchlobohm, line 109: is there a call to action here? seems like we want to ask the user to confirm that their packages were restored and maybe other things.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorSourceUpdater.cs
```diff
@@ -48,59 +56,332 @@ public async Task<IUpdaterResult> IsApplicableAsync(IUpgradeContext context, Imm
...
+                    break;
+                }
+                else
+                {
+                    diagnosticsCount = newDiagnosticCount;
+                }
+            }
+
+            // Update cshtml based on changes made to generated source code
+            _logger.LogDebug("Updating Razor documents based on changes made by code fix providers");
```
KSchlobohm, line 119: assuming an error pops up, it would be helpful to know which diagnostic we're currently fixing

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorSourceUpdater.cs
```diff
@@ -48,59 +56,332 @@ public async Task<IUpdaterResult> IsApplicableAsync(IUpgradeContext context, Imm
...
 
-            return new FileUpdaterResult(_diagnostics.Any(), _diagnostics.Select(d => d.Location.GetMappedLineSpan().Path).Distinct());
+            return new FileUpdaterResult(true, textReplacements.Select(r => r.OriginalText.FilePath).Distinct());
         }
 
-        public Task<IUpdaterResult> ApplyAsync(IUpgradeContext context, ImmutableArray<RazorCodeDocument> inputs, CancellationToken token)
+        private async Task<IList<TextReplacement>> GetReplacements(Project originalProject, IEnumerable<Document> updatedDocuments, CancellationToken token)
         {
-            throw new NotImplementedException();
+            List<TextReplacement> replacements = new List<TextReplacement>();
```
KSchlobohm, line 134: nit: use var

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorSourceUpdater.cs
```diff
@@ -48,59 +56,332 @@ public async Task<IUpdaterResult> IsApplicableAsync(IUpgradeContext context, Imm
...
+
+                    // If the updated group is empty, then the original elements all pair with empty source text
+                    else if (!updatedGroup.Any())
+                    {
+                        AddReplacementCandidates(replacements, originalGroup.Select(m => new TextReplacement(m, SourceText.From(string.Empty))));
+                    }
+
+                    // This is the tricky one. If there are less updated code blocks than original code blocks, it will be necesary to guess which original code blocks
+                    // pair with the remaining updated code blocks based on text similarity.
+                    else
```
KSchlobohm, line 168: how does this scenario come up? is this intended to handle having more than one code fixer per diagnostic?

If we only have one code fixer per diagnostic, and we're running one diagnostic at a time, then I would not expect overlapping code changes.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorSourceUpdater.cs
```diff
@@ -48,59 +56,332 @@ public async Task<IUpdaterResult> IsApplicableAsync(IUpgradeContext context, Imm
...
+            {
+                // StreamSourceDoc.Lines is 0-based but line directives (as used in MappedSubText) are 1-based,
+                // so subtract one from i.
+                offset += razorDoc.Source.Lines.GetLineLength(i - 1);
+            }
+
+            return offset;
+        }
 
+        private static string GetGeneratedFilePath(RazorCodeDocument doc) => $"{doc.Source.FilePath}.cs";
```
KSchlobohm, line 305: are the compiled razor files written to disk? if so, where?

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/TextReplacement.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using Microsoft.CodeAnalysis.Text;
+
+namespace Microsoft.DotNet.UpgradeAssistant.Steps.Razor
+{
+    public class TextReplacement : IEquatable<TextReplacement>
```
KSchlobohm, line 9: based on the PR i think the reason we use this class is to store a list of replacements, without performing the replacement, as we analyze because replacing would change the location of n+1 diagnostic.

seems like a comment on the class could help clarify why it was necessary. when i first saw it i wondered why we didn't just do the replacement immediately.

KSchlobohm (approved):
Looks good to me overall. I am  interested in the feedback about unit tests and breaking apart the large class.

KSchlobohm:
relates to #422

KSchlobohm:
needs a changelog

sunandabalu (approved):
@mjrousos Could you add some integration tests around the cshtml source updater scenario?

mjrousos:
Yep, I'll update our integration tests to make sure they hit this.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorSourceUpdater.cs
mjrousos: Good point. I'll get it from there.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorSourceUpdater.cs
mjrousos: At the end of this loop we check whether the count of diagnostics is decreasing. If not, we break out of the loop with a warning.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorSourceUpdater.cs
mjrousos: Fixed.

mjrousos: Not really, it's just a generic warning that the diagnostics weren't fixed as expected so something went wrong. It's hard to know what, though. Could be that packages weren't restored but could also be a bug in the analyzer or code fix provider.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorSourceUpdater.cs
mjrousos: This is fixing lots of different diagnostics at once, though, so it's hard to isolate one.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorSourceUpdater.cs
mjrousos: Good catch. Fixed.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorSourceUpdater.cs
mjrousos: The scenario is that some document has X pieces of C# that are mapped back to cshtml. After applying code fixers (could be one, could be more), we end up with X-1 pieces of C# (because one was removed entirely). The challenge is that we don't know which one was removed so it's tricky to map the updated C# to the original. I actually improved this logic in a newer commit, but it's still solving that problem.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorSourceUpdater.cs
mjrousos: No, only in memory.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/TextReplacement.cs
mjrousos: I'll add a comment.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorSourceUpdater.cs
sunandabalu: +1 on making this class smaller.

mjrousos:
Also, changelog entry added.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorSourceUpdater.cs
mjrousos: Yep, the class has been shrunk :)

PR #466, opened 2021-04-27
Fix backup step to back up different projects to different sub-dirs

### LINKED ISSUE #462: Backup step doesn't work for multi-project solutions
### To Reproduce

1. Run Upgrade Assistant on a solution with multiple entry points. Choose one of the entry points and upgrade using all defaults.
2. Run Upgrade Assistant a second time on the sample solution and choose a different entry point. **The tool will _not_ back up this project** because it stores the backup location for the solution from the previous run and believes backup is complete (since a project *is* backed up there).

mjrousos:
Fixes #462


tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Backup.Tests/BackupStepTests.cs
```diff
@@ -0,0 +1,134 @@
...
+        [InlineData("B\\B.sln", "B\\B.csproj", true, "B.backup\\B.0")]
+        public async Task InitializeTests(string inputPath, string projectPath, bool backupComplete, string expectedBackupPath)
+        {
+            // Arrange
+            inputPath = Path.GetFullPath(Path.Combine("TestAssets", inputPath));
+            projectPath = Path.GetFullPath(Path.Combine("TestAssets", projectPath));
+            expectedBackupPath = Path.GetFullPath(Path.Combine("TestAssets", expectedBackupPath));
+
+            using var mock = GetMock(GetDefaultNonInteractiveOptions(), inputPath, projectPath);
+            var context = mock.Mock<IUpgradeContext>();
```
twsouthwick, line 33: nit: `mock.Mock<>` should be used for objects that will be injected via autofac in AutoMock. Objects that are passed as args should be created with `new Mock<>`. It's a subtl difference, but has caught me off guard many times.

The main scenario I've seen is where I want multiple projects, but subsequent calls to `mock.Mock<>` will mock the same instance, while `new Mock<>` will create new ones.

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Backup.Tests/BackupStepTests.cs
```diff
@@ -0,0 +1,134 @@
...
+                SkipBackup = false,
+            };
+
+        private static AutoMock GetMock(UpgradeOptions options, string inputPath, string? projectPath)
+        {
+            var mock = AutoMock.GetLoose(cfg =>
+            {
+                cfg.RegisterInstance(options);
+                cfg.RegisterInstance<IUpgradeContextProperties>(new UpgradeContextProperties());
+                cfg.RegisterType<NonInteractiveUserInput>().As<IUserInput>();
```
twsouthwick, line 116: Do you need to use the `NonInteractiveUserInput` or just mock `IUserInput` to return that it's non-interactive?

twsouthwick (approved):
A couple of comments but LGTM

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Backup.Tests/BackupStepTests.cs
mjrousos: Good point.

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Backup.Tests/BackupStepTests.cs
mjrousos: I'm new to AutoMock, so I could definitely be getting a best practice wrong here, but I feel like `mock.Mock<>` is ok here because I only have the one context (and one project that is used by the context) that the tests retrieve.

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Backup.Tests/BackupStepTests.cs
mjrousos: Actually, I just checked. BackupStep uses ChooseAsync and AskUserAsync (rather than IsInteractive), so it will be simpler just to use our existing NonInteractiveUserInput type here.

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Backup.Tests/BackupStepTests.cs
mjrousos: Ah, I just noticed this comment is in the test methods (rather than in GetMock). I can change this to use `new Mock<>` for that.

sunandabalu (approved):
Thanks for adding some tests @mjrousos

PR #467, opened 2021-04-28
Additional sample upgrade step

mjrousos:
A customer was asking what an upgrade step would look like that just performed simple text replacement, so this adds an additional sample Upgrade Step to our extension samples demonstrating that.

Also, fixes a bug in `ProjectItemMatcher`.


src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ProjectItemMatcher.cs
```diff
@@ -33,12 +33,19 @@ public ProjectItemMatcher(string match)
...
+            if (input is null)
+            {
+                throw new ArgumentNullException(nameof(input));
+            }
+
+            return _match switch
             {
                 Regex regex => regex.IsMatch(Path.GetFileName(input)),
-                string str => str.EndsWith(input, StringComparison.OrdinalIgnoreCase),
+                string str => input.EndsWith(str, StringComparison.OrdinalIgnoreCase),
```
mjrousos, line 45: This is the one functional change to Upgrade Assistant in this PR. Based on my understanding of the item matcher, I think this check was backwards. I didn't write this, though, so if anyone knows why it was the way it was, let me know!

samples/FindReplaceStepSample/FindReplaceStep/FindReplaceExtensionServiceProvider.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// service collection to register services in and the extension's configuration file.</param>
+        /// <returns>The service collection updated with services Upgrade Assistant should use.</returns>
+        public IServiceCollection AddServices(ExtensionServiceConfiguration serviceConfiguration)
+        {
+            if (serviceConfiguration is null)
+            {
+                throw new ArgumentNullException(nameof(serviceConfiguration));
+            }
+
+            return serviceConfiguration.ServiceCollection.AddScoped<UpgradeStep, FindReplaceUpgradeStep>();
```
twsouthwick, line 36: Isn't there an `AddStep` extension method that could be used?

samples/FindReplaceStepSample/FindReplaceStep/FindReplaceStep.csproj
```diff
@@ -0,0 +1,20 @@
...
+       so that it will be present in the extension's binary output. -->
+  <ItemGroup>
+    <Content Include="ExtensionManifest.json">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </Content>
+  </ItemGroup>
+  
+  <ItemGroup>
+    <PackageReference Include="Microsoft.DotNet.UpgradeAssistant.Abstractions" Version="0.2.222701" />
+    <PackageReference Include="Microsoft.DotNet.UpgradeAssistant.Extensions" Version="0.2.222701" />
```
twsouthwick, line 17: We should look at combining these....

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ProjectItemMatcher.cs
twsouthwick: I wrote this :) not sure, but this seems right. I'm surprised it didn't change anything else!

We should add some tests here

twsouthwick (approved):
A couple of comments, but LGTM

It may be a good idea to provide some base classes of upgrade steps that would simplify common tasks like this that can then be extended in custom code.

Another option would be to provide steps that an extension could customize in a config file and give it a new id. We would then add that step into the pipeline. I'll right up a proposal for that so we can track it.

samples/FindReplaceStepSample/FindReplaceStep/FindReplaceExtensionServiceProvider.cs
mjrousos: There is; good catch.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ProjectItemMatcher.cs
mjrousos: Yea, I was surprised too! But all the tests still pass. I took a look and it looks like the other places where we call with a string, we were comparing the full string.

PR #469, opened 2021-04-28
Enable LTS/Current/Preview selection

### LINKED ISSUE #41: As a user, I can choose to migrate to either LTS, Current, or Preview
Added Preview to consider supporting the preview release for the TFM (in this case .NET 6).

twsouthwick:
This change does a couple of things to enable LTS/Current/Preview selection:

- Adds a command line argument to allow selection
- Maps the selected support model to identify the TFM that is needed
- Updates the PackageLoader to limit searches to results that match the TFMs in a project

Fixes #41


CHANGELOG.md
```diff
@@ -3,10 +3,15 @@ All notable changes to the .NET Upgrade Assistant will be documented in this fil
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
+## Future
+
+### Added
+- A new command line option (`--target-tfm-support` to select the support model of LTS/Preview/Current that is desired [#41](https://github.com/dotnet/upgrade-assistant/issues/41)
```
mjrousos, line 9: Nit-pick - I think our pattern has been to link to the PR rather than the issue.

README.md
```diff
@@ -82,29 +82,20 @@ Arguments:
...
-                                               globbing patterns such as '*' for match.
-  -v, --verbose                                Enable verbose diagnostics
-  --non-interactive                            Automatically select each first option in
-                                               non-interactive mode.
-  --non-interactive-wait                       Wait the supplied seconds before moving
-  <non-interactive-wait>                       on to the next option in non-interactive
-                                               mode.
-  --version                                    Show version information
-  -?, -h, --help                               Show help and usage information
+  --skip-backup                                  Disables backing up the project. This is not recommended unless the project is in source control since this tool
```
mjrousos, line 85: Another nit-pick - it might be good to update this without making the lines quite as long. This will look fine in a maximized window but may require scrolling if the person reading the docs doesn't have their browser maximized.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Analyzers/TargetCompatibilityReferenceAnalyzer.cs
```diff
@@ -83,16 +83,16 @@ public async Task<PackageAnalysisState> AnalyzeAsync(IProject project, PackageAn
...
+        private async Task<NuGetReference?> GetUpdatedPackageVersionAsync(NuGetReference packageReference, IEnumerable<TargetFrameworkMoniker> tfms, CancellationToken token)
         {
-            var latestMinorVersions = await _packageLoader.GetNewerVersionsAsync(packageReference, true, token).ConfigureAwait(false);
+            var latestMinorVersions = await _packageLoader.GetNewerVersionsAsync(packageReference, tfms, true, token).ConfigureAwait(false);
             NuGetReference? prereleaseCandidate = null;
 
             foreach (var newerPackage in latestMinorVersions)
             {
-                if (await _packageLoader.DoesPackageSupportTargetFrameworksAsync(newerPackage, targetFramework, token).ConfigureAwait(false))
+                if (await _packageLoader.DoesPackageSupportTargetFrameworksAsync(newerPackage, tfms, token).ConfigureAwait(false))
```
mjrousos, line 93: Can this `if` go away now that `GetNewerVersionsAsync` filters for TFMs?

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/PackageLoader.cs
```diff
@@ -96,57 +109,64 @@ public async Task<IEnumerable<NuGetReference>> GetNewerVersionsAsync(NuGetRefere
...
-                {
-                    var metadata = await GetSourceRepository(source).GetResourceAsync<PackageMetadataResource>(token).ConfigureAwait(false);
-                    var searchResults = await CallWithRetryAsync(() => metadata.GetMetadataAsync(packageName, includePrerelease: includePreRelease, includeUnlisted: false, _cache, _nugetLogger, token)).ConfigureAwait(false);
-                    var highestVersionResult = searchResults.Select(r => r.Identity.Version).Max(v => v);
-                    if (highestVersionResult is null)
-                    {
-                        continue;
-                    }
+                var max = results
+                    .GroupBy(r => r.Identity.Version)
```
mjrousos, line 130: Shouldn't we be grouping by Version.Major for the `latestMinorAndBuildOnly` path?

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/PackageLoader.cs
```diff
@@ -96,57 +109,64 @@ public async Task<IEnumerable<NuGetReference>> GetNewerVersionsAsync(NuGetRefere
...
-                try
-                {
-                    var metadata = await GetSourceRepository(source).GetResourceAsync<PackageMetadataResource>(token).ConfigureAwait(false);
-                    var searchResults = await CallWithRetryAsync(() => metadata.GetMetadataAsync(packageName, includePrerelease: includePreRelease, includeUnlisted: false, _cache, _nugetLogger, token)).ConfigureAwait(false);
-                    var highestVersionResult = searchResults.Select(r => r.Identity.Version).Max(v => v);
-                    if (highestVersionResult is null)
-                    {
-                        continue;
-                    }
+                var max = results
```
mjrousos, line 129: This code path seems incomplete. We get a max version but don't do anything with it. As far as I can tell, setting `latestMinorAndBuildOnly` to true won't change anything here.

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/PackageLoader.cs
```diff
@@ -96,57 +109,64 @@ public async Task<IEnumerable<NuGetReference>> GetNewerVersionsAsync(NuGetRefere
...
+                if (results is null)
                 {
-                    _logger.LogWarning("Failed to get package versions from source {PackageSource} due to an HTTP error ({StatusCode})", source.Source, exc.StatusCode);
+                    return Enumerable.Empty<NuGetReference>();
                 }
             }
 
-            if (highestVersion is null)
+            var groupResult = results
+                .OrderByDescending(r => r.Identity.Version)
```
mjrousos, line 140: Why `OrderByDescending` here?

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/PackageLoader.cs
```diff
@@ -96,57 +109,64 @@ public async Task<IEnumerable<NuGetReference>> GetNewerVersionsAsync(NuGetRefere
...
+                            {
+                                unsupported = unsupported.Remove(t);
+                            }
+                        }
+                    }
+
+                    return unsupported.Count;
+                })
+                .OrderBy(g => g.Key)
+                .FirstOrDefault();
```
mjrousos, line 159: If I'm understanding this right, it appears to be grouping the versions according to how many of the TFMs they don't support and then choosing the one with the least unsupported TFMs. Would it be better to not group and, instead, filter to only versions that have 0 unsupported TFMs?

mjrousos (changes requested):
Thanks, @twsouthwick! I left some comments inline, mostly questions about PackageLoader.

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Program.cs
```diff
@@ -181,29 +182,18 @@ private static void ShowHeader()
...
-
-            private void WriteString(string input)
-            {
-                foreach (var line in SplitText(input, MaxWidth))
-                {
-                    Console.Out.WriteLine(line);
-                }
-
-                Console.Out.WriteLine();
+                const string Title = "Makes a best-effort attempt to upgrade .NET Framework projects to .NET 5.\n\n" +
```
KSchlobohm, line 193: nit: use {Environment.Newline}

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/UpgradeTarget.cs
```diff
@@ -6,6 +6,7 @@ namespace Microsoft.DotNet.UpgradeAssistant
     public enum UpgradeTarget
     {
         LTS,
-        Current
+        Current,
```
KSchlobohm, line 9: Was surprised to see we already had this class. How does the UpgradeTarget get reflected in our integration testing?

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/PackageLoader.cs
KSchlobohm: I was wondering something similar. if there's 1 unsupported TFM is it a valid package upgrade?

KSchlobohm: are there scenarios that were built for that would make good unit tests?

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Analyzers/TargetCompatibilityReferenceAnalyzer.cs
twsouthwick: Maybe, but I'll keep that for now and we can check on another PR

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/PackageLoader.cs
twsouthwick: I had wondered about that and was developing it for the solution where we could do a "best fit". For now, though, I'd rather just replace it with ones that will actually work and we can deal with it later if it doesn't.

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/PackageLoader.cs
twsouthwick: Removed this as we order it later

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/UpgradeTarget.cs
twsouthwick: It doesn't - we just use the default. The one place where it actually matters has unit tests that cover it.

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Program.cs
twsouthwick: I'm using a const string, so that doesn't work. This should be moved out to a resource file anyway, so I'll leave it as-is and create a task to use resources.

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/PackageLoader.cs
twsouthwick: Fixed

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/PackageLoader.cs
```diff
@@ -68,68 +79,35 @@ public async Task<bool> DoesPackageSupportTargetFrameworksAsync(NuGetReference p
...
+            return SearchByNameAsync(reference.Name, tfms, currentVersion: reference.GetNuGetVersion(), latestMinorAndBuildOnly: latestMinorAndBuildOnly, token: token);
+        }
 
-            _logger.LogDebug("Found versions for package {PackageName}: {PackageVersions}", reference.Name, versionsToReturn);
+        public async Task<NuGetReference?> GetLatestVersionAsync(string packageName, IEnumerable<TargetFrameworkMoniker> tfms, bool includePreRelease, CancellationToken token)
+        {
+            var result = await SearchByNameAsync(packageName, tfms, includePreRelease, token: token).ConfigureAwait(false);
 
-            return versionsToReturn.OrderBy(v => v).Select(v => reference with { Version = v.ToNormalizedString() });
+            return result.FirstOrDefault();
```
mjrousos, line 96: Should this be `LastOrDefault`? It looks like `SearchByNameAsync` orders results least to greatest.

mjrousos (approved):
Looks good. Once comment inline about a `First` that I think should be a `Last` but looks good otherwise.

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/PackageLoader.cs
```diff
@@ -141,31 +119,59 @@ public async Task<IEnumerable<NuGetReference>> GetNewerVersionsAsync(NuGetRefere
...
+                    }
+
+                    return unsupported.IsEmpty;
+                })
+                .Select(r => r.Identity.Version)
+                .OrderBy(v => v)
+                .Select(v => new NuGetReference(name, v.ToNormalizedString()));
         }
 
         private async Task<T> CallWithRetryAsync<T>(Func<Task<T>> func)
```
KSchlobohm, line 177: I think we should start thinking about bringing in polly. I don't think it holds up this PR

twsouthwick: I agree. I started breaking things apart and almost added polly, but I was able to add testing to a static method, so I didn't continue down that path.

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Program.cs
```diff
@@ -181,29 +181,18 @@ private static void ShowHeader()
...
+                                   "This tool does not completely automate the upgrade process and it is expected that projects will have build errors after the tool runs. Manual changes will be required to complete the upgrade to .NET 5.\n\n" +
```
sunandabalu, line 193: Now with this change user can choose to not move to .NET 5, so we should reword here to call that out.

KSchlobohm (approved):

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Program.cs
twsouthwick: Hmmm what would be a good name? .NET Core/.NET 5?

Naming is hard.....

src/components/Microsoft.DotNet.UpgradeAssistant/TargetFramework/TargetFrameworkSelector.cs
```diff
@@ -47,7 +41,14 @@ public async ValueTask<TargetFrameworkMoniker> SelectTargetFrameworkAsync(IProje
                 throw new ArgumentNullException(nameof(project));
             }
 
-            var appBase = _upgradeTarget == UpgradeTarget.Current ? _currentTFMBase : _ltsTFMBase;
+            var appBase = _upgradeTarget switch
+            {
+                UpgradeTarget.Current => _selectorOptions.Current,
+                UpgradeTarget.Preview => _selectorOptions.Preview,
+                UpgradeTarget.LTS => _selectorOptions.LTS,
+                _ => _selectorOptions.LTS,
```
sunandabalu, line 49: Does this mean this defaults to LTS?

tests/components/Microsoft.DotNet.UpgradeAssistant.Tests/TargetFramework/TargetFrameworkSelectorTests.cs
```diff
@@ -16,10 +16,10 @@ namespace Microsoft.DotNet.UpgradeAssistant.Tests
 {
     public class TargetFrameworkSelectorTests
     {
-        private readonly TFMSelectorOptions _options = new TFMSelectorOptions
+        private readonly DefaultTfmOptions _options = new DefaultTfmOptions
         {
-            CurrentTFMBase = Current,
-            LTSTFMBase = LTS,
+            Current = Current,
+            LTS = LTS,
```
sunandabalu, line 22: No need for Preview here?

sunandabalu (approved):
Overall looks good, some minor comments.

src/components/Microsoft.DotNet.UpgradeAssistant/TargetFramework/TargetFrameworkSelector.cs
twsouthwick: No - this is if there's a different enum value (probably should never even happen, but just in case)

tests/components/Microsoft.DotNet.UpgradeAssistant.Tests/TargetFramework/TargetFrameworkSelectorTests.cs
twsouthwick: Good catch. Added

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Program.cs
sunandabalu: You missed LN 193 :)

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Program.cs
```diff
@@ -181,29 +181,18 @@ private static void ShowHeader()
...
-
-            private void WriteString(string input)
-            {
-                foreach (var line in SplitText(input, MaxWidth))
-                {
-                    Console.Out.WriteLine(line);
-                }
-
-                Console.Out.WriteLine();
+                const string Title = "Makes a best-effort attempt to upgrade .NET Framework projects to the selected LTS, Current, or Preview target framework (defaults to Current).\n\n" +
```
sunandabalu, line 192: Probably makes sense to use @mjrousos 's version of wording - ```Makes a best-effort attempt to upgrade .NET Framework projects to current, preview or LTS versions of .NET```

src/components/Microsoft.DotNet.UpgradeAssistant/TargetFramework/TargetFrameworkSelector.cs
sunandabalu: I can't seem to find where you are defaulting to Current, could you point me to that?

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Program.cs
KSchlobohm: What-if it's just a format string and we use a placeholder to format in the default target name

mjrousos: That's not a bad idea. But we still have both LTS and Current so we'd have to pick which one of them to use.

Issue #410, opened 2021-04-02
VB WinForm: Compile errors due to missing Import System.Windows.Forms


KSchlobohm:
![image](https://user-images.githubusercontent.com/23503973/113459603-a3ca5800-93db-11eb-9fd2-6fe4778abf2f.png)

VB Form compile errors due to missing import.

Adding `Imports System.Windows.Forms` to *Form1.Designer.vb* resolves this issue

PR #474, opened 2021-04-30
Update try-convert

KSchlobohm:
**PR Title**
Update try-convert to version 0.7.222801

**PR Description**
- Updates the way we invoke try-convert. We now send the MSBuildPath as an argument
- Version 0.7.222801 resolves a bug that could surface [when multiple versions of Visual Studio are installed](https://github.com/dotnet/try-convert/issues/395)

Addresses #410

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@ All notable changes to the .NET Upgrade Assistant will be documented in this fil
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
+## Current
+
+### Fixed
+- VB Win Forms projects should keep import for 'System.Windows.Forms' [#474](https://github.com/dotnet/upgrade-assistant/pull/474)
+
 ## Version 0.2.222702 - 2021-04-27 ([Link](https://www.nuget.org/packages/upgrade-assistant/0.2.222702))
 
 ### Added
--- a/src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildRegistrationStartup.cs
+++ b/src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildRegistrationStartup.cs
@@ -25,15 +25,12 @@ namespace Microsoft.DotNet.UpgradeAssistant.MSBuild
         private static VisualStudioInstance? _instance;
 
         private readonly ILogger _logger;
+        private readonly MSBuildPathLocator _locator;
 
-        public MSBuildRegistrationStartup(ILogger<MSBuildRegistrationStartup> logger)
+        public MSBuildRegistrationStartup(ILogger<MSBuildRegistrationStartup> logger, MSBuildPathLocator locator)
         {
-            _logger = logger;
-        }
-
-        public static string GetMSBuildPath()
-        {
-            return _instance.MSBuildPath;
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            _locator = locator ?? throw new ArgumentNullException(nameof(locator));
         }
 
         public Task<bool> StartupAsync(CancellationToken token)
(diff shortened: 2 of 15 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,11 @@ All notable changes to the .NET Upgrade Assistant will be documented in this fil
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
+## Current
+
+### Fixed
+- VB Win Forms projects should keep import for 'System.Windows.Forms' [#474](https://github.com/dotnet/upgrade-assistant/pull/474)
```
KSchlobohm, line 9: are we linking pull requests?

mjrousos: Hm... seems we're not consistent. @cathysull, @sunandabalu, @twsouthwick, should we be linking issues or PRs in the change log?

I vote for PR since that's where the change actually happens, but I don't feel strongly about it.

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildRegistrationStartup.cs
```diff
@@ -31,6 +31,11 @@ public MSBuildRegistrationStartup(ILogger<MSBuildRegistrationStartup> logger)
             _logger = logger;
         }
 
+        public static string GetMSBuildPath()
```
mjrousos, line 34: I don't care for the public static method on an instance class. Can we tweak this to either:

1. Be an instance method and register MSBuildRegistrationStartup as a singleton
2. Leave this service as transient but give it a dependency on MSBuildPathLocator and have it set the MSBuildPath property when it runs so that future services can get the property from (the singleton instance of) MSBuildPathLocator

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.ProjectFormat/TryConvertTool.cs
```diff
@@ -42,9 +42,21 @@ public class TryConvertTool : ITryConvertTool
...
             Version = GetVersion();
         }
 
+        /// <summary>
+        /// Path to MSBuild is often located in "Program Files" folder. We wrap the path with quotes
+        /// to handle the space as continuous argument and must also handle the backslash escape characters.
+        /// </summary>
+        private static string HandleEscapeCharacters(string folderPath)
+        {
+            return folderPath.Replace("\\", "\\\\");
```
mjrousos, line 55: It's not clear to me why these need escaped. We don't need C:\\foo\\etc when we pass paths to try-convert from the command line, do we?

mjrousos (commented):
A few comments inline, but looks good otherwise.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.ProjectFormat/TryConvertTool.cs
KSchlobohm: The path to MSBUILD is `C:\Program Files\dotnet\sdk\5.0.202\` so we needed to add quotes

The escape chars prevent a sql injection style formatting problem with the TryConvertArgumentsFormat string

"--no-backup -m \"{0}\" --force-web-conversion --keep-current-tfms -p \"{1}\""

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildRegistrationStartup.cs
KSchlobohm: went with option 2 I liked the idea that there's only one place to get this value.

CHANGELOG.md
twsouthwick: I'm good with either. If prefer the pr since that's the actual change, but sometimes an issue may be better as multiple prs could cover a change

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.ProjectFormat/TryConvertTool.cs
```diff
@@ -54,6 +57,11 @@ public class TryConvertTool : ITryConvertTool
         public string GetCommandLine(IProject project)
             => Invariant($"{Path} {GetArguments(project.Required())}");
 
+        private string? GetMSBuildPath()
+        {
+            return _locator.MSBuildPath?.Replace("\\", "\\\\");
```
mjrousos, line 62: I feel like I'm still missing something here. Can you give an example of a scenario that won't work without the `.Replace` to help me get my head around it?

mjrousos (approved):
Looks good.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.ProjectFormat/TryConvertTool.cs
KSchlobohm: I think it's because we're passing args to the process as a single string rather than an array of strings. but when i run without the escapes I see that try-convert receives all arguments after the msBuildPath as part of the path

image

PR #480, opened 2021-05-03
Add design for chem integration

twsouthwick:
This change adds a design spec for the work that will be included to enable chem integration into Upgrade Assistant.

### DIFF
--- a/docs/design/Chem Integration.md
+++ b/docs/design/Chem Integration.md
@@ -27,7 +27,7 @@ upgrade-assistant some-project.vbproj --extension .\path\to\UpgradeAssistantChem
 ### Decoupling the data
 Initially, any data needed will be provided with the extension. This data would be what is used to map the fingerprints to actual packages. They often range around the 100mb-200mb range for useful data, but could be potentially larger if desired.
 
-However, it may be nice to decouple the data from the actual extension itself so they can be updated independently. In order to support his, Upgrade Assistant should provide a way for steps/extensions to provide step-dependent data. This could be something that a step could register it as having, and then Upgrade Assistant could provide commands such as the following:
+However, it may be nice to decouple the data from the actual extension itself so they can be updated independently. In order to support this, Upgrade Assistant should provide a way for steps/extensions to provide step-dependent data. This could be something that a step could register it as having, and then Upgrade Assistant could provide commands such as the following:
 
 - `upgrade-assistnat manage-data list`: List registered data sources
 - `upgrade-assistant manage-data update --name [name]`: Update a registered data source (managed by the step itself)
--- a/docs/design/Chem Integration.md
+++ b/docs/design/Chem Integration.md
@@ -1,6 +1,6 @@
 # Loose binary identification with Upgrade Assistant
 
-It is not uncommon in large code bases to have a collection of dependencies that have been persisted as only the `.dll` files with no known provenance. This makes it challenging to identify which binary is being used, especially when a different target (ie .NET Standard) is needed. In order to solve this, this document proposes integration with a data source that can help match these loose assemblies to known NuGet packages.
+It is not uncommon in large code bases to have a collection of dependencies that have been persisted as only the `.dll` files with no known provenance. These files are considered "loose binaries". This makes it challenging to identify which binary is being used, especially when a different target (ie .NET Standard) is needed. In order to solve this, this document proposes integration with a data source that can help match these loose assemblies to known NuGet packages.
 
 Chem is a tool developed internally to gain insights from the ecosystem of code attached to .NET. As a result of the data that it gathers, it is able to match loose assemblies with their probable origins.  As customers are looking to move to .NET 5, this may pose a challenge as these versions are highly likely to be .NET Framework compilations that may not work as expected on newer platforms. Often, these libraries have NuGet packages now and newer versions that could work on newer platforms.
 
--- a/docs/design/Chem Integration.md
+++ b/docs/design/Chem Integration.md
@@ -1,6 +1,6 @@
 # Loose binary identification with Upgrade Assistant
 
-It is not uncommon in large code bases to have a collection of dependencies that have been persisted as only the `.dll` files with no known provenance. These files are considered "loose binaries". This makes it challenging to identify which binary is being used, especially when a different target (ie .NET Standard) is needed. In order to solve this, this document proposes integration with a data source that can help match these loose assemblies to known NuGet packages.
+It is not uncommon in large code bases to have a collection of dependencies that have been persisted as only the `.dll` files with no known provenance. These files are considered "loose binaries". This makes it challenging to identify what a binary is while upgrading where an updated version is often needed to support .NET Standard/Core/5. In order to solve this, this document proposes integration with a data source that can help match these loose assemblies to known NuGet packages.
 
 Chem is a tool developed internally to gain insights from the ecosystem of code attached to .NET. As a result of the data that it gathers, it is able to match loose assemblies with their probable origins.  As customers are looking to move to .NET 5, this may pose a challenge as these versions are highly likely to be .NET Framework compilations that may not work as expected on newer platforms. Often, these libraries have NuGet packages now and newer versions that could work on newer platforms.
 
@@ -11,7 +11,7 @@ The algorithm used to match the binaries makes use of fingerprints that may be u
 - File contents (and hashes thereof)
 - and others
 
-Work has already been done to generate a database of known packages along these and other identity dimensions. This data can then be used to generate potential indexes that could be used from within Upgrade Assistant to identify along this potential dimensions, both for notification and even to update a project file to reference the NuGet package rather than the loose assembly.
+Work has already been done to generate a database of known packages along these and other identity dimensions. This data can then be used to generate potential indexes that could be used from within Upgrade Assistant to identify along these potential dimensions, both for notification and even to update a project file to reference the NuGet package rather than the loose assembly.
 
 ## User Experience
 
--- a/docs/design/Chem Integration.md
+++ b/docs/design/Chem Integration.md
@@ -2,8 +2,6 @@
 
 It is not uncommon in large code bases to have a collection of dependencies that have been persisted as only the `.dll` files with no known provenance. These files are considered "loose binaries". This makes it challenging to identify what a binary is while upgrading where an updated version is often needed to support .NET Standard/Core/5. In order to solve this, this document proposes integration with a data source that can help match these loose assemblies to known NuGet packages.
 
-Chem is a tool developed internally to gain insights from the ecosystem of code attached to .NET. As a result of the data that it gathers, it is able to match loose assemblies with their probable origins.  As customers are looking to move to .NET 5, this may pose a challenge as these versions are highly likely to be .NET Framework compilations that may not work as expected on newer platforms. Often, these libraries have NuGet packages now and newer versions that could work on newer platforms.
-
 The algorithm used to match the binaries makes use of fingerprints that may be useful in identifying what it may be:
 
 - Assembly name
@@ -20,10 +18,10 @@ The main scenario this will focus on is one where a project has downloaded an as
 - Identify what package the dependency came from
 - Update a project file to reference the NuGet package rather than a binary reference
 
-The expected UX for this will be to implement it as an extension that a user can opt into. As a user, in order to run the Chem integration, they will obtain the extension (ie `UpgradeAssitantChem.zip`) and then run as they normally would:
+The expected UX for this will be to implement it as an extension that a user can opt into. As a user, in order to run the extension, they will obtain the extension (ie `UpgradeAssitant-Loose-Assembly-Identification.zip`) and then run as they normally would:
 
 ```
-upgrade-assistant some-project.vbproj --extension .\path\to\UpgradeAssistantChem.zip
+upgrade-assistant some-project.vbproj --extension .\path\to\UpgradeAssitant-Loose-Assembly-Identification.zip
 ```
 
 ### Decoupling the data
@@ -55,9 +53,9 @@ Below are the main work items that will be explored broken into two sprints. The
 - As a user, I know which loose assemblies do not have any available matches (#485)
 
 ### Sprint 2
-- As a user, I can update my chem data index independently of the extension (#486)
+- As a user, I can update my loose assembly data index independently of the extension (#486)
 - As a user, I can manage data steps need at a global level (#487)
-- As a user, I can receive recommendations from chem using multiple indexes (#488)
+- As a user, I can receive recommendations from multiple indexes for loose assembly identification (#488)
 
 ## Out of scope/Future Work
 
@@ -66,4 +64,4 @@ Some work that are currently not in scope:
 - Custom feeds (such as AzDO, Artifactory, file system, etc). NuGet.org will be the focus of the initial spike
 - Providing tooling to generate custom indexes
 - Identifying any security concerns with switching to NuGet packages. We may want to explore identifying if a package has a security concern, but it's still probably the same binary.
-- Hosting the chem lookup in any sort of service. This design is to integrate Chem in as simply as possible without external dependencies.
+- Hosting the lookup in any sort of service. This design is to integrate the analysis in as simply as possible without external dependencies.
--- a/docs/design/Chem Integration.md
+++ b/docs/design/Chem Integration.md
@@ -41,6 +41,9 @@ There are a number of challenges with this, as it is completely based on heurist
 - If an assembly was never published in a package on NuGet.org, it will never show up in the available indexes
 - Private builds of projects may not match external builds
 - Currently only planning on supporting packages from NuGet.org so loose assemblies from a company's internal teams will not be found. May be included in future work.
+- Some loose assemblies may be "ride-along" in packages that they do not actually belong to (ie Newtonsoft.Json is included in a number of packages)
+- Early NuGet days had evolving best practice
+- NetStandard and other modern patterns are represented as bifurcations in package publish (ex. Foo and Foo.NetStandard packages)
 
 ## Planned Work
 
--- a/docs/design/Chem Integration.md
+++ b/docs/design/Chem Integration.md
@@ -1,0 +1,66 @@
+# Chem integration into Upgrade Assistant
+
+Chem is a tool developed internally to solve the problem of matching loose assemblies (ie `.dll` files added into projects with unknown provenance). As customers are looking to move to .NET 5, this may pose a challenge as these versions are highly likely to be .NET Framework compilations that may not work as expected on newer platforms. Often, these libraries have NuGet packages now and newer versions that could work on newer platforms.
+
+The main idea behind Chem is based on the idea of chemical fingerprints (hence the naming) that some small amount of data may be used to infer the original identity. In the case of binaries, there are a number of fingerprints that may be useful in identifying what it may be:
+
+- File name
+- Public key token
+- File contents (and hashes thereof)
+- and others
+
+Work has already been done to generate a database of known packages along these and other identity dimensions. This data can then be used to generate potential indexes that could be used from within Upgrade Assistant to identify along this potential dimensions, both for notification and even to update a project file to reference the NuGet package rather than the loose assembly.
+
+## User Experience
+
+The main scenario this will focus on is one where a project has downloaded an assembly and added it to its repo and referenced it from various projects. The end result of this work should be able to:
+
+- Identify what package the dependency came from
+- Update a project file to now reference this from
+
+The expected UX for this will be to implement it as an extension that a user can opt into. As a user, in order to run the Chem integration, they will obtain the extension (ie `UpgradeAssitantChem.zip`) and then run as they normally would:
+
+```
+upgrade-assistant some-project.vbproj --extension .\path\to\UpgradeAssistantChem.zip
+```
+
+### Decoupling the data
+Initially, any data needed will be provided with the extension. This data would be what is used to map the fingerprints to actual packages. They often range around the 100mb-200mb range for useful data, but could be potentially larger if desired.
+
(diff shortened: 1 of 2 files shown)


docs/design/Chem Integration.md
```diff
@@ -0,0 +1,66 @@
...
+- and others
+
+Work has already been done to generate a database of known packages along these and other identity dimensions. This data can then be used to generate potential indexes that could be used from within Upgrade Assistant to identify along this potential dimensions, both for notification and even to update a project file to reference the NuGet package rather than the loose assembly.
+
+## User Experience
+
+The main scenario this will focus on is one where a project has downloaded an assembly and added it to its repo and referenced it from various projects. The end result of this work should be able to:
+
+- Identify what package the dependency came from
+- Update a project file to now reference this from
```
mjrousos, line 19: Nit-pick, something seems off about this sentence. Maybe a missing word or something? I think it's meant to say "Update a project file to now reference this package" or something along those lines.

docs/design/Chem Integration.md
```diff
@@ -0,0 +1,66 @@
...
+
+The expected UX for this will be to implement it as an extension that a user can opt into. As a user, in order to run the Chem integration, they will obtain the extension (ie `UpgradeAssitantChem.zip`) and then run as they normally would:
```
mjrousos, line 21: This might be a good time to mention that the feature of loading an extension from an archive isn't finished yet :) Extensions load from folders fine but the zip path will give an error currently.

It shouldn't be too much work to get that implemented, though. I can work on it in parallel with the chem work so hopefully it's in place by the time this is ready!

docs/design/Chem Integration.md
```diff
@@ -0,0 +1,66 @@
...
+
+```
+upgrade-assistant some-project.vbproj --extension .\path\to\UpgradeAssistantChem.zip
+```
+
+### Decoupling the data
+Initially, any data needed will be provided with the extension. This data would be what is used to map the fingerprints to actual packages. They often range around the 100mb-200mb range for useful data, but could be potentially larger if desired.
+
+However, it may be nice to decouple the data from the actual extension itself so they can be updated independently. In order to support his, Upgrade Assistant should provide a way for steps/extensions to provide step-dependent data. This could be something that a step could register it as having, and then Upgrade Assistant could provide commands such as the following:
```
mjrousos, line 30:
```suggestion
However, it may be nice to decouple the data from the actual extension itself so they can be updated independently. In order to support this, Upgrade Assistant should provide a way for steps/extensions to provide step-dependent data. This could be something that a step could register it as having, and then Upgrade Assistant could provide commands such as the following:
```

docs/design/Chem Integration.md
```diff
@@ -0,0 +1,66 @@
...
+
+- `upgrade-assistnat manage-data list`: List registered data sources
+- `upgrade-assistant manage-data update --name [name]`: Update a registered data source (managed by the step itself)
+
+There are potentially many questions here as to handle the data. This is provided solely as a way to demonstrate a potential way to decouple the data required. Initially, the data will be included with the extension, but potentially have optional indexes (or indexes that carry more data and are much larger) will require some sort of way to manage the data.
```
mjrousos, line 35:
```suggestion
There are potentially many questions here as to how to handle the data. This is provided solely as a way to demonstrate a potential way to decouple the data required. Initially, the data will be included with the extension, but potentially have optional indexes (or indexes that carry more data and are much larger) will require some sort of way to manage the data.
```

mjrousos, line 35: I like the idea that the extension has the initial data set but that it can be updated in this way. I think `manage-data` is a nice option to have but I expect it will be less discoverable than just using the extension directly so we'll want the extension to work even if no data is updated.

docs/design/Chem Integration.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Challenges
+
+There are a number of challenges with this, as it is completely based on heuristics in order to match package identities. Potential limitations:
+
+- If a package was never released to NuGet, it will never show up in the available indexes
+- Private builds of projects may not match external builds
+- Currently only planning on supporting packages from NuGet.org so loose assemblies from a company's internal teams will not be found.
```
mjrousos, line 43: As a super long-term pie-in-the-sky idea, could we make Chem's indexing technology available in some form for customers to create their own data sets from internal packages?

docs/design/Chem Integration.md
```diff
@@ -0,0 +1,66 @@
...
+- As a user, I can update my chem data index independently of the extension (#486)
+- As a user, I can have access to the same data from multiple projects (#487)
+- As a user, I can receive recommendations from chem using multiple indexes (#488)
+
+## Out of scope/Future Work
+
+Some work that are currently not in scope:
+
+- Custom feeds (such as AzDO, Artifactory, file system, etc). NuGet.org will be the focus of the initial spike
+- Identifying any security concerns with switching to NuGet packages. We may want to explore identifying if a package has a security concern, but it's still probably the same binary.
```
mjrousos, line 65: This feels low priority as it would be uncommon (though not impossible) for an old loose binary to be secure but the latest packaged version of a binary to be insecure.

mjrousos (approved):
Looks good!

docs/design/Chem Integration.md
```diff
@@ -0,0 +1,66 @@
...
+upgrade-assistant some-project.vbproj --extension .\path\to\UpgradeAssistantChem.zip
+```
+
+### Decoupling the data
+Initially, any data needed will be provided with the extension. This data would be what is used to map the fingerprints to actual packages. They often range around the 100mb-200mb range for useful data, but could be potentially larger if desired.
+
+However, it may be nice to decouple the data from the actual extension itself so they can be updated independently. In order to support his, Upgrade Assistant should provide a way for steps/extensions to provide step-dependent data. This could be something that a step could register it as having, and then Upgrade Assistant could provide commands such as the following:
+
+- `upgrade-assistnat manage-data list`: List registered data sources
+- `upgrade-assistant manage-data update --name [name]`: Update a registered data source (managed by the step itself)
```
sunandabalu, line 33: Should this include an ```add``` option too?

docs/design/Chem Integration.md
sunandabalu: There are situations where the content of the library end up in a random nuget package that then was de-listed due to security concerns. But we will def need more data to know the % of such occurrences to make this feature a priority.

sunandabalu (approved):
You have captured the essence of everything that we have discussed with Chem integration, LGTM.

docs/design/Chem Integration.md
```diff
@@ -0,0 +1,66 @@
+# Chem integration into Upgrade Assistant
+
+Chem is a tool developed internally to solve the problem of matching loose assemblies (ie `.dll` files added into projects with unknown provenance). As customers are looking to move to .NET 5, this may pose a challenge as these versions are highly likely to be .NET Framework compilations that may not work as expected on newer platforms. Often, these libraries have NuGet packages now and newer versions that could work on newer platforms.
```
marklio, line 3: This is not an accurate statement. It happens to be good at this due to the data that it contains, but wasn't created for this purpose (and isn't even what it ends up being used for most). I think a public-facing statement of what Chem is for would be something like "Chem is a tool developed internally to gain insights from the ecosystem of code attached to .NET. As a result of the data that it gathers, it is able to match loose assemblies with their probable origins" (or something like that).

docs/design/Chem Integration.md
sunandabalu: Yes this would def be a useful feature for customers to set up their own index, @marklio can comment on if this something on Chem's radar.

docs/design/Chem Integration.md
```diff
@@ -0,0 +1,66 @@
...
+
+The main idea behind Chem is based on the idea of chemical fingerprints (hence the naming) that some small amount of data may be used to infer the original identity. In the case of binaries, there are a number of fingerprints that may be useful in identifying what it may be:
```
marklio, line 5: Again, this is a little wonky. ;) I'd avoid trying to write a history of Chem here and concentrate on the idea that Chem has the data needed to correlate these fingerprints with packages.

docs/design/Chem Integration.md
```diff
@@ -0,0 +1,66 @@
...
+
+- File name
```
marklio, line 7: I'd say assembly name, specifically, rather than file name.

docs/design/Chem Integration.md
```diff
@@ -0,0 +1,66 @@
...
+```
+upgrade-assistant some-project.vbproj --extension .\path\to\UpgradeAssistantChem.zip
+```
+
+### Decoupling the data
+Initially, any data needed will be provided with the extension. This data would be what is used to map the fingerprints to actual packages. They often range around the 100mb-200mb range for useful data, but could be potentially larger if desired.
+
+However, it may be nice to decouple the data from the actual extension itself so they can be updated independently. In order to support his, Upgrade Assistant should provide a way for steps/extensions to provide step-dependent data. This could be something that a step could register it as having, and then Upgrade Assistant could provide commands such as the following:
+
+- `upgrade-assistnat manage-data list`: List registered data sources
```
marklio, line 32:
```suggestion
- `upgrade-assistant manage-data list`: List registered data sources
```

docs/design/Chem Integration.md
```diff
@@ -0,0 +1,66 @@
...
+- `upgrade-assistant manage-data update --name [name]`: Update a registered data source (managed by the step itself)
+
+There are potentially many questions here as to handle the data. This is provided solely as a way to demonstrate a potential way to decouple the data required. Initially, the data will be included with the extension, but potentially have optional indexes (or indexes that carry more data and are much larger) will require some sort of way to manage the data.
+
+## Challenges
+
+There are a number of challenges with this, as it is completely based on heuristics in order to match package identities. Potential limitations:
+
+- If a package was never released to NuGet, it will never show up in the available indexes
```
marklio, line 41:
```suggestion
- If an assembly was never published in a package on NuGet.org, it will never show up in the available indexes
```

docs/design/Chem Integration.md
marklio: Yes. We've discussed this, and we can definitely do something here.

docs/design/Chem Integration.md
```diff
@@ -0,0 +1,66 @@
...
+- Private builds of projects may not match external builds
+- Currently only planning on supporting packages from NuGet.org so loose assemblies from a company's internal teams will not be found.
+
+## Planned Work
+
+Below are the main work items that will be explored broken into two sprints. There is currently a pre-analysis migration effort being worked on, and this work may be updated to ensure that the information flows as expected into the reports generated there as those features become available.
+
+### Sprint 1
+- As a user, I can consume an extension that integrates chem (#482)
```
marklio, line 50: I'm reluctant to word it this way. Chem is MUCH bigger than this. I'd say we should talk about specific capabilities rather than try to use Chem as a synonym for package lookup.

docs/design/Chem Integration.md
```diff
@@ -0,0 +1,66 @@
...
+- As a user, the chem integration can identify what loose assemblies I have (#483)
+- As a user, the chem integration will replace my loose assemblies with ones it identifies (#484)
+- As a user, I know which loose assemblies do not have any available matches (#485)
+
+### Sprint 2
+- As a user, I can update my chem data index independently of the extension (#486)
+- As a user, I can have access to the same data from multiple projects (#487)
```
marklio, line 57: What does this mean?

marklio (changes requested):
Overall, this looks good. As I've said in individual comments, I'd like to avoid trying to use "Chem" as a synonym for "package lookup". It's fine to talk about Chem as the source of the data, but we're not "integrating Chem".

docs/design/Chem Integration.md
marklio: Just to be clear, this isn't on "Chem's radar", but it would definitely be possible to generate similar indexes from "local" sources.

docs/design/Chem Integration.md
twsouthwick: I imagine I'll find a few of these corner cases, but that's good to know

docs/design/Chem Integration.md
marklio: I think a principle we're dancing around here is that we don't want to ever recommend moving to a package that has a known security vulnerability. The question is, what role should these indexes play in that? And I think this statement here is just saying that this is out of scope at this time from the perspective of the integrating the package indexes.

mjrousos: That makes sense. The data collected by Chem doesn't have information on whether packages have security vulnerabilities I take it? In general, having a database like that somewhere for UA to use would be useful. I agree that it doesn't seem in scope for this work, though.

docs/design/Chem Integration.md
twsouthwick: I've added an issue https://github.com/dotnet/upgrade-assistant/issues/490 to cover this and marked it as blocking the epic.

docs/design/Chem Integration.md
twsouthwick: Yeah, depends on how large it is. Any useful data seems to be on the order of a few indexes of 100mb+

docs/design/Chem Integration.md
twsouthwick: Good call. Once we have a way to generate indexes, we can explore additional ways to generate it from local sources if needed

twsouthwick: I had something similarly called out as out of scope for now, but will clarify the wording

docs/design/Chem Integration.md
twsouthwick: Yup I agree with all that is said here. It makes sense potentially as as a separate step for Upgrade Assistant to analyze for any known CVE's, but didn't want to put it in scope here.

docs/design/Chem Integration.md
twsouthwick: The example here is to update data sources that the steps register and thus know about. I didn't want to go too deep here besides to show a possibility for managing data, but that a separate design would probably be needed to fully flesh that out.

docs/design/Chem Integration.md
twsouthwick: I've reworded it, but it's about being able to manage data at a global level in upgrade assistant (ie simplify storage of indexes so they can easily be consumed/updated/etc)

twsouthwick:
Thanks for the feedback - I've updated the doc and would appreciate another once over

docs/design/Chem Integration.md
```diff
@@ -0,0 +1,69 @@
...
+There are potentially many questions here as to how to handle the data. This is provided solely as a way to demonstrate a potential way to decouple the data required. Initially, the data will be included with the extension, but potentially have optional indexes (or indexes that carry more data and are much larger) will require some sort of way to manage the data.
+
+## Challenges
+
+There are a number of challenges with this, as it is completely based on heuristics in order to match package identities. Potential limitations:
+
+- If an assembly was never published in a package on NuGet.org, it will never show up in the available indexes
+- Private builds of projects may not match external builds
+- Currently only planning on supporting packages from NuGet.org so loose assemblies from a company's internal teams will not be found. May be included in future work.
+
```
marklio, line 46: I don't know if you're trying to be exhaustive here, but there are lots of other relevant challenges:
* "Loose assemblies" in apps are more likely to be "ride-along" assemblies in packages, making ownership hard to determine
* Early NuGet days had evolving best practices.
* NetSandard and other modern patterns are represented as bifurcations in package publishing (ex. Foo and Foo.NetStandard packages)

marklio (approved):
Looks good

docs/design/Chem Integration.md
```diff
@@ -0,0 +1,69 @@
+# Loose binary identification with Upgrade Assistant
+
+It is not uncommon in large code bases to have a collection of dependencies that have been persisted as only the `.dll` files with no known provenance. These files are considered "loose binaries". This makes it challenging to identify what a binary is while upgrading where an updated version is often needed to support .NET Standard/Core/5. In order to solve this, this document proposes integration with a data source that can help match these loose assemblies to known NuGet packages.
```
KSchlobohm, line 3: There's an unstated assumption here that the reader is familiar with our upgrade guidance and understands that they should update their dependencies as part of the upgrade process. Where possible, we actually prefer this is done before they run upgrade assistant to smooth out the process.

Maybe something simple in the intro like:
`One of the earliest steps in the process of upgrading to .NET Standard/Core/5 is to upgrade your dependencies.`

KSchlobohm, line 3: wasn't clear how long this document will live but I think it's better to introduce a term like *loose binaries* with some prominence that we keep using whenever we reference the word as one of the nouns/topics distinct to our project's ubiquitous language. If this doc lives on I recommend replacing quotes with italic emphasis for a subtle but *distinct* highlight.

It might be helpful to start a glossary at some point.

docs/design/Chem Integration.md
```diff
@@ -0,0 +1,69 @@
...
+
+Chem is a tool developed internally to gain insights from the ecosystem of code attached to .NET. As a result of the data that it gathers, it is able to match loose assemblies with their probable origins.  As customers are looking to move to .NET 5, this may pose a challenge as these versions are highly likely to be .NET Framework compilations that may not work as expected on newer platforms. Often, these libraries have NuGet packages now and newer versions that could work on newer platforms.
```
KSchlobohm, line 5: Do we care about the phrase "chem"? I think the doc could be better described as "Loose Assembly Resolution Process"

KSchlobohm, line 5: "As a result of the data that it gathers" - what data, does it collect favicons?

how about "By analyzing each assembly and comparing it to those we've previously analyzed we are able to infer an assembly's probably origin"

docs/design/Chem Integration.md
```diff
@@ -0,0 +1,69 @@
...
+The expected UX for this will be to implement it as an extension that a user can opt into. As a user, in order to run the Chem integration, they will obtain the extension (ie `UpgradeAssitantChem.zip`) and then run as they normally would:
+
+```
+upgrade-assistant some-project.vbproj --extension .\path\to\UpgradeAssistantChem.zip
+```
+
+### Decoupling the data
+Initially, any data needed will be provided with the extension. This data would be what is used to map the fingerprints to actual packages. They often range around the 100mb-200mb range for useful data, but could be potentially larger if desired.
+
+However, it may be nice to decouple the data from the actual extension itself so they can be updated independently. In order to support this, Upgrade Assistant should provide a way for steps/extensions to provide step-dependent data. This could be something that a step could register it as having, and then Upgrade Assistant could provide commands to view and update. An example of what this might look like is the following:
```
KSchlobohm, line 32: I worry less about step dependent data and more about providing a data source that lists 1 package in 2 locations. As a customer, I may have Unity in my project and Unity exists on the public feed but I might have an enterprise arhcitecture team that wants me to use the "approved" version of Unity to increase compatibility between other projects.

strictly speaking from a compatibility perspective, these customers would centralize control on a few well-known packages but still allow the freedom to choose other projects from NuGet that meet specific needs.

KSchlobohm: speaking of providing competing data sources... can i scan my own NuGet repo? would anyone want to do that?

docs/design/Chem Integration.md
```diff
@@ -0,0 +1,69 @@
...
+
+- `upgrade-assistant manage-data list`: List registered data sources
+- `upgrade-assistant manage-data update --name [name]`: Update a registered data source (managed by the step itself)
+
+There are potentially many questions here as to how to handle the data. This is provided solely as a way to demonstrate a potential way to decouple the data required. Initially, the data will be included with the extension, but potentially have optional indexes (or indexes that carry more data and are much larger) will require some sort of way to manage the data.
+
+## Challenges
+
+There are a number of challenges with this, as it is completely based on heuristics in order to match package identities. Potential limitations:
```
KSchlobohm, line 41: when someone says the approach is heuristic based, i start wondering if i can affect the heuristics to "fit" the mapping process to my project. AutoMapper works great, but sometimes i need to override a default mapping.

docs/design/Chem Integration.md
```diff
@@ -0,0 +1,69 @@
...
+### Sprint 2
+- As a user, I can update my chem data index independently of the extension (#486)
+- As a user, I can manage data steps need at a global level (#487)
+- As a user, I can receive recommendations from chem using multiple indexes (#488)
+
+## Out of scope/Future Work
+
+Some work that are currently not in scope:
+
+- Custom feeds (such as AzDO, Artifactory, file system, etc). NuGet.org will be the focus of the initial spike
```
KSchlobohm, line 66: 😄

docs/design/Chem Integration.md
marklio: I totally agree with this.

marklio: I'd love to not talk about how Chem works, but if we need to, this is reasonable.

sunandabalu: +1 I had this lingering thought about not calling it Chem instead call it by a name that specifies what its trying to do at least from a UA perspective.

twsouthwick: I've removed all references to chem

twsouthwick: Probably not needed, so I just removed that paragraph.

docs/design/Chem Integration.md
twsouthwick: That's a good point - the idea is to identify heuristics on how to identify from multiple. It's still an open question.

Own data sources is in the future work/not in scope

docs/design/Chem Integration.md
twsouthwick: Good question - there are still definitely open questions about that effectiveness. Hence this spike

docs/design/Chem Integration.md
twsouthwick: Add this to the glossary!

Issue #463, opened 2021-04-26
Create a VB Code Analyzer/Fixer


KSchlobohm:

KSchlobohm:
research task for #405

PR #489, opened 2021-05-03
ApiController Code Analyzer and Fixer

KSchlobohm:
**PR Title**
Language Agnostic analyzer and code fixer for ApiController

**PR Description**
Will replace the base class of ApiController.

Addresses #463


src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/ApiControllerAnalyzer.cs
```diff
@@ -0,0 +1,100 @@
...
+        public static Project AddMetadataReferences(Project project)
+        {
+            if (project is null)
+            {
+                return project!;
+            }
+
+            // todo - still an open question about how we locate metadatareferences
+            const string assemblyFolder = @"C:\deleteMe\";
+            var assemblyPath = Path.Combine(assemblyFolder, $"System.Web.Http.dll");
```
KSchlobohm, line 95: This assembly comes from NuGet while others could come from locally installed .NET framework

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/ApiControllerAnalyzer.cs
```diff
@@ -0,0 +1,100 @@
...
+            context.RegisterSyntaxNodeAction(AnalyzeNode, CodeAnalysis.CSharp.SyntaxKind.SimpleBaseType);
+            context.RegisterSyntaxNodeAction(AnalyzeNode, CodeAnalysis.VisualBasic.SyntaxKind.InheritsStatement);
+        }
+
+        private void AnalyzeNode(SyntaxNodeAnalysisContext context)
+        {
+            var namedTypeSymbol = context.ContainingSymbol as INamedTypeSymbol;
+
+            if (namedTypeSymbol != null && namedTypeSymbol.BaseType != null
+                && namedTypeSymbol.BaseType.ToDisplayString().Equals($"{BadNamespace}.{BadClassName}", StringComparison.Ordinal))
```
KSchlobohm, line 50: uses symbol analysis to drive language agnostic detection

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/assets/TestProject.Fixed.csproj
```diff
@@ -0,0 +1,22 @@
...
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNet.Razor" Version="3.2.7" />
+    <PackageReference Include="Microsoft.AspNet.Mvc" Version="3.2.7" />
+    <PackageReference Include="Microsoft.Owin" Version="4.1.1" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Compile Remove="*\UA*.cs" />
+    <Compile Include="*\UA*.Fixed.cs" />
```
KSchlobohm, line 15: Not sure how we get around having classes with the same name in 2 separate files but this was causing problems in the VBproj so I decided to apply it to the CSproj to mitigate any potential future issues.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/ApiControllerAnalyzer.cs
```diff
@@ -0,0 +1,108 @@
...
+
+        /// <summary>
+        /// Uses syntax analysis to highlight the correct part of the node. At this point
+        /// the context.Node looks like 'Inherits System.Web.Http.ApiController' and we
+        /// do not want to highlight the VisualBasic.SyntaxKind.InheritsKeyword
+        /// </summary>
+        /// <param name="context"></param>
+        private static void ReportVisualBasicSyntax(SyntaxNodeAnalysisContext context)
+        {
+            var baseClass = context.Node.DescendantNodes()
```
twsouthwick, line 79: Language agnostic means that you can't reference any of the c# or vb roslyn assemblies. That can cause problems in cases where the compiler does not load the vb stuff or the cs stuff.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/ApiControllerAnalyzer.cs
twsouthwick: Let's chat about this tomorrow. There's a reference assembly nupkg that we can grab

twsouthwick (changes requested):

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/ApiControllerAnalyzer.cs
```diff
@@ -0,0 +1,108 @@
...
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
+
+            context.RegisterSyntaxNodeAction(AnalyzeNode, cs.SyntaxKind.SimpleBaseType);
+            context.RegisterSyntaxNodeAction(AnalyzeNode, vb.SyntaxKind.InheritsStatement);
+        }
+
+        private void AnalyzeNode(SyntaxNodeAnalysisContext context)
+        {
+            var namedTypeSymbol = context.ContainingSymbol as INamedTypeSymbol;
```
mjrousos, line 49: By the time this analyzer runs, UA will have already removed the System.Web dependency, so I wouldn't expect this symbol to be resolved. Have you tested this end-to-end in UA to see if it works?

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/ApiControllerAnalyzer.cs
```diff
@@ -0,0 +1,108 @@
...
+
+            if (namedTypeSymbol != null && namedTypeSymbol.BaseType != null
```
mjrousos, line 51: I may be overlooking something here, but if you're getting the symbol for a C# SimpleBaseType syntax, I would think the type you need to look at is the type of that symbol itself rather than its BaseType.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/ApiControllerAnalyzer.cs
```diff
@@ -0,0 +1,108 @@
...
+        /// do not want to highlight the VisualBasic.SyntaxKind.InheritsKeyword
+        /// </summary>
+        /// <param name="context"></param>
+        private static void ReportVisualBasicSyntax(SyntaxNodeAnalysisContext context)
+        {
+            var baseClass = context.Node.DescendantNodes()
+                .OfType<vb.Syntax.QualifiedNameSyntax>()
+                .FirstOrDefault() as SyntaxNode;
+
+            if (baseClass is null)
```
mjrousos, line 83: Out of curiosity, why is digging in to get the identifier name syntax needed in the VB code path but not the C# one?

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/ApiControllerAnalyzer.cs
mjrousos: Ah, I see you're adding that back. There's also a problem with removing the `using System.Web.Http` statement. We may need to turn that analyzer off (or at least run it later).

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/ApiControllerAnalyzer.cs
```diff
@@ -0,0 +1,108 @@
...
+                baseClass = context.Node.DescendantNodes()
+                .OfType<vb.Syntax.IdentifierNameSyntax>()
+                .FirstOrDefault();
+            }
+
+            var diagnostic = Diagnostic.Create(Rule, baseClass.GetLocation(), baseClass.ToString());
+            context.ReportDiagnostic(diagnostic);
+        }
+
+        public static Project AddMetadataReferences(Project project)
```
mjrousos, line 94: This isn't the right place for this method to live. The source updater shouldn't have a dependency on any particular extension.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source/SourceUpdaterStep.cs
```diff
@@ -159,6 +161,13 @@ public async Task GetDiagnosticsAsync(CancellationToken token)
             }
         }
 
+        public static Project AddMetadataReferences(Project project)
+        {
+            project = Extensions.Default.CSharp.Analyzers.ApiControllerAnalyzer.AddMetadataReferences(project);
```
mjrousos, line 166: The source updater step shouldn't depend on extensions (not even the default ones). I'm actually surprised that it has a transitive reference to that package.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/ApiControllerCodeFixer.cs
```diff
@@ -0,0 +1,89 @@
...
+                var namespaceIdentifier = docEditor.Generator.IdentifierName(GoodNamespace);
+                var classNameIdentifier = docEditor.Generator.IdentifierName(GoodClassName);
+                var controllerIdentifierSyntax = docEditor.Generator.QualifiedName(namespaceIdentifier, classNameIdentifier);
+                docEditor.ReplaceNode(node, controllerIdentifierSyntax);
+            }
+            else
+            {
+                var controllerIdentifierSyntax = docEditor.Generator.IdentifierName($"{GoodNamespace}.{GoodClassName}");
+                docEditor.ReplaceNode(node, controllerIdentifierSyntax);
+            }
```
mjrousos, line 81: It would be nice to add a `using` statement and use a simplifier annotation so that users end up with `: Controller` instead of `: Microsoft.AspNetCore.Mvc.Controller` which is a bit odd.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/ApiControllerCodeFixer.cs
```diff
@@ -0,0 +1,89 @@
...
+            var docEditor = await slnEditor.GetDocumentEditorAsync(document.Id, cancellationToken).ConfigureAwait(false);
+
+            ReplaceController(node, docEditor);
+
+            return docEditor.GetChangedDocument();
+        }
+
+        private static void ReplaceController(SyntaxNode node, DocumentEditor docEditor)
+        {
+            if (IsQualifiedNameSyntax(node))
```
mjrousos, line 70: Is this check really necessary? I would have thought you could just replace the node with a new `QualifiedName` regardless of its type.

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/AnalyzerTests.cs
```diff
@@ -190,26 +211,41 @@ public async Task UpgradeAnalyzers(string diagnosticId)
...
+            }
         }
 
-        private static void AssertDiagnosticsCorrect(IEnumerable<Diagnostic> diagnostics, ExpectedDiagnostic[] expectedDiagnostics)
+        private static void AssertDiagnosticsCorrect(IEnumerable<Diagnostic> diagnostics, IEnumerable<ExpectedDiagnostic> expectedDiagnostics)
         {
-            Assert.Equal(expectedDiagnostics.Length, diagnostics.Count());
+            Assert.Equal(expectedDiagnostics.Count(), diagnostics.Count());
             var count = 0;
             foreach (var d in diagnostics.OrderBy(d => d.Location.SourceSpan.Start))
```
mjrousos, line 243: This could be replaced by `Assert.Collection`

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/assets/TestClasses/UA0013.Fixed.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.Web.Http;
+using System.Collections.Generic;
+
+namespace TestProject.TestClasses
+{
+    public class ValuesController : Microsoft.AspNetCore.Mvc.Controller
```
mjrousos, line 6: Simplifying this to just 'Controller' would be ideal.

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/assets/TestClasses/UA0013.cs
```diff
@@ -0,0 +1,36 @@
...
+    public class MoviesController : System.Web.Http.ApiController
+    {
+        // GET api/values
+        public IEnumerable<string> GetValues()
+        {
+            return new[] { "Star Wars", "Iron Man", "Star Trek" };
+        }
+    }
+
+    public class NotAWebController : Foo.ApiController
```
mjrousos, line 24: Maybe include a `using Foo` at the beginning of this namespace so that the inheritance here can just be `: ApiController` (should work the same, but looks more ambiguous that way).

mjrousos (changes requested):

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/ApiControllerCodeFixer.cs
twsouthwick: If using the generator, you can have it auto-simplify to whatever is in scope with the given usings

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/ApiControllerAnalyzer.cs
KSchlobohm: Hadn't thought about sorting the analyzers yet and have not run this through as an integration test. Wanted to start the discussion about the assembly reference first and see if there was other feedback.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/ApiControllerAnalyzer.cs
KSchlobohm: I think the extension needs a way to declare a requirement of what it needs to run. Do you have another recommendation about how to declare this requirement?

Also, I liked that the requirement of the Analyzer is declared with the Analyzer but I can see that it could be more clearly highlighted.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source/SourceUpdaterStep.cs
KSchlobohm: Maybe I'm too attached to the word "requirement", maybe the SourceUpdater step should check to see if the analyzer needs to be initialized and adding the metadataReference can be part of initialization

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/ApiControllerCodeFixer.cs
KSchlobohm: I was thinking about that too but decided to go with this simple approach pending our design review.

One idea surfaced is the idea of building a more composable and abstract approach. If we use that approach, then the code here would just be part of the overall solution and could become the BaseTypeMapper or something similar. To achieve the idea you're mentioning I think we would combine the BaseTypeMapper with the NamespaceMapper.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/ApiControllerCodeFixer.cs
KSchlobohm: I'll double check the unit tests because I was seeing weird behavior from having the vbproj file include both test and fixed files. At the time, it appeared that the replacement was failing because it wasn't replacing a node of the same type.

PR #491, opened 2021-05-04
Add RazorSourceUpdater tests

### LINKED ISSUE #456: As a developer, I'm confident my changes won't break RazorSourceUpdater or its associated services

mjrousos:
Fixes  #456


src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorTextReplacer.cs
```diff
@@ -34,36 +34,57 @@ public RazorTextReplacer(ILogger<RazorTextReplacer> logger)
...
-                    MinimizeReplacement(ref originalText, ref updatedText);
+
+                    // Generally, it's not necessary to minimize replacements since the text should be the same in both documents
+                    // However, in the specific case of text being added before the first #line pragma in a Razor doc, it's possible
+                    // that C# unrelated to the original document will show up with the replaced text, so minimize replacements only
+                    // in the case that text is being inserted before the original first line.
+                    if (replacement.StartingLine == 0)
+                    {
+                        MinimizeReplacement(ref originalText, ref updatedText);
+                    }
```
mjrousos, line 80: I'm still not entirely happy with how we minimize replacements. I'm going to re-work this but will do it in a separate PR since this is focused on implementing tests.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IUpgradeContextProperties.cs
```diff
@@ -1,12 +1,7 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-using System;
-using System.Collections;
 using System.Collections.Generic;
```
twsouthwick, line 4: I wish an analyzer would flag these so we don't have to do it manually periodically....

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorTextReplacer.cs
```diff
@@ -104,22 +129,32 @@ public void ApplyTextReplacements(IList<TextReplacement> replacements)
...
+            if (text.Any())
             {
-                // StreamSourceDoc.Lines is 0-based but line directives (as used in MappedSubText) are 1-based,
-                // so subtract one from i.
-                offset += lines[i - 1].Length + Environment.NewLine.Length;
+                yield return 0;
+            }
+
+            // Subsequent lines
+            for (var i = 0; i < text.Length; i++)
```
twsouthwick, line 144: I wonder if using `string.IndexOf(...)` would be more readable, and may be more efficient (I've seen some simd related implementations behind the scenes)

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/TextReplacement.cs
```diff
@@ -78,8 +80,10 @@ public TextReplacement(SourceText originalText, SourceText newText, string fileP
         public override int GetHashCode()
         {
             var hashcode = default(HashCode);
+            hashcode.Add(StartingLine);
+            hashcode.Add(FilePath, StringComparer.OrdinalIgnoreCase);
             hashcode.Add(NewText.ToString(), StringComparer.Ordinal);
```
twsouthwick, line 85: calling `.ToString` seems like it may be a potentially heavy operation. Does NewText not implement `GetHashCode()`?

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/TextReplacement.cs
```diff
@@ -78,8 +80,10 @@ public TextReplacement(SourceText originalText, SourceText newText, string fileP
...
-            hashcode.Add(OriginalText);
+            hashcode.Add(OriginalText.ToString(), StringComparer.Ordinal);
```
twsouthwick, line 86: Same as above

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor.Tests/DefaultTextMatcherTests.cs
```diff
@@ -0,0 +1,177 @@
...
+
+            // Act
+            var replacements = matcher.MatchOrderedSubTexts(originalTexts, newTexts);
+
+            // Assert
+            Assert.Collection(replacements, expectedReplacements.Select<TextReplacement, Action<TextReplacement>>(expected => actual => Assert.Equal(expected, actual)).ToArray());
+        }
+
+        public static IEnumerable<object[]> MatchOrderedSubTextsData =>
+            new List<object[]>
```
twsouthwick, line 53: nit: I've often used a `yield return` syntax here. Seems like that would be slightly more readable

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor.Tests/MappedSubTextTestData.cs
```diff
@@ -0,0 +1,73 @@
+﻿namespace Razor
```
twsouthwick, line 1: I'm assuming this is generated. Can you make it clear so that we don't change it manually? Maybe comments at the top or a directory with RazorGeneratedSource or something

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor.Tests/Microsoft.DotNet.UpgradeAssistant.Steps.Razor.Tests.csproj
```diff
@@ -10,8 +10,22 @@
...
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </Content>
+  </ItemGroup>
+
+  <ItemGroup>
+    <Compile Remove="MappedSubTextTestData.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <None Remove="RazorUpdaterStepViews\TestViews\Simple.cshtml" />
```
twsouthwick, line 23: Maybe glob all cshtml to be removed

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor.Tests/RazorSourceUpdaterTests.cs
```diff
@@ -0,0 +1,387 @@
...
+        }
+
+        [Theory]
+        [MemberData(nameof(ApplyData))]
+        public async Task ApplyTests(LocationLookup[][] diagnosticLocations, string[] expectedUpdatedFiles, TextReplacement[] expectedReplacements)
+        {
+            // Arrange
+            using var mock = GetMock("RazorUpdaterStepViews/Test.csproj", diagnosticLocations);
+            var razorDocs = await GetRazorCodeDocumentsAsync(mock).ConfigureAwait(true);
+            var context = mock.Mock<IUpgradeContext>();
```
twsouthwick, line 183: nit: `mock.Mock<>()` should only be used when it's for DI. If it's something you're passing into a message, it can cause problems (ie, all calls to `mock.Mock<>()` will return the same instance). Otherwise, you can use `new Mock<>()` to get a separate instance.

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor.Tests/RazorSourceUpdaterTests.cs
```diff
@@ -0,0 +1,387 @@
...
+                else
+                {
+                    builder.RegisterInstance(Enumerable.Empty<DiagnosticAnalyzer>());
+                    builder.RegisterInstance(Enumerable.Empty<CodeFixProvider>());
+                }
+            });
+
+            var projectFile = mock.Mock<IProjectFile>();
+            var project = projectPath is not null ? mock.Mock<IProject>() : null;
+            project?.Setup(p => p.FileInfo).Returns(new FileInfo(projectPath!));
```
twsouthwick, line 361: this may be more readable if you just do 

```csharp
if(projectPath is not null)
{
.....
}
```

twsouthwick (approved):
A couple of comments, but LGTM

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IUpgradeContextProperties.cs
mjrousos: I agree!

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorTextReplacer.cs
mjrousos: Sure, I can do that. I'd say either way is similarly readable in my opinion.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/TextReplacement.cs
mjrousos: Hm.. that's a good point. I don't think `SourceText` (which NewText is) has a GetHashCode, though, since it doesn't implement `IEquatable`.  Now that  look at the code, though, I'll just make the property a string type. We're not using any `SourceText` APIs (the type is just an artifact of where the data comes from in UA's workflow). That way, at least, we won't have multiple ToString calls.

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor.Tests/DefaultTextMatcherTests.cs
mjrousos: Good idea. I'll try that.

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor.Tests/MappedSubTextTestData.cs
mjrousos: Yep, good point. I'll add a comment.

PR #515, opened 2021-05-10
api analyzer roslyn

KSchlobohm:
**PR Title**
Language Agnostic Code Fixer and Code Analyzer for ApiController

**PR Description**
* Provides code mapping from *System.Web.ApiController* to *Microsoft.AspNetCore.Mvc.Controller*
* Adds integration test scenario for ApiControllerAnalyzer

Addresses #463


tests/tool/Integration.Tests/IntegrationScenarios/AspNetSample/csharp/Original/MyControllerLibrary/MyControllerLibrary.csproj
```diff
@@ -0,0 +1,49 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
```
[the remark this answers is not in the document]
KSchlobohm: I think I agree we don't need these post merge. But I plan to leave them in until the PR is approved to accelerate my inner loop.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/ApiControllerCodeFixer.cs
```diff
@@ -0,0 +1,96 @@
...
+            var newRoot = generator.ReplaceNode(root, node, mvcBaseType);
+
+            var newDocument = document.WithSyntaxRoot(newRoot);
+
+            newDocument = await ImportAdder.AddImportsAsync(newDocument, Simplifier.AddImportsAnnotation, cancellationToken: cancellationToken).ConfigureAwait(false);
+
+            return newDocument;
+        }
+
+        private static SyntaxNode CreateMvcNamespace(SyntaxGenerator generator)
```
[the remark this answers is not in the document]
KSchlobohm: I'll update them to use the => syntax

separating them seemed more self describing than combining them all together

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/AnalyzerExtensions.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        /// <summary>
+        /// Review the Attributes of the <paramref name="analyzer"/> to determine
+        /// if which languages apply.
+        /// </summary>
+        /// <param name="analyzer">True</param>
+        /// <returns>A list of supported languages.</returns>
+        private static IEnumerable<string> GetSpecifiedLanguage(Type analyzer)
+        {
+            var analyzerAttr = analyzer.CustomAttributes.FirstOrDefault(a => a.AttributeType.FullName.Equals(typeof(DiagnosticAnalyzerAttribute).FullName, StringComparison.Ordinal));
```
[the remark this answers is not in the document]
KSchlobohm: specifically comparing names. The attribute for an analyzer expects a string. We populate that string from the static class `Microsoft.CodeAnalysis.LanguageNames` rather than the type `Microsoft.DotNet.UpgradeAssistant.Language`

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/ApiControllerAnalyzer.cs
```diff
@@ -0,0 +1,67 @@
...
+            if (baseType is null)
+            {
+                return;
+            }
+
+            // Find just the named type symbols with names containing lowercase letters.
+            if (baseType.ToDisplayString().Equals("System.Web.Http.ApiController", StringComparison.Ordinal))
+            {
+                // For all such symbols, produce a diagnostic.
+                var node = namedTypeSymbol.DeclaringSyntaxReferences[0].GetSyntax();
```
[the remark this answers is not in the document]
KSchlobohm: There is no [SymbolKind](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.symbolkind?view=roslyn-dotnet-3.9.0) for baseType so `RegisterSymbolAction` has to use *SymbolKind.NamedType` to identify what we're interested in analyzing.

Though, I think your point is more focused on comparing the value between Syntax Analysis and Semantic Analysis for this Analyzer.

If we used Syntax Analysis then when we see *ApiController* we would have to evaluate the using statements in the document to determine if that *ApiController* is actually from *System.Web.Http*. In VB the language uses the import key word. To do this in a language agnostic way we would need a SyntaxGenerator which was not accessible without creating the *Document* or *AdHocWorkspace*. In contrast, using Sematic Analysis to determine if the symbol resolves to the well-known type simplifies this language specific challenge.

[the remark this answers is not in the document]
KSchlobohm: @mjrousos, please let me know if i missed the intent of your feedback.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/ApiControllerAnalyzerCSharp.cs
```diff
@@ -0,0 +1,43 @@
...
+            {
+                return;
+            }
+
+            if (node is null)
+            {
+                throw new ArgumentNullException(nameof(node));
+            }
+
+            var baseClass = node.DescendantNodes()
```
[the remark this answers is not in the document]
KSchlobohm: we could if we changed what we're syntactically highlighting.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/ApiControllerAnalyzer.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            if (context is null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
+
+            context.RegisterSymbolAction(AnalyzeSymbols, SymbolKind.NamedType);
```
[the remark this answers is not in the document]
KSchlobohm: validated that it does not trigger if we have multiple instances of a type used in a single class file

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/AnalyzerExtensions.cs
twsouthwick, line 40: This may be a very expensive operation if done within an analyzer. What scenario would this be used in?

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/ApiControllerAnalyzer.cs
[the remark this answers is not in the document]
twsouthwick: This is a named type which I believe is the symbol bound to type definitions

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/ApiControllerCodeFixer.cs
```diff
@@ -0,0 +1,100 @@
...
+            var types = generator.GetBaseAndInterfaceTypes(node);
+            if (types == null || types.Count == 0)
+            {
+                types = generator.GetBaseAndInterfaceTypes(node.Parent);
+            }
+
+            return types[0];
+        }
+
+        private static SyntaxNode CreateMvcNamespace(SyntaxGenerator generator) => generator.QualifiedName(CreateAspNetCoreNamesapce(generator), generator.IdentifierName("Mvc"));
```
twsouthwick, line 94: Could these be generalized? So we could pass in a fully qualified name and it spits out the syntax node we want?

I could foresee this being used to replace additional base types

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/LanguageExtensions.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+
+namespace Microsoft.DotNet.UpgradeAssistant
+{
+    public static class LanguageExtensions
```
twsouthwick, line 8: Does this need to be in the abstractions?

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/AnalyzerExtensions.cs
```diff
@@ -0,0 +1,66 @@
...
+using System.Linq;
+using System.Text;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers
+{
+    public static class AnalyzerExtensions
+    {
+        public static bool IsCorrectLanguage(this DiagnosticAnalyzer analyzer, Compilation comp)
```
twsouthwick, line 15: Is this only used in tests?

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/ApiControllerAnalyzer.cs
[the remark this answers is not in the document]
mjrousos: Got it. Thanks for checking, @KSchlobohm.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/AnalyzerExtensions.cs
KSchlobohm: actually, we don't need this entire file. I had built this to create a guard clause. While running our unit tests we were loading both VB and C# child class analyzers which don't exist anymore. I'll remove the file.

In hindsight, I'm not sure why I used reflection here instead of a constant in the child class.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/ApiControllerCodeFixer.cs
KSchlobohm: Yeah, I'm going to replace this with a QualifiedName helper that will enable us to build the syntax from string. I think we could evolve that going forward to do some other QualifiedName analysis

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/LanguageExtensions.cs
KSchlobohm: Looks like I only used this in the test project. I'll move it there and we can shift it around later if we want access in other locations.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/ApiControllerAnalyzer.cs
```diff
@@ -0,0 +1,83 @@
...
+        private static readonly LocalizableString MessageFormat = new LocalizableResourceString(nameof(Resources.ApiControllerDescription), Resources.ResourceManager, typeof(Resources));
+        private static readonly LocalizableString Description = new LocalizableResourceString(nameof(Resources.ApiControllerDescription), Resources.ResourceManager, typeof(Resources));
+
+        protected static readonly DiagnosticDescriptor Rule = new(DiagnosticId, Title, MessageFormat, Category, DiagnosticSeverity.Warning, isEnabledByDefault: true, description: Description);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(Rule);
+
+        public override void Initialize(AnalysisContext context)
+        {
+            if (context is null)
```
CyrusNajmabadi, line 31: fwiw, this should not ever happen.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/ApiControllerAnalyzer.cs
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        private static bool IsBaseTypeAnImplicitReferenceToApiController(INamedTypeSymbol baseType)
+        {
+            return baseType.ToDisplayString().Equals(ApiControllerClassName, StringComparison.Ordinal) && baseType.TypeKind == TypeKind.Error;
+        }
+
+        private static bool IsBaseTypeAQualifiedReferenceToApiController(INamedTypeSymbol baseType)
+        {
+            return baseType.ToDisplayString().Equals($"{ApiControllerNamespace}.{ApiControllerClassName}", StringComparison.Ordinal);
```
CyrusNajmabadi, line 68: you might want to consider making constants out of these concatenated strings.  otherwise you'll generate this for every type you look at.  allocations can affect perf.  but as you may be a run-once set of analyzers, that might be ok.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/ApiControllerCodeFixer.cs
```diff
@@ -0,0 +1,100 @@
...
+        public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+
+            if (root is null)
+            {
+                return;
+            }
+
+            var node = root.FindNode(context.Span, false, true);
```
CyrusNajmabadi, line 39: i recommend named arguments for these constants.  i, for one, forget what they are without the names :)

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/ApiControllerCodeFixer.cs
```diff
@@ -0,0 +1,100 @@
...
+        {
+            // See https://github.com/dotnet/roslyn/blob/master/docs/analyzers/FixAllProvider.md for more information on Fix All Providers
+            return WellKnownFixAllProviders.BatchFixer;
+        }
+
+        public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+
+            if (root is null)
```
CyrusNajmabadi, line 34: should be impossible.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/ApiControllerCodeFixer.cs
```diff
@@ -0,0 +1,100 @@
...
+            {
+                return;
+            }
+
+            var node = root.FindNode(context.Span, false, true);
+
+            if (node is null)
```
CyrusNajmabadi, line 41: should be impossible.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/ApiControllerCodeFixer.cs
```diff
@@ -0,0 +1,100 @@
...
+
+        private static SyntaxNode GetBaseType(SyntaxNode node, SyntaxGenerator generator)
+        {
+            var types = generator.GetBaseAndInterfaceTypes(node);
+            if (types == null || types.Count == 0)
+            {
+                types = generator.GetBaseAndInterfaceTypes(node.Parent);
+            }
+
+            return types[0];
```
CyrusNajmabadi, line 91: note: This may not work in the cae of a partial type.  Consider a partial type like so:

```
partial class Foo {
}

partial calss Foo : TheBaseType {
}
```

If you are running on the part that has no base types, this will index out of bounds.  A way to address this is to make sure on the analyzer side that this helper returns a non-null node.  as per our convo, you could also then just store that node in the AdditionalLocations part of hte diagnostic to make it easy to find in the fixer.

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/LanguageExtensions.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    public static class LanguageExtensions
+    {
+        public static string GetFileExtension(this Language lang)
+        {
+            return lang switch
+            {
+                Language.CSharp => "cs",
+                Language.VisualBasic => "vb",
+                Language.FSharp => "fs",
```
CyrusNajmabadi, line 16: interesting.  do you support f#?

CyrusNajmabadi (commented):
Overall, looks fine to me.  Some things that could be tweaked.  And def considr partial types.  In general, if the fixer could do something that could fail, it's best if hte analyzer checks for that first.  

looks like you had the same thoughts as me that some of these pieces could be generalized out further to help other analyzers in the future.

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/LanguageExtensions.cs
KSchlobohm: nope, we have that in the enumeration so I figured I'd add it here to the extension method.

starting to think we should remove fsharp references so that the expectation is clear.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/ApiControllerAnalyzer.cs
KSchlobohm: I usually build these because we have [CA062](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1062) enabled.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/ApiControllerCodeFixer.cs
KSchlobohm: Actually, I only included these because I borrowed that from another analyzer. Unit test shows we don't need these so I'll drop them.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/ApiControllerCodeFixer.cs
KSchlobohm: updated unit test to include this scenario and reworked the analyzer to detect it

KSchlobohm:
Still planning to remove the integration test changes which increase cycle time and are covered by unit tests.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/ApiControllerAnalyzer.cs
```diff
@@ -0,0 +1,110 @@
...
+
+            // the first descendent of ImportStatementSyntax will be a QualifiedNameSyntax or IdentifierNameSyntax for VB
+            // the first descendent of BaseListSyntax will be a SimpleBaseTypeSyntax for CS
+            var baseTypeNode = context.Node.DescendantNodes().First();
+
+            if (baseTypeNode.IsKind(CodeAnalysis.CSharp.SyntaxKind.SimpleBaseType))
+            {
+                // In CSharp syntax, the SimpleBaseTypeSyntax the first child and we want the QualifiedNameSyntax or IdentifierNameSyntax
+                // resolving this node must be done before we check for symbol info
+                baseTypeNode = baseTypeNode.DescendantNodes().First();
```
KSchlobohm, line 64: the integration test identified that the symbol for ApiController was not correctly resolved when i tried to use the SimpleBaseType syntax node. asking for the symbol again when using the child node resolved the false positive i added as EmailApi in the test project.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/ApiControllerCodeFixer.cs
```diff
@@ -0,0 +1,85 @@
...
+            }
+
+            var node = root.FindNode(context.Span);
+
+            if (node.IsKind(CodeAnalysis.CSharp.SyntaxKind.SimpleBaseType))
+            {
+                // When the analyzer runs it selects a portion of the code to be fixed by this class.
+                // In CSharp syntax, the SimpleBaseTypeSyntax is at the same location as the QualifiedNameSyntax or IdentifierNameSyntax
+                // since we just found this node by text location, we need to adjust what we selected to match our assumption for CSharp
+                node = node.DescendantNodes().First();
```
KSchlobohm, line 47: was looking for, but did not find, a way to pass a reference to the SyntaxNode from the Analyzer to the Code Fixer. the only overloads I saw were location based details.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/ApiControllerAnalyzer.cs
```diff
@@ -0,0 +1,110 @@
...
+            {
+                // we found a symbol, and the symbol information tells us this isn't an ApiController
+                // bail out
+
+                // note: at this point of upgrade assistant's workflow we are expecting the System.Web.Http reference
+                // to have been removed so if we didn't find the symbol, that's okay
+                return;
+            }
+
+            if (IsBaseTypeAQualifiedReferenceToApiController(baseTypeNode.ToString())
```
mjrousos, line 79: Will ToString include trivia? Probably not. But if it does, we should get the syntax w/o trivia so we don't miss the match because of a trailing comment or something like that.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/ApiControllerCodeFixer.cs
```diff
@@ -0,0 +1,90 @@
...
+            }
+
+            var node = root.FindNode(context.Span);
+
+            if (node.IsKind(CodeAnalysis.CSharp.SyntaxKind.SimpleBaseType))
+            {
+                // When the analyzer runs it selects a portion of the code to be fixed by this class.
+                // In CSharp syntax, the SimpleBaseTypeSyntax is at the same location as the QualifiedNameSyntax or IdentifierNameSyntax
+                // since we just found this node by text location, we need to adjust what we selected to match our assumption for CSharp
+                node = node.DescendantNodes().FirstOrDefault();
```
mjrousos, line 47: Again, using an API on SymbolBaseTypeSyntax if possible seems like it might be better here.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/QualifiedNameBuilder.cs
```diff
@@ -0,0 +1,60 @@
...
+            {
+                throw new ArgumentNullException(nameof(nameSegments));
+            }
+            else if (nameSegments.Length == 1)
+            {
+                return generator.IdentifierName(nameSegments[0]);
+            }
+
+            var rightNode = generator.IdentifierName(nameSegments.Last());
+            var leftNode = BuildQualifiedSyntax(generator, nameSegments.Take(nameSegments.Length - 1).ToArray());
```
mjrousos, line 56: Can we use `Span<string>` here to avoid creating a bunch of arrays?

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/TestHelper.cs
```diff
@@ -81,27 +91,33 @@ private static async Task<IEnumerable<Diagnostic>> GetDiagnosticsFromProjectAsyn
...
         {
             if (documentPath is null)
             {
                 throw new ArgumentNullException(nameof(documentPath));
             }
 
             using var workspace = MSBuildWorkspace.Create();
-            var project = await workspace.OpenProjectAsync(TestProjectPath).ConfigureAwait(false);
+            var projectLanguage = lang.GetFileExtension();
+            var path = TestProjectPath.Replace("{lang}", $"Fixed.{projectLanguage}", StringComparison.OrdinalIgnoreCase);
```
mjrousos, line 103: Is this supposed to be including 'Fixed.'? It looks like previously, this would just give the path specified without interjecting 'fixed'.

tests/tool/Integration.Tests/IntegrationScenarios/AspNetSample/csharp/Original/MyControllerLibrary/MyControllerLibrary.csproj
```diff
@@ -0,0 +1,49 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
mjrousos, line 1: Unit tests are probably sufficient. I think you're planning to remove these pre-merge, though.

mjrousos (approved):
Looks good! A few comments/questions inline.

tests/tool/Integration.Tests/IntegrationScenarios/AspNetSample/csharp/Original/TemplateMvc/App_Start/WebApiConfig.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Web.Http;
+
+namespace TemplateMvc
```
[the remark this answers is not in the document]
KSchlobohm: updated the integration test scenario to make my inner loop more effective. this change adds Web API support to the TemplateMvc app.

tests/tool/Integration.Tests/IntegrationScenarios/AspNetSample/csharp/Original/TemplateMvc/Global.asax.cs
```diff
@@ -13,9 +14,10 @@ public class MvcApplication : System.Web.HttpApplication
         protected void Application_Start()
         {
             AreaRegistration.RegisterAllAreas();
+            GlobalConfiguration.Configure(WebApiConfig.Register);
             FilterConfig.RegisterGlobalFilters(GlobalFilters.Filters);
             RouteConfig.RegisterRoutes(RouteTable.Routes);
-            BundleConfig.RegisterBundles(BundleTable.Bundles);
+            //BundleConfig.RegisterBundles(BundleTable.Bundles);
```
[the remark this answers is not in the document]
KSchlobohm: When I modified the TemplateMvc app I noted it didn't actually start. The route registration fixes Web Api mapping and the other reason the app didn't start seemed to be due to missing Scripts folder.

Again, unrelated to the intent of this PR. Will revert today.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/ApiControllerAnalyzer.cs
```diff
@@ -0,0 +1,110 @@
...
+            var node = context.Node; // an ImportStatementSyntax for VB or a BaseListSyntax for CS
+
+            if (!node.DescendantNodes().Any())
+            {
+                return;
+            }
+
+            // the first descendent of ImportStatementSyntax will be a QualifiedNameSyntax or IdentifierNameSyntax for VB
+            // the first descendent of BaseListSyntax will be a SimpleBaseTypeSyntax for CS
+            var baseTypeNode = context.Node.DescendantNodes().First();
```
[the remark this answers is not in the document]
KSchlobohm: Yes, we could have used a DescendentNodes.OfType filter.

I felt the flow looked simpler without additional C# and VB specific syntax noise. LMK if you want to get specific about this.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/ApiControllerAnalyzer.cs
KSchlobohm: I'll add a comment that *ToString()* is intentional here because we don't want Trivia. The method *ToFullString()* would include trivia.

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/TestHelper.cs
KSchlobohm: Yes, I found that there was a behavioral difference in the testing (when adding VB support) surfacing from 2 classes having exactly the same name, included in the same project file, but not being partials.

You'll see now that we have 2 csproj files and 2 vbproj files to manage the before and after projects.

tests/tool/Integration.Tests/IntegrationScenarios/AspNetSample/csharp/Original/MyControllerLibrary/MyControllerLibrary.csproj
KSchlobohm: Of note: switching from 2 projects with a solution file back to a single project integration test scenario changes the runtime of that integration test from an average of 2min to an average of 1min on my machine.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/QualifiedNameBuilder.cs
KSchlobohm: big thanks Mike, I forgot that's where slice was!

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/ApiControllerAnalyzer.cs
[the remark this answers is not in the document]
KSchlobohm: made more sense as an approach when I considered the fragility of depending on a specific node tree and the possibility of abstracting away language syntax to extension methods that focused on intent rather than syntax.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/ApiControllerCodeFixer.cs
KSchlobohm: same as feedback on ApiControllerAnalyzer. now incorporated.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/ApiControllerCodeFixer.cs
twsouthwick: We have nullable reference types on so there may be a problem with the annotation of the result here that could be fixed in roslyn.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/ApiControllerCodeFixer.cs
```diff
@@ -0,0 +1,83 @@
...
+        public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
+        {
+            var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+
+            if (root is null)
+            {
+                return;
+            }
+
+            var baseTypeNode = ApiControllerAnalyzer.GetBaseTypeFromSyntax(root.FindNode(context.Span));
```
twsouthwick, line 40: This method may be better as an extension method outside of the analyzer itself

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/QualifiedNameBuilder.cs
```diff
@@ -0,0 +1,60 @@
...
+            {
+                throw new ArgumentNullException(nameof(generator));
+            }
+
+            if (qualifiedName is null || string.IsNullOrWhiteSpace(qualifiedName))
+            {
+                throw new ArgumentNullException(nameof(qualifiedName));
+            }
+
+            return BuildQualifiedSyntax(generator, qualifiedName.Split('.'));
```
twsouthwick, line 36: You're passing it into a method that takes a span, but you're doing arrays here. Can you switch this to `qualifiedName.AsSpan()....`?

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/QualifiedNameBuilder.cs
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        private static SyntaxNode BuildQualifiedSyntax(SyntaxGenerator generator, Span<string> nameSegments)
```
twsouthwick, line 39: Probably actually want `ReadOnlySpan<string>`

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/LanguageExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+
+using System;
+
+namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test
+{
+    public static class LanguageExtensions
+    {
+        public static string GetFileExtension(this Language lang)
+        {
+            return lang switch
```
twsouthwick, line 12: nit: Expression syntax here would be simpler:

```csharp
public static string GetFileExtension(this Language lang) => lang swich
{
  ...
}
```

twsouthwick (approved):
A few comments, but LGTM

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/QualifiedNameBuilder.cs
KSchlobohm: swapped to ReadOnlySpan for implicit conversion

Issue #510, opened 2021-05-07
After successful upgrade execution clean up undesired changes


IEvangelist:
## Summary

I propose that after a successful upgrade has been executed, that only required changes / files remain. Currently, this is not the case. For example, after running an upgrade:

- There is a _.upgrade-assitant_ JSON file at the root, this file is no longer needed and should be deleted
- There are `<PackageReference>` entries for upgrade specific analyzers, these should be removed

```xml
<PackageReference Include="Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers" Version="0.2.222702">
      <PrivateAssets>all</PrivateAssets>
</PackageReference>
```

## Motivation and goals

I understand that as part of the execution of the upgrade, peripheral changes and files are required. But after, these should be removed or undone.

## In scope

Add a clean up routine which is able to spin through projects removing package references that were added, and delete any files that were intended to be temporary.

## Out of scope

Nothing comes to mind here.

## Risks / unknowns

None.

## Examples

I ran the upgrade assistant with a project that was targeting `netcoreapp3.0`. It upgrade to `net5.0`, but left a _.upgrade-assitant_ file at the root of the .sln directory, and left a `<PackageReference>` entry - which should be no longer needed.

Somewhat related to #386, but more detailed and larger scope as it proposes a clean up routine.

mjrousos:
Removing .upgrade-assistant may make sense. @twsouthwick, is there anything in that file that needs to persist after upgrade is finalized?

The analyzers being left is intentional, though. In the (hopefully near) future, we're planning to include analyzers in that package that will alert users to APIs that they need to manually migrate away from that the tool wasn't able to handle automatically. Having the package included in the project will allow us to add diagnostics for these sorts of scenarios.

sunandabalu:
Making the finalize solution step to provide an option for the user to confirm that its Ok to clean up all traces of upgrade-assistant from the solution, and remove if user chooses so would make sense. @IEvangelist Feel free to create a PR for the solution mentioned here, we will review it.

IEvangelist:
Hi friends - @sunandabalu @mjrousos and @twsouthwick - please see #522

sunandabalu:
#522 is merged, closing this

HughWarrington:
[...]

PR #522, opened 2021-05-13
Remove the .upgrade-assistant JSON file when complete, and only create it when needed.

IEvangelist:
**PR Title**

Remove the _.upgrade-assistant_ JSON file when complete, and only create it when needed.

**PR Description**

- Added logic to avoid writing the _.upgrade-assistant_ file when it's in an empty state.
- Added logic to delete the file when the upgrade is fully completed.
- When running integration tests, noticed a few times that the order was incorrect and it would report false negative - use explicit ordering instead
- In `ConsoleUpgrade` use switch expression instead of switch statement, refactored to reduce total lines of code by eight.
- In `UpgradeRunner` use proper overload instead of `(_, services)`

Addresses #510


src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.NuGetPackages.cs
```diff
@@ -156,8 +156,9 @@ private IEnumerable<LockFileTargetLibrary> GetAllDependencies(TargetFrameworkMon
             var lockFile = LockFileUtilities.GetLockFile(LockFilePath, NuGet.Common.NullLogger.Instance);
 
             return lockFile.Targets
-                .First(t => t.TargetFramework.DotNetFrameworkName.Equals(parsedTfm.DotNetFrameworkName, StringComparison.Ordinal))
-                .Libraries;
+                .FirstOrDefault(t => t.TargetFramework.DotNetFrameworkName.Equals(parsedTfm.DotNetFrameworkName, StringComparison.Ordinal))
```
twsouthwick, line 159: The fix for this is actually a little different - I'm planning on implementing that today

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/ConsoleUpgrade.cs
```diff
@@ -151,28 +151,20 @@ await foreach (var step in steps.ToAsyncEnumerable().WhereAwait(async s => await
 
         private void WriteStepStatus(UpgradeStep step, bool isNextStep)
         {
-            switch (step.Status)
+            (ConsoleColor? color, var output) = step.Status switch
```
twsouthwick, line 154: Awesome!

twsouthwick (commented):
Can you add a comment to the finalize step that sets the `IsComplete=true` that finalizing will remove any progress tracked by Upgrade Assistant?

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.NuGetPackages.cs
IEvangelist: Can I leave this as is for this PR, or do you want it removed from here?

twsouthwick: It'll be a merge conflict most likely, so if you could remove it, that would be ideal

IEvangelist: Removed

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Solution/FinalizeSolutionStep.cs
```diff
@@ -29,6 +29,7 @@ public FinalizeSolutionStep(ILogger<FinalizeSolutionStep> logger)
 
         protected override Task<UpgradeStepApplyResult> ApplyImplAsync(IUpgradeContext context, CancellationToken token)
         {
+            // Finalizing will remove any progress tracked by Upgrade Assistant.
```
twsouthwick, line 32: Sorry - I meant to surface it to the user - probably in the description from the initialize method. We want it to be clear to them that if they select to finalize the solution, it will remove anything tracked by Upgrade Assistant

IEvangelist: Ah, ok - will do.

twsouthwick (approved):
LGTM! Thanks for the contribution @IEvangelist

twsouthwick:
I reworded the message slightly, but otherwise LGTM. I'll merge it once the build completes

PR #524, opened 2021-05-13
Implement zip archive for extensions

### LINKED ISSUE #490: Enable extensions to load zip files
## Summary

Extensions currently work best when given a path. This however, is not great for redistribution of extensions. We should have support for an archive format (ie zip) that could be used as well.

## Motivation and goals

Right now, distribution of an extension requires it to be a directory, but it would simplify things if it could just be an archived format.

## In scope

- Support for zip format

## Out of scope

- Distribution of the zip files

## Risks / unknowns

There may be issues with loading assemblies from these directly, so they may need to be expanded to disk. If so, zip bombs may be an issue.

twsouthwick:
This change enables zip archives for extensions. While doing this, it became simpler to not have an `IExtension` but just a POCO that holds the extension info (`IFileProvider` and `IConfiguration`) so that the only zip-related change was the implementation of IFileProvider.

Fixes #490


mjrousos (approved):
Looks great! Would be good to add some unit tests.

PR #531, opened 2021-05-18
Add hintpath to Reference object

twsouthwick:


src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildExtensions.cs
```diff
@@ -19,8 +19,34 @@ public static NuGetReference AsNuGetReference(this ProjectItemElement item)
             return new NuGetReference(packageName, packageVersion);
         }
 
-        public static Reference AsReference(this ProjectItemElement item) =>
-            new(item.Include.Split(',').First());
+        public static string? GetMetadata(this ProjectItemElement item, string name)
```
KSchlobohm, line 22: I don't think we want to make this method public.

If someone wants the HintPath then a public method to access the HintPath is more intention revealing. The fact that the HintPath is stored as metadata is a detail that should be abstracted away from the caller.

In this scenario, the magic string is "HintPath", I'd prefer not to have to look for the list of magic strings and just have methods that clearly state their intent.

twsouthwick: This method in general is internal and are a collection of helper methods. Getting metadata from an element is something we may need to do multiple times for different causes with many different kinds of metadata.

twsouthwick: I've made it private and added a public method for hintpath so we could minimize knowledge of those magic strings

mjrousos (approved):
LGTM

PR #534, opened 2021-05-19
Refactor IdentifierUpgradeAnalyzer and IdentifierUpgradeCodeFixer

### LINKED ISSUE #514: Replace base type System.Web.Mvc.Controller with Microsoft.AspNetCore.Mvc.Controller
`ControllerBase`, also. These are primarily about enabling controller types that derived from these base classes.

mjrousos:
This updates `IdentifierUpgradeAnalyzer` and `IdentifierUpgradeCodeFixer` to work with both C# and VB and consolidates ApiController analyzers and code fixers to use those base classes.

It also expands UA0013 analyzers and code fixers to handle both WebApi and MVC controllers.

We talked about combining all the IdentifierUpgrade analyzers and code fixers into one. I'll make that change in a follow-up PR to keep things manageable.

Fixes #514


src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/SyntaxNodeExtensions.cs
```diff
@@ -0,0 +1,212 @@
...
+        public static bool HasUsingStatement(this SyntaxNode node, string namespaceName)
+        {
+            if (node is null)
+            {
+                throw new ArgumentNullException(nameof(node));
+            }
+
+            var nodes = node.DescendantNodesAndSelf(_ => true, false);
+
+            return nodes.OfType<CSSyntax.UsingDirectiveSyntax>().Any(u => u.Name.ToString().Equals(namespaceName, StringComparison.Ordinal))
```
twsouthwick, line 188: This doesn't seem to track if it would be in scope. Maybe not a big deal, but it would be good to call that out as a corner case we're not worried about

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/AnalyzerExtensions.cs
```diff
@@ -98,33 +75,33 @@ public static void RegisterMemberAccess(this AnalysisContext context, Action<Inv
                 action(newCtx);
             }, SymbolKind.Property, SymbolKind.Method, SymbolKind.Parameter, SymbolKind.Field);
         }
+    }
 
-        public readonly struct InvocationAnalysisContext
-        {
-            private readonly Action<Diagnostic> _reportDiagnostic;
+    public record InvocationAnalysisContext
```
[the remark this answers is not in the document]
mjrousos: Good point. I'll switch this back to a struct.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/AnalyzerExtensions.cs
```diff
@@ -4,43 +4,20 @@
...
 using Microsoft.CodeAnalysis;
-using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 using Microsoft.CodeAnalysis.Operations;
 
-namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers
+namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Default
 {
-    internal static class DiagnosticAnalyzerExtensions
+    public static class AnalyzerExtensions
```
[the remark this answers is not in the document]
mjrousos: I updated all the extension classes to be public so that they can be shared between the analyzer project and the code fix provider project. This particular one is only used by analyzers, so we could move it back to the analyzer project and make it internal.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/IdentifierUpgradeAnalyzer.cs
```diff
@@ -31,46 +48,69 @@ public override void Initialize(AnalysisContext context)
...
 
-            // If the node isn't an identifier, bail out
-            if (name is null)
-            {
-                return;
-            }
+        private void AnalyzeVBIdentifier(SyntaxNodeAnalysisContext context)
+        {
+            var identifier = (VBSyntax.IdentifierNameSyntax)context.Node;
+            AnalyzeIdentifier(context, identifier.Identifier.ValueText);
```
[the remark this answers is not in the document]
mjrousos: They were separate because the `Identifier` property is defined separately on VB and C# identifier syntax types. Looking at it now, though, I can just get the identifier name from an extension method and register a single SyntaxNodeAction. I'll do that.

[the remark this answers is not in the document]
mjrousos: Actually, I still need to register to syntax node actions because those are language-specific, but I can still get the simple name in an extension method so that we can get rid of these methods.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/SyntaxNodeExtensions.cs
mjrousos: It doesn't. That's probably why its behavior is different from the simplifier (though, the simplifier apparently doesn't check scope either, it just defaults differently than we do - it assumes not in scope while this assumes in-scope).

Originally, I left it out as a low-pri corner case, but it's probably not too hard to do right. I'll update the method to at least make an effort to consider scope.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/QualifiedNameBuilder.cs
```diff
@@ -0,0 +1,69 @@
...
+            {
+                throw new ArgumentNullException(nameof(generator));
+            }
+
+            if (qualifiedName is null || string.IsNullOrWhiteSpace(qualifiedName))
+            {
+                throw new ArgumentNullException(nameof(qualifiedName));
+            }
+
+            return BuildRecursiveNameSyntax(generator, qualifiedName.Split('.'), generator.MemberAccessExpression);
```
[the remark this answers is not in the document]
mjrousos: Caching the arrays would only help if we're splitting the same strings repeatedly, but I can see how that could come up. I'll update it to cache the arrays.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/AnalyzerExtensions.cs
```diff
@@ -4,43 +4,20 @@
...
-            while (nameSyntax.Parent is QualifiedNameSyntax qualifiedParent)
-            {
-                nameSyntax = qualifiedParent;
-            }
-
-            return nameSyntax;
-        }
-
-        public static bool NameEquals(this IAssemblySymbol? symbol, string name, bool startsWith = true)
+        public static void RegisterMemberAccess(this AnalysisContext context, Action<InvocationAnalysisContext> action)
```
KSchlobohm, line 14: Does the `SystemDeploymentAnalyzer` really need to register for all of these triggers?

KSchlobohm: When should I reuse this method?

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/SyntaxNodeExtensions.cs
```diff
@@ -0,0 +1,212 @@
...
+                || nodes.OfType<VBSyntax.SimpleImportsClauseSyntax>().Any(i => i.Name.ToString().Equals(namespaceName, StringComparison.Ordinal));
+        }
+
+        /// <summary>
+        /// Adds a using directive for a given namespace to the document root only if the directive is not already present.
+        /// </summary>
+        /// <param name="documentRoot">The document to add the directive to.</param>
+        /// <param name="namespaceName">The namespace to reference with the using directive.</param>
+        /// <returns>An updated document root with the specific using directive.</returns>
+        public static CSSyntax.CompilationUnitSyntax AddUsingIfMissing(this CSSyntax.CompilationUnitSyntax documentRoot, string namespaceName)
```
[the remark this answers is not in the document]
mjrousos: It probably would be, but I may save that for a future update. This PR is just moving the API and it's only used by a couple analyzers (which may be changed themselves soon).

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/QualifiedNameBuilder.cs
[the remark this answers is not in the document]
mjrousos: After further discussion, I'll leave this as is for now since it's not a high traffic code path. I'll add a comment, though, that we likely want to update this in the future.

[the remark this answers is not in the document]
KSchlobohm: This split only occurs once during a code fixer action. Is the concern over about the caller being unaware of what this method does? If that's the case this overload could be removed and we would rely on the `BuildRecursiveNameSyntax` method.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/AnalyzerExtensions.cs
mjrousos: Good question. It does seem like the symbol action would be sufficient. I'm just moving this code, though, so I'm leaving it as is for now.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/SyntaxNodeExtensions.cs
```diff
@@ -0,0 +1,212 @@
...
+        /// Gets a method declared in a given syntax node.
+        /// </summary>
+        /// <typeparam name="T">The type of method declaration syntax node to search for.</typeparam>
+        /// <param name="node">The syntax node to find the method declaration in.</param>
+        /// <param name="methodName">The name of the method to return.</param>
+        /// <param name="requiredParameterTypes">An optional list of parameter types that the method must accept.</param>
+        /// <returns>The first method declaration in the syntax node with the given name and parameter types, or null if no such method declaration exists.</returns>
+        public static T? GetMethodDeclaration<T>(this SyntaxNode node, string methodName, params string[] requiredParameterTypes)
+            where T : SyntaxNode =>
+            node?.DescendantNodes()
```
KSchlobohm, line 155: i prefer not to use expression syntax for scenarios where a break point could be set on multiple lines. Specifically due to error logs that can surface exception by line number. All of this, from a compiler perspective is one line isn't it?

KSchlobohm: Could we make this language agnostic by dropping the generic type?

KSchlobohm: Am trying to remember, do you care about Node expansion here?

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/HttpContextCurrentCodeFixer.cs
```diff
@@ -173,7 +173,7 @@ private static void InitializeHttpContextHelperInStartup(DocumentEditor editor,
             }
 
             // Add Initialize call in Configure method
-            var configureMethod = documentRoot.GetMethodDeclaration("Configure", "IApplicationBuilder");
+            var configureMethod = documentRoot.GetMethodDeclaration<MethodDeclarationSyntax>("Configure", "IApplicationBuilder");
```
KSchlobohm, line 176: won't let me comment on the other section... but i noticed that we're doing `.WithWhitespaceTriviaFrom(configureServicesBody.Statements.First());`

can we always assume the configureServicesBody has a first statement?

KSchlobohm, line 176: again, doesn't seem like you care about the CSharp syntax type. you just care that you found the right node.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/UrlHelperAnalyzer.cs
```diff
@@ -86,10 +89,29 @@ private static string GetReplacementType(SyntaxNode node)
...
-                ObjectCreationExpressionSyntax => AspNetCoreUrlHelperName, // Object creation requires HtmlHelper
-                MemberAccessExpressionSyntax => AspNetCoreUrlHelperName, // Member access requires HtmlHelper
-                _ => AspNetCoreIUrlHelperName // Default to IHtmlHelper
+                ObjectCreationExpressionSyntax => AspNetCoreUrlHelperName, // Object creation requires UrlHelper
+                MemberAccessExpressionSyntax => AspNetCoreUrlHelperName, // Member access requires UrlHelper
+                _ => AspNetCoreIUrlHelperName // Default to IUrlHelper
             };
         }
+
+        private static SyntaxNode GetQualifiedName(SyntaxNode name)
```
KSchlobohm, line 98: I think we just added a helper method for this. see SyntaxNodeExtensions.cs

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/IdentifierUpgradeCodeFixer.cs
```diff
@@ -22,32 +21,44 @@ public abstract class IdentifierUpgradeCodeFixer : CodeFixProvider
...
                 return;
             }
 
-            var node = root.FindNode(context.Span);
+            // The node is found by location, so it's possible `FindNode` will find a parent with the same span.
+            // To get the correct node, find the first name or MAE in the given span.
+            var node = root.FindNode(context.Span)
+                .DescendantNodesAndSelf(n => true)
+                .FirstOrDefault(n => n.Span.Equals(context.Span)
+                    && (n is CSSyntax.NameSyntax
```
KSchlobohm, line 43: seems like a candidate for language agnostic helper functions

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/IdentifierUpgradeCodeFixer.cs
```diff
@@ -60,78 +71,80 @@ public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
...
-
-                updatedName = SyntaxFactory.ParseExpression(newIdentifier)
-                    .WithTriviaFrom(name)
-                    .WithAdditionalAnnotations(Simplifier.Annotation);
-            }
-
-            // Work on the document root instead of updating nodes with the editor directly so that
-            // additional changes can be made later (adding using statement) if necessary.
-            if (editor.OriginalRoot is not CompilationUnitSyntax documentRoot)
+            // Note that the simplifier does not recognize using statements within namespaces, so this
```
KSchlobohm, line 99: is this a known issue being tracked? shouldn't it do this?

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/IdentifierUpgradeCodeFixer.cs
```diff
@@ -60,78 +71,80 @@ public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
...
+            {
+                return generator.IdentifierName(name);
+            }
+
+            return node switch
+            {
+                // Many usages (constructing a type, casting to a type, etc.) will use a qualified name syntax
+                // to refer to the type.
+                var nameSyntax when
+                    nameSyntax is CSSyntax.NameSyntax
```
KSchlobohm, line 137: could leverage the language agnostic IsNameSyntax helper method suggested above

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/IdentifierUpgradeCodeFixer.cs
```diff
@@ -60,78 +71,80 @@ public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
...
+                    || nameSyntax is VBSyntax.NameSyntax => generator.QualifiedName(QualifiedNameBuilder.BuildQualifiedNameSyntax(generator, qualifier), generator.IdentifierName(name)),
```
KSchlobohm, line 138: Guess a more concise name would have helped here, oops.

Also, i think you could just use `QualifiedNameBuilder.BuildQualifiedNameSyntax(generator, newIdentifier)` here. The reason I used the split previously was simply to avoid string concatenation but you already have that string data.

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/AnalyzerTests.cs
```diff
@@ -150,10 +150,18 @@ public class AnalyzerTests
                 "UA0013",
                 new[]
                 {
-                    new ExpectedDiagnostic("UA0013", new TextSpan(143, 13)),
-                    new ExpectedDiagnostic("UA0013", new TextSpan(571, 29)),
-                    new ExpectedDiagnostic("UA0013", new TextSpan(153, 13), Language.VisualBasic),
-                    new ExpectedDiagnostic("UA0013", new TextSpan(439, 29), Language.VisualBasic),
+                    new ExpectedDiagnostic("UA0013", new TextSpan(166, 13)),
```
KSchlobohm, line 153: why did the indexes for nearly every analyzer move?

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/HtmlHelperAnalyzer.cs
```diff
@@ -91,5 +94,24 @@ private static string GetReplacementType(SyntaxNode node)
                 _ => AspNetCoreIHtmlHelperName // Default to IHtmlHelper
             };
         }
+
+        private static SyntaxNode GetQualifiedName(SyntaxNode name)
```
KSchlobohm, line 98: this is now available as an extension method isn't it?

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/assets/TestClasses/UA0008.Fixed.cs
```diff
@@ -7,7 +7,7 @@ public class UA0008 : IUrlHelper
     {
         public IUrlHelper Method1(this IUrlHelper h)
         {
-            var x = h;
+            IUrlHelper x = h;
```
KSchlobohm, line 10: this looks like a behavioral change. is this intentional?

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/assets/TestClasses/UA0013.Fixed.cs
```diff
@@ -1,10 +1,11 @@
 ﻿using System.Web.Http;
+using System.Web.Mvc;
 using System.Collections.Generic;
 using Microsoft.AspNetCore.Mvc;
 
 namespace TestProject.TestClasses
 {
-    public partial class ValuesController : Controller
+    public partial class ValuesController : ControllerBase
```
KSchlobohm, line 8: Why does the fixed version of ApiController now appear to reference System.Web.Mvc.ControllerBase?

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/assets/TestClasses/UA0013.Fixed.cs
```diff
@@ -35,6 +36,25 @@ public class NotAWebController : Foo.ApiController
     {
 
     }
+
+    public partial class Controller2
+    {
+        Controller Controller;
+    }
+
+    public partial class Controller2 : Controller
```
KSchlobohm, line 45: When this compiles, is this System.Web.Mvc.Controller or Microsoft.AspNetCore.Mvc.Controller?

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/assets/TestClasses/UA0013.Fixed.vb
```diff
@@ -1,11 +1,12 @@
 ﻿Imports System.Web.Http
+Imports System.Web.Mvc
 Imports System.Collections.Generic
 Imports Microsoft.AspNetCore.Mvc
 
 Namespace TestProject.TestClasses
 
     Public Class ValuesController2
-        Inherits Controller
+        Inherits ControllerBase
```
KSchlobohm, line 9: Same as above, I think I'm missing something

tests/tool/Integration.Tests/IntegrationScenarios/AspNetSample/csharp/Upgraded/Views/Home/Index.cshtml
```diff
@@ -15,7 +14,7 @@
     <div class="col-md-4">
         <h2>Getting started</h2>
         @{ 
-            var x = new Microsoft.AspNetCore.Mvc.ActionResult();
+            var x = new ActionResult();
```
KSchlobohm, line 17: seems like a behavioral change. Would expect a correlating change to `_ViewImports.cshtml` to make this work

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/SyntaxNodeExtensions.cs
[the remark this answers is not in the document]
KSchlobohm: can all of our identifierUpgradeCodeFixer changes be queued up?

KSchlobohm, line 229: does this have to be CSharp aware?

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/IdentifierUpgradeAnalyzer.cs
```diff
@@ -31,46 +48,69 @@ public override void Initialize(AnalysisContext context)
             context.EnableConcurrentExecution();
             context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
 
-            context.RegisterSyntaxNodeAction(AnalyzeIdentifiers, SyntaxKind.IdentifierName);
+            // Register actions for handling both C# and VB identifiers
+            context.RegisterSyntaxNodeAction(AnalyzeCSharpIdentifier, CS.SyntaxKind.IdentifierName);
```
KSchlobohm: The ApiContollerAnalyzer was registered for BaseListSyntax, which I think occurs less frequently than IdentifierName syntax.

Can we safely increase the scope without impacting performance for users that are referencing the analyzers through Visual Studio?

KSchlobohm (changes requested):
I didn't understand the reason for differences in test case expectations

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/IdentifierUpgradeAnalyzer.cs
KSchlobohm: Maybe the child class could declare an immutable array of things it would like to register

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/SyntaxNodeExtensions.cs
mjrousos: The problem with dropping the generic type is that it returns that type. We could drop it and make it language agnostic, but would have to return `SyntaxNode` then. I thought it might be better to just get the type of method declaration syntax that's expected (C# or VB) and only find/return nodes of that type.

And yes, it's one line. I'll split it up into more, though, to make it more debuggable.

mjrousos: And as far as expanding nodes, I'm just expanding an existing helper method to work with both C# and VB so I'm leaving it as unchanged functionally as possible.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/SyntaxNodeExtensions.cs
mjrousos: Ideally, this will go away and be replaced by `ImportAdder`. It's only here temporarily because it was previously used by some code fix providers that haven't been updated to be cross-language yet (HttpContext.Current and HttpContext.IsDebuggingEnabled). I'll put a comment noting that it should be removed and address it in a separate PR.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/HtmlHelperAnalyzer.cs
mjrousos: Yep, good catch. I must have added this before extracting that into an extension method. I'll remove it from here.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/UrlHelperAnalyzer.cs
mjrousos: Yep. Good catch. Fixed.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/HttpContextCurrentCodeFixer.cs
mjrousos: Hm... I suppose not. We do check that the method *exists*, but there's an edge case where it exists but is empty I guess. I'll update the check to make sure the method is both present and not empty before doing this.

mjrousos: I do care because I use `configureMethod` as a MethodDeclarationSyntax object which is C#-specific.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/IdentifierUpgradeCodeFixer.cs
mjrousos: Yea, I can see that. I'll add extension methods.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/IdentifierUpgradeCodeFixer.cs
mjrousos: I'm not sure. I'll look for an issue. It's possible it's by design because recognizing using statements within a namespace would mean that the simplifier would have to consider the scope of those namespaces which is trickier than just recognizing an import that's global for the whole file.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/IdentifierUpgradeCodeFixer.cs
mjrousos: Yep, updated.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/IdentifierUpgradeCodeFixer.cs
mjrousos: Good point. I'll skip the generated.QualifiedName call here.

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/assets/TestClasses/UA0013.Fixed.cs
mjrousos: I changed the code fixer for ApiController to use ControllerBase instead of Controller because I think that's the recommended base type for API controllers in ASP.NET Core.

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/assets/TestClasses/UA0008.Fixed.cs
mjrousos: Yes, it's a result of me not applying the reducer now which I think is good because it keeps the var/not-var details the same as before the code fix was applied.

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/AnalyzerTests.cs
mjrousos: Previously the diagnostic's location was just the simple name of the identifier that needs updated. This PR changes the analyzer to instead mark the entire name as the location so that the code fix can work on the location it gets directly instead of having to find the fully qualified name at that point. Most of these changes are updating the expected diagnostic locations to include the namespaces for qualified type names.

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/assets/TestClasses/UA0013.Fixed.cs
mjrousos: Likely it won't compile, unfortunately, until the `using System.Web.Mvc` is removed. We could have the code fix do that, but it's possible that other code in the class is using that namespace so it's hard to make this change without breaking something. We have other code fixes to remove those namespaces, so I think this is ok.

Now that you point this out, though, I'm kind of surprised that Roslyn's simplifier removed the namespace qualification here. My hunch is that it only did that because the test doesn't have a System.Web reference and in the real world, it would have been left as `Microsoft.AspNetCore.Mvc.Controller` (unless, of course, UA had previously removed the System.Web.Mvc namespace).

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/assets/TestClasses/UA0013.Fixed.vb
mjrousos: Yea, I made that change intentionally because I think it's the more correct translation to Core. Let me know if you disagree; it would be very easy to switch back.

tests/tool/Integration.Tests/IntegrationScenarios/AspNetSample/csharp/Upgraded/Views/Home/Index.cshtml
mjrousos: This is expected because of Razor magic I'm afraid :( When ASP.NET Core builds its Razor pages, it automatically includes a few namespaces. One of them is Microsoft.AspNetCore.Mvc.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.Analyzers/IdentifierUpgradeAnalyzer.cs
mjrousos: Should be ok for now because the PR I'm opening as soon as this one merges will combine all the IdentifierUpgrade* analyzers into one so we only have to analyze identifiers once.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/HttpContextCurrentCodeFixer.cs
KSchlobohm: missed that somehow

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/assets/TestClasses/UA0013.Fixed.cs
KSchlobohm: I thought the `UsingSystemWebAnalyzer` UA0001 would remove the System.Web.Mvc namespace. We're testing UA0013 but don't all of the analyzers run in the way this is configured?

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CSharp.CodeFixes/IdentifierUpgradeCodeFixer.cs
mjrousos: Makes sense. I'll see if I can find something to link to.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/SyntaxNodeExtensions.cs
```diff
@@ -237,11 +237,26 @@ public static CSSyntax.CompilationUnitSyntax AddUsingIfMissing(this CSSyntax.Com
...
+
+        /// <summary>
+        /// Determines whether a node if a NameSyntax (either C# or VB).
+        /// </summary>
+        /// <param name="node">The node to inspect.</param>
+        /// <returns>True if the node derives from Microsoft.CodeAnalysis.CSharp.Syntax.NameSyntax or Microsoft.CodeAnalysis.VisualBasic.Syntax.NameSyntax, false otherwise.</returns>
+        public static bool IsNameSyntax(this SyntaxNode node) => node is CSSyntax.NameSyntax || node is VBSyntax.NameSyntax;
+
+        /// <summary>
+        /// Determines whether a node if a MemberAccessExpressionSyntax (either C# or VB).
```
KSchlobohm, line 256: nit: Determines whether a node **_is_** a MemberAccessExpressoin

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/assets/TestClasses/UA0013.Fixed.cs
mjrousos: UA0001 will remove the using statement, but in our unit tests we only apply the code fix being tested.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/SyntaxNodeExtensions.cs
mjrousos: Good catch. Fixed.

KSchlobohm (approved):
Looks good, thanks for discussing!

mjrousos:
Thanks for the review!

PR #535, opened 2021-05-19
Ensure analyzer packaging works for CS/VB

twsouthwick:
This change removes CSharp from the name of the package as well as ensures the analzyer/codefixers are added for vb projects when added as NuGet packages.


sunandabalu (approved):
LGTM

KSchlobohm (approved):
looks good

PR #536, opened 2021-05-19
Add a codefixer to refactor HttpContext.Current

twsouthwick:
This adds an additional codefixer for HttpContext.Current that helps to
refactor its usage to instead use method injection. The codefixer will
update calling methods to pass in HttpContext.Current, which can be
called multiple times to move HttpContext.Current up the stack.


twsouthwick:
The VB tests are disabled until @mjrousos PR is merged in (#534)

twsouthwick:
Nm - #534 wasn't enabling VB for that analyzer. I've added an operation level analyzer for now for VB so that it'll unblock usage.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CodeFixes/HttpContextRefactor/HttpContextRefactorCodeFixProvider.cs
```diff
@@ -0,0 +1,297 @@
...
+            return slnEditor.GetChangedSolution();
+        }
+
+        private IOperation? GetEnclosingMethodOperation(IOperation? operation)
+            => operation.GetParentOperation(IsEnclosedMethodOperation);
+
+        private async Task<SyntaxNode?> GetOrAddMethodParameterAsync(DocumentEditor editor, Document document, IOperation methodOperation, IPropertyReferenceOperation propertyOperation, CancellationToken token)
+        {
+            // Get the symbol representing the type to be renamed.
+            var semanticModel = await document.GetSemanticModelAsync(token).ConfigureAwait(false);
```
KSchlobohm, line 109: is this the same semantic model used above?

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/HttpContextRefactor/VisualBasicHttpContextRefactorCodeFixProviderTests.cs
```diff
@@ -0,0 +1,442 @@
...
+                .WithFixed(fix1)
+                .WithFixed(fix2)
+                .WithExpectedDiagnosticsAfter(expected)
+                .RunAsync();
+        }
+
+        [Fact]
+        public async Task MultipleFilesNoSystemWebImport()
+        {
+            var test1 = @"
```
KSchlobohm, line 387: seems to be a theory

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/HttpContextRefactor/CSharpHttpContextRefactorCodeFixProviderTests.cs
```diff
@@ -0,0 +1,463 @@
...
+        }
+    }";
+            var fixtest = @"
+    using System.Web;
+
+    namespace ConsoleApplication1
+    {
+        class Program
+        {
+            public void Test(HttpContext currentContext)
```
KSchlobohm, line 114: currentContext seems like a relatively common name.

is it easy to enumerate the params already on the method and determine if we should use currentContext+{instanceCount}?

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/Verifiers/IAnalyzerTest.cs
```diff
@@ -0,0 +1,42 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Options;
+using Microsoft.CodeAnalysis.Testing;
+
+#pragma warning disable CA1002
```
KSchlobohm, line 12: you spoiled me with comments in other files that explained what the diagnostic was

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/Verifiers/VisualBasicCodeFixVerifier.cs
```diff
@@ -0,0 +1,28 @@
...
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Testing;
+using Microsoft.CodeAnalysis.Testing.Verifiers;
+using Microsoft.CodeAnalysis.VisualBasic.Testing;
+
+#pragma warning disable CA1000 // Do not declare static members on generic types
+#pragma warning disable CA1034 // Nested types should not be visible
+#pragma warning disable CA1724 // The type name Test conflicts
+
+namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test
```
KSchlobohm, line 14: seems to be a lot of test plumbing code in this PR. is there enough to make a test.common library?

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CodeFixes/HttpContextRefactor/HttpContextRefactorCodeFixProvider.cs
```diff
@@ -0,0 +1,297 @@
...
+        where TArgument : SyntaxNode
+    {
+        private const string DefaultArgumentName = "currentContext";
+
+        public sealed override ImmutableArray<string> FixableDiagnosticIds
+        {
+            get { return ImmutableArray.Create(HttpContextCurrentAnalyzer.DiagnosticId); }
+        }
+
+        public sealed override FixAllProvider GetFixAllProvider()
```
KSchlobohm, line 27: should we just make a new base class we reuse? I think we're putting this in every codefixer

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CodeFixes/HttpContextRefactor/HttpContextRefactorCodeFixProvider.cs
```diff
@@ -0,0 +1,297 @@
...
+
+            // Find the type declaration identified by the diagnostic.
+            var node = root.FindNode(diagnostic.Location.SourceSpan, getInnermostNodeForTie: true);
+
+            if (semantic.GetOperation(node, context.CancellationToken) is not IPropertyReferenceOperation property)
+            {
+                return;
+            }
+
+            var methodOperation = GetEnclosingMethodOperation(property);
```
KSchlobohm, line 62: we're in the code fixer. is this check still necessary? seems like something the analyzer should handle. if there is no enclosingMethodOperation then we wouldn't raise the diagnostic

KSchlobohm:
i think we would want to update the CHANGELOG.md to describe the different behavior for HttpContext.Current

twsouthwick:
@CyrusNajmabadi I'd love your feedback on this codefixer. The goal of it is to move HttpContext.Current usage up the levels of a callstack via method injection (potentially more later), so it does some heavy lifting in the codefixer and not sure if there's better patterns I could explore there. Thanks!

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CodeFixes/HttpContextRefactor/HttpContextRefactorCodeFixProvider.cs
twsouthwick: Good catch. Passing it through...

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/Verifiers/VisualBasicCodeFixVerifier.cs
twsouthwick: Maybe - I'm planning on seeing if we can just integrate some of it into the Microsoft.CodeAnalysis.Testing library itself

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CodeFixes/HttpContextRefactor/HttpContextRefactorCodeFixProvider.cs
twsouthwick: It's used to identify where to add parameters

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CodeFixes/HttpContextRefactor/HttpContextRefactorCodeFixProvider.cs
```diff
@@ -0,0 +1,297 @@
...
+                {
+                    continue;
+                }
+
+                if (!slnEditor.OriginalSolution.TryGetDocument(location.SourceTree, out var document))
+                {
+                    continue;
+                }
+
+                var editor = await slnEditor.GetDocumentEditorAsync(document.Id, token).ConfigureAwait(false);
```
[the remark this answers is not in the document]
twsouthwick: Not sure - I assumed they'd be cached. @CyrusNajmabadi is this ok to do?

CyrusNajmabadi:
In vacation. Will try to take a look when I can though :-)

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CodeFixes/HttpContextRefactor/CSharpHttpContextRefactorCodeFixProvider.cs
```diff
@@ -0,0 +1,19 @@
...
+using Microsoft.CodeAnalysis.Operations;
+
+namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CodeFixes
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp, Name = nameof(CSharpHttpContextRefactorCodeFixProvider))]
+    [Shared]
+    public class CSharpHttpContextRefactorCodeFixProvider : HttpContextRefactorCodeFixProvider<InvocationExpressionSyntax, ArgumentSyntax>
+    {
+        private protected override bool IsEnclosedMethodOperation(IOperation operation)
+            => operation is IMethodBodyOperation;
```
mjrousos, line 14: Can we just check whether the operation is `CS.IMethodBodyOperation` or `VB.IBlockOperation` in the base class so that we only have one code fix provider? I don't mind have a base class with two language-specific children if we need to, but these differences are so small it seems like it may be better to just have separate code paths in these two places in the base class.

mjrousos (approved):
This code fix looks for an existing property/parameter or adds a parameter with the same type as the original property had. So we'll be updating methods to accept System.Web.HttpContext parameters. Eventually, we don't want people running UA to depend on System.Web.HttpContext, though. I can think of two solutions to this:

1. Use System.Web.HttpContext but make sure this code fix runs earlier in the upgrade process (before TFM change, though we don't currently have a good way to do that). Then, after TFM change, have a different code fix that replaces all System.Web.HttpContext usage with Microsoft.AspNetCore.HttpContext.
2. Update this code fix to replace HttpContext.Current with properties/parameters of type Microsoft.AspNetCore.HttpContext. This would fit better with the current architecture of the tool since it would make sense to run at the same time as the other code fixes, but it has the drawback of not having System.Web symbols available (though we would have Microsoft.AspNetCore.* symbols).

Probably the first option is more useful when this is used as a standalone analyzer/code fix, but the second one works better in the context of UA. In fact, I'm not sure this would work in UA anyhow because applying it is likely to add more diagnostics than it fixes the first few times the code fix is applied and that might trip up the code fix upgrade step (which expects the number of diagnostics to decrease as code fixes are applied). 

Maybe the best option is two have two separate code fixes - this one which is used manually or in a yet-to-be-implemented pre-TFM-change UA step (though we'd have to be careful because we try harder not to break things at that point in UA's execution) and a separate but similar code fix that translates to Microsoft.AspNetCore.HttpContext after the TFM change.

KSchlobohm:
Regarding @mjrousos 's comment. If this doesn't change types to Microsoft.AspNetCore.HttpContext then isn't this a feature regression?

KSchlobohm:
We should be able to fix the type by running the new TypeUpgradeAnalyzer after this one runs. More than a few times now I've started wondering if we need to support sort order by declaring dependencies for Analyzers like we do for steps.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/HttpContextCurrentAnalyzer.cs
```diff
@@ -40,7 +41,33 @@ public override void Initialize(AnalysisContext context)
...
+                            return;
+                        }
+
+                        if (NameMatcher.HttpContextCurrent.Matches(propertyReference.Property))
+                        {
+                            var diagnostic = Diagnostic.Create(Rule, ctx.Operation.Syntax.GetLocation());
+
+                            ctx.ReportDiagnostic(diagnostic);
+                        }
+                    }, OperationKind.PropertyReference);
```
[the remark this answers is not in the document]
twsouthwick: I was punting that till later so I could just get it in, but I went back and made it work with syntax level and tried to reuse what I could of the current one.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/NameMatcher.cs
```diff
@@ -0,0 +1,68 @@
+﻿using System;
+using Microsoft.CodeAnalysis;
+
+namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Default
+{
+    public class NameMatcher
+    {
+        public static NameMatcher HttpContextCurrent { get; } = new NameMatcher("System.Web.HttpContext", "Current");
```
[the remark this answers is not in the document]
twsouthwick: Exposed static helpers to construct the matcher based on what we want to match

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/NameMatcher.cs
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        public bool Matches(ITypeSymbol? typeSymbol)
+        {
+            if (typeSymbol is null)
+            {
+                return false;
+            }
+
+            var symbol = (INamespaceOrTypeSymbol)typeSymbol;
```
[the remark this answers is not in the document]
twsouthwick: Mostly because we only want to take in a type, but a types parent could be a type or namespace. Added a comment to that effect.

twsouthwick:
The goal here is solely to replace usage of HttpContext.Current with method injection which is something that can be done before upgrading. Replacing HttpContext with an abstraction would be the next step, but I don't think we'd want to do that in this code fixer.  This enables the following scenario:

1. Find where HttpContext.Current is being used
2. Inject HttpContext rather than access via static method
3. Update calling patterns to pass through from controller/filter/etc
4. [TODO] Replace HttpContext with custom abstraction
5. Upgrade to ASP.NET Core
6. No HttpContext[.Current] usage in project so no need to deal with it now

The benefit of this is:

1. Gets the code base into a better place without static accessors
2. Allows for the code to be updated without breaking anything

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CodeFixes/HttpContextRefactor/CSharpHttpContextRefactorCodeFixProvider.cs
twsouthwick: Done. Using extension methods

mjrousos (approved):
This looks good to me. I think it might be hard to incorporate into UA (if we want to do that) since it is most useful if run before the TFM change and because the total number of diagnostics won't initially go down when the code fix is applied (since the user is expected to apply it iteratively, several times), but the analyzer and code fix provider look useful for manual use alongside UA.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/SymbolExtensions.cs
```diff
@@ -2,12 +2,68 @@
...
     public static class SymbolExtensions
     {
+        public static bool NameEquals(this ISymbol? symbol, string name)
+            => string.Equals(symbol?.Name, name, symbol.GetStringComparison());
+
+        public static StringComparison GetStringComparison(this ISymbol? symbol)
+            => symbol?.Language switch
+            {
+                LanguageNames.CSharp => StringComparison.Ordinal,
+                LanguageNames.VisualBasic => StringComparison.OrdinalIgnoreCase,
```
KSchlobohm, line 22: 👍

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/SymbolExtensions.cs
```diff
@@ -2,12 +2,68 @@
...
+
+        public static StringComparer GetStringComparer(this ISymbol? symbol)
+            => symbol?.Language switch
+            {
+                LanguageNames.CSharp => StringComparer.Ordinal,
+                LanguageNames.VisualBasic => StringComparer.OrdinalIgnoreCase,
+                _ => throw new NotImplementedException(Resources.UnknownLanguage),
+            };
+
+        public static IEnumerable<ISymbol> GetAllMembers(this INamedTypeSymbol? symbol, bool includeInstance)
```
KSchlobohm, line 34: Syntactically speaking, what is a member?  Let's update with a comment block.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/SymbolExtensions.cs
```diff
@@ -24,5 +80,64 @@ public static bool NameEquals(this IAssemblySymbol? symbol, string name, bool st
                 return string.Equals(symbol.Name, name, StringComparison.OrdinalIgnoreCase);
```
KSchlobohm, line 80: This looks like it should use one of the helper methods identified above
`GetStringComparison`

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/SymbolExtensions.cs
```diff
@@ -24,5 +80,64 @@ public static bool NameEquals(this IAssemblySymbol? symbol, string name, bool st
...
+
+        /// <summary>
+        /// Gets an existing parameter of the type from node if it is a method.
+        /// </summary>
+        /// <param name="node">A node that is potentially a method.</param>
+        /// <param name="semanticModel">The current semantic model.</param>
+        /// <param name="type">The type that should be matched.</param>
+        /// <param name="token">A cancellation token.</param>
+        /// <returns>A parameter symbol if a match is found.</returns>
+        public static IParameterSymbol? GetExistingParameterSymbol(this SyntaxNode node, SemanticModel semanticModel, ITypeSymbol? type, CancellationToken token)
```
KSchlobohm, line 92: nit: prefer name GetExistingParameter**By**Symbol

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/SymbolExtensions.cs
```diff
@@ -24,5 +80,64 @@ public static bool NameEquals(this IAssemblySymbol? symbol, string name, bool st
                 return string.Equals(symbol.Name, name, StringComparison.OrdinalIgnoreCase);
             }
         }
+
+        /// <summary>
+        /// Gets an existing parameter of the type from node if it is a method.
```
KSchlobohm, line 85: nit: get's the **first** instance of an existing parameter

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/SymbolExtensions.cs
```diff
@@ -24,5 +80,64 @@ public static bool NameEquals(this IAssemblySymbol? symbol, string name, bool st
...
+            });
+        }
+
+        /// <summary>
+        /// Gets a parent operation that satisfies the given predicate.
+        /// </summary>
+        /// <param name="operation">A starting operation.</param>
+        /// <param name="func">A predicate to check matches against.</param>
+        /// <returns>A matching operation if found.</returns>
+        public static IOperation? GetParentOperation(this IOperation? operation, Func<IOperation, bool> func)
```
KSchlobohm, line 123: nit: GetParentOperation**Where**

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/SyntaxNodeExtensions.cs
```diff
@@ -4,15 +4,38 @@
...
 
 namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Default
 {
     public static class SyntaxNodeExtensions
     {
+        public static bool IsVisualBasic(this SyntaxNode node) => node?.Language == LanguageNames.VisualBasic;
+
+        public static bool IsCSharp(this SyntaxNode node) => node?.Language == LanguageNames.CSharp;
+
+        public static SyntaxNode AddArgumentToInvocation(this SyntaxNode invocationNode, SyntaxNode argument)
```
KSchlobohm, line 23: this looks nice, let's merge this PR first so I can reuse this in the PR for binaryFormatter

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/WorkspaceExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            if (tree is null || sln is null)
+            {
+                document = null;
+                return false;
+            }
+
+            foreach (var project in sln.Projects)
+            {
+                var doc = project.GetDocument(tree);
```
KSchlobohm, line 28: If I copy and paste the same class file from projectA into projectB then this will find the first instance right?

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/HttpContextCurrentAnalyzer.cs
```diff
@@ -5,14 +5,18 @@
...
+using CSSyntax = Microsoft.CodeAnalysis.CSharp.Syntax;
+
+using VB = Microsoft.CodeAnalysis.VisualBasic;
+using VBSyntax = Microsoft.CodeAnalysis.VisualBasic.Syntax;
+
 namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers
 {
     [ApplicableComponents(ProjectComponents.AspNetCore)]
-    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
```
KSchlobohm, line 19: 👍

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/HttpContextCurrentAnalyzer.cs
```diff
@@ -40,12 +44,50 @@ public override void Initialize(AnalysisContext context)
             context.EnableConcurrentExecution();
             context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
 
-            context.RegisterSyntaxNodeAction(AnalyzeMemberAccessExpressions, SyntaxKind.SimpleMemberAccessExpression);
+            context.RegisterCompilationStartAction(context =>
+            {
+                if (context.Compilation.Language == LanguageNames.CSharp)
```
KSchlobohm, line 49: What's the advantage here to registering compilation?

Should we retrofit other analyzers with this pattern?

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/HttpContextCurrentAnalyzer.cs
```diff
@@ -40,12 +44,50 @@ public override void Initialize(AnalysisContext context)
...
+
+            // If the accessed member isn't named "Current" bail out
+            if (!TargetMember.Equals(memberAccessExpression.Name.ToString(), StringComparison.Ordinal))
+            {
+                return;
+            }
+
+            // If the call is to a method called Current then bail out since they're
+            // not using the HttpContext.Current property
+            if (memberAccessExpression.Parent is CSSyntax.InvocationExpressionSyntax)
```
KSchlobohm, line 72: nit: try to put string comparisons later in the analyzer "bail out" flow. I don't have a doc recommending this but I'm basing that on string comparison always being slower than other conditionals we can run.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CodeFixes/HttpContextRefactorCodeFixProvider.cs
```diff
@@ -0,0 +1,79 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.Operations;
+using Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers;
+
+namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CodeFixes
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp, LanguageNames.VisualBasic, Name = nameof(HttpContextRefactorCodeFixProvider))]
+    [Shared]
```
KSchlobohm, line 17: 👍

KSchlobohm (approved):
added some nits, otherwise looks good

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/SymbolExtensions.cs
twsouthwick: actually, this is an assembly name comparison which is case insensitive AFAIK (maybe depends on the file system, though)

twsouthwick: Added a comment, though

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/WorkspaceExtensions.cs
twsouthwick: No - it's still a separate document

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/HttpContextCurrentAnalyzer.cs
twsouthwick: Good question - I'm not sure which is better. This will make it so only the one we care about is actually registered. Not sure how much idfference that has

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/HttpContextCurrentAnalyzer.cs
twsouthwick: The best comparison we have here is if Current is not the name, then we shouldn't continue at all. It's the best filter I can think of

PR #540, opened 2021-05-20
Consolidate type upgrade analyzers into a single configurable analyzer

mjrousos:
This consolidates all the analyzers derived from `IdentifierUpgradeAnalyzer` and replaces them with a single analyzer/code fix provider for identifying and updating all cases where a type name needs replaced with a more modern equivalent.

It also updates this new `TypeUpgradeAnalyzer` to read the types to be replaced dynamically from additional texts so that users can easily extend the mapping of types to be replaced (either by including a .typemap file in their project or by registering an AdditionalText file with that extension in UA's DI container).


src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/TypeMapping.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers
+{
+    public record TypeMapping(string OldName, string NewName)
+    {
+        public string SimpleName => OldName.LastIndexOf('.') < 0
```
KSchlobohm, line 8: do we need a null check here? I keep forgetting where we're using nullable-reference types.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/TypeUpgradeAnalyzer.cs
```diff
@@ -0,0 +1,174 @@
...
+        /// <param name="location">The location the diagnostic occurs at.</param>
+        /// <param name="properties">Properties (including the name of the new identifier that code fix providers should substitute in) that should be included in the diagnostic.</param>
+        /// <param name="messageArgs">Arguments (the simple name of the identifier to be replaced and the full name of the identifier to replace it) to be used in diagnotic messages.</param>
+        /// <returns>A diagnostic to be shown to the user.</returns>
+        private static Diagnostic CreateDiagnostic(Location location, ImmutableDictionary<string, string?> properties, params object[] messageArgs)
+            => Diagnostic.Create(Rule, location, properties, messageArgs);
+
+        private void AnalyzeCSharpIdentifier(SyntaxNodeAnalysisContext context)
+        {
+            var identifier = (CSSyntax.IdentifierNameSyntax)context.Node;
```
KSchlobohm, line 102: if we had an abstraction of IdentifierNameSyntax we could use a helper function to convert this to 1 standard type.

Obviously not a big help here but I think this is starting to look like a recurring problem. MemberAccessExpression was the other place I remember seeing this a lot and I think we actually have some helpers that use generics that are trying to solve the same problem.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/TypeMapping.cs
mjrousos: Good point. Since this type is public, we should probably check for null. I'll fix it.

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/AnalyzerTests.cs
```diff
@@ -145,22 +106,76 @@ public class AnalyzerTests
...
+                    new ExpectedDiagnostic("UA0002", new TextSpan(513, 25)),
+                    new ExpectedDiagnostic("UA0002", new TextSpan(563, 14)),
+                    new ExpectedDiagnostic("UA0002", new TextSpan(612, 18))
+                }
+            },
+            {
+                "FilterUpgrade",
+                new[]
+                {
+                    new ExpectedDiagnostic("UA0002", new TextSpan(88, 28)),
```
KSchlobohm, line 152: always concerned when refactoring has to change test assertions

```
new ExpectedDiagnostic("UA0004", new TextSpan(82, 28))
new ExpectedDiagnostic("UA0004", new TextSpan(156, 36)),
new ExpectedDiagnostic("UA0004", new TextSpan(321, 37)),
```

KSchlobohm: reading further, maybe this was just a drift created by renaming the classes in these test files

KSchlobohm (approved):
LGTM

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/AnalyzerTests.cs
mjrousos: Yep, that's the culprit. You'll notice the offsets change very predictably by the different in class name length.

KSchlobohm: That looked good to me, glad you named things by scenario it clears up things like the previously named UA0004A

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/TypeUpgradeAnalyzer.cs
```diff
@@ -0,0 +1,174 @@
...
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
+
+            context.RegisterCompilationStartAction(context =>
+            {
+                // Load analyzer configuration defining the types that should be mapped.
+                _mappings = LoadMappings(context.Options.AdditionalFiles);
```
twsouthwick, line 78: You've got a race conditiion here. The mappings should be passed through the actions that are registered in case a mapping is changed at some point and then they'd be swapped from underneath you at the next compilation

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/WebTypeReplacements.cs
```diff
@@ -0,0 +1,45 @@
...
+            { "System.Web.Mvc.FileResult", "Microsoft.AspNetCore.Mvc.FileResult" },
+            { "System.Web.Mvc.HttpNotFoundResult", "Microsoft.AspNetCore.Mvc.NotFoundResult" },
+            { "System.Web.Mvc.HttpStatusCodeResult", "Microsoft.AspNetCore.Mvc.StatusCodeResult" },
+            { "System.Web.Mvc.HttpUnauthorizedResult", "Microsoft.AspNetCore.Mvc.UnauthorizedResult" },
+            { "System.Web.Mvc.RedirectResult", "Microsoft.AspNetCore.Mvc.RedirectResult" },
+            { "System.Web.Mvc.PartialViewResult", "Microsoft.AspNetCore.Mvc.PartialViewResult" },
+            { "System.Web.Mvc.ViewResult", "Microsoft.AspNetCore.Mvc.ViewResult" },
+        };
+
+        public override string Path => "WebTypeReplacements.typemap";
```
twsouthwick, line 40: Do we want to limit this to web types? What about any type?

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CodeFixes/CodeFixResources.resx
```diff
@@ -147,6 +147,9 @@
   <data name="ResultTypeTitle" xml:space="preserve">
     <value>Replace with ASP.NET Core result type</value>
   </data>
+  <data name="TypeUpgradeTitle" xml:space="preserve">
+    <value>Replace type with .NET Core type</value>
```
twsouthwick, line 151: Is it necessarily a .NET Core type? Or is it just a type replacement?

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorSourceUpdater.cs
```diff
@@ -56,10 +57,16 @@ public class RazorSourceUpdater : IUpdater<RazorCodeDocument>
         /// <param name="textMatcher">The text matching service to use for correlating old sections of text in Razor documents with updated texts.</param>
         /// <param name="textReplacer">The text replacing service to use for updating replaced texts in the Razor documents.</param>
         /// <param name="logger">An ILogger to log diagnostics.</param>
-        public RazorSourceUpdater(IEnumerable<DiagnosticAnalyzer> analyzers, IEnumerable<CodeFixProvider> codeFixProviders, ITextMatcher textMatcher, ITextReplacer textReplacer, ILogger<RazorSourceUpdater> logger)
+        public RazorSourceUpdater(IEnumerable<DiagnosticAnalyzer> analyzers, IEnumerable<CodeFixProvider> codeFixProviders, IEnumerable<AdditionalText> additionalTexts, ITextMatcher textMatcher, ITextReplacer textReplacer, ILogger<RazorSourceUpdater> logger)
```
twsouthwick, line 60: Why not pass an ImmutableArray<AdditionalText> here since that's what you actually want

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source/SourceUpdaterStep.cs
```diff
@@ -50,16 +51,22 @@ public class SourceUpdaterStep : UpgradeStep
             WellKnownStepIds.NextProjectStepId,
         };
 
-        public SourceUpdaterStep(IEnumerable<DiagnosticAnalyzer> analyzers, IEnumerable<CodeFixProvider> codeFixProviders, ILogger<SourceUpdaterStep> logger)
+        public SourceUpdaterStep(IEnumerable<DiagnosticAnalyzer> analyzers, IEnumerable<CodeFixProvider> codeFixProviders, IEnumerable<AdditionalText> additionalTexts, ILogger<SourceUpdaterStep> logger)
```
twsouthwick, line 54: Again, why not pass an ImmutableArray<AdditionalText>?

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/AnalyzerTests.cs
```diff
@@ -178,62 +193,62 @@ public async Task NegativeTest()
...
+                var diagnostics = await TestHelper.GetDiagnosticsAsync(language, $"{scenarioName}.{fileExtension}", expectedDiagnostics.Select(e => e.Id).Distinct()).ConfigureAwait(false);
                 AssertDiagnosticsCorrect(diagnostics, expectedDiagnostics);
             }
         }
 
         [InlineData("UA0001")]
-        [InlineData("UA0002")]
-        [InlineData("UA0003")]
-        [InlineData("UA0004")]
+        [InlineData("HtmlStringUpgrade")]
```
twsouthwick, line 226: Do we need tests for each of these? Going forward, are you expecting a full code analysis test for each upgrade pair? What about verifying the list contains something specific?

twsouthwick (changes requested):

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/WebTypeReplacements.cs
mjrousos: The analyzer/code fix provider should work for any type, but this config file is specifically for web types just to keep things organized. I imagine if we have other types we want to replace, we would have a separate typemap for those. The same analyzer/code fix would be used (it can pick up any number of typemaps) but keeping the mapping files organized by feature area seemed good.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CodeFixes/CodeFixResources.resx
mjrousos: No, I suppose not. I'll update this to make it more general.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorSourceUpdater.cs
mjrousos: I'm using ctor injection here. I wasn't sure whether Microsoft.Extensions.DependencyInjection would build an ImmutableArray for me. I suppose that's easy enough to test, though. I'll try it.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source/SourceUpdaterStep.cs
mjrousos: Same reason (ctor injection typically uses IEnumerable). I'll check to see if ImmutableArray works here and switch the code if it does.

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/AnalyzerTests.cs
mjrousos: I think we may as well leave the tests since we already have them. I like the idea of having a test to make sure that expected pairs are present, though. I'll add that and, in the future, we can just update the unit test when new pairs are added.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorSourceUpdater.cs
mjrousos: Unfortunately, it looks like we'll need to stick with IEnumerable because Autofact doesn't know how to supply ImmutableArray:

```
Autofac.Core.DependencyResolutionException: An exception was thrown while activating Microsoft.DotNet.UpgradeAssistant.Cli.ConsoleUpgrade -> Microsoft.DotNet.UpgradeAssistant.UpgraderManager -> Microsoft.DotNet.UpgradeAssistant.UpgradeStepOrderer -> ?:Microsoft.DotNet.UpgradeAssistant.UpgradeStep[] -> Microsoft.DotNet.UpgradeAssistant.Steps.Razor.RazorUpdaterStep -> ?:Microsoft.DotNet.UpgradeAssistant.IUpdater`1[[Microsoft.AspNetCore.Razor.Language.RazorCodeDocument, Microsoft.AspNetCore.Razor.Language, Version=5.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60]][] -> Microsoft.DotNet.UpgradeAssistant.Steps.Razor.RazorSourceUpdater.
 ---> Autofac.Core.DependencyResolutionException: None of the constructors found with 'Autofac.Core.Activators.Reflection.DefaultConstructorFinder' on type 'Microsoft.DotNet.UpgradeAssistant.Steps.Razor.RazorSourceUpdater' can be invoked with the available services and parameters:
Cannot resolve parameter 'System.Collections.Immutable.ImmutableArray`1[Microsoft.CodeAnalysis.AdditionalText] additionalTexts' of constructor 'Void .ctor(System.Collections.Generic.IEnumerable`1[Microsoft.CodeAnalysis.Diagnostics.DiagnosticAnalyzer], System.Collections.Generic.IEnumerable`1[Microsoft.CodeAnalysis.CodeFixes.CodeFixProvider], System.Collections.Immutable.ImmutableArray`1[Microsoft.CodeAnalysis.AdditionalText], Microsoft.DotNet.UpgradeAssistant.Steps.Razor.ITextMatcher, Microsoft.DotNet.UpgradeAssistant.Steps.Razor.ITextReplacer, Microsoft.Extensions.Logging.ILogger`1[Microsoft.DotNet.UpgradeAssistant.Steps.Razor.RazorSourceUpdater])'.
   at Autofac.Core.Activators.Reflection.ReflectionActivator.GetAllBindings(ConstructorBinder[] availableConstructors, IComponentContext context, IEnumerable`1 parameters)
```

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/TypeUpgradeAnalyzer.cs
mjrousos: Good point. I don't expect the mappings to change, but it makes sense to pass them along since we're checking here that they're non-empty. I'll make that change.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/TypeMapping.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers
+{
+    public record TypeMapping(string OldName, string NewName)
+    {
+        public string? SimpleName => (OldName?.LastIndexOf('.') ?? -1) < 0
```
twsouthwick, line 8: Should this be set in the constructor?

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/TypeUpgradeAnalyzer.cs
```diff
@@ -0,0 +1,169 @@
...
+            // that the code fix provider can directly replace the node without needing to consider its parents.
+            var diagnostic = CreateDiagnostic(fullyQualifiedNameNode.GetLocation(), properties, simpleName, mapping.NewName);
+            context.ReportDiagnostic(diagnostic);
+        }
+
+        /// <summary>
+        /// Load type mappings from additional files.
+        /// </summary>
+        /// <returns>Type mappings as defined in *.typemap files in the project's additional files.</returns>
+        private static IEnumerable<TypeMapping> LoadMappings(ImmutableArray<AdditionalText> additionalTexts)
```
twsouthwick, line 151: If this were a separate class (or at least an extension method), you could easily construct some unit tests for various combinations

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/TypeUpgradeAnalyzer.cs
```diff
@@ -0,0 +1,169 @@
...
+        {
+            foreach (var file in additionalTexts)
+            {
+                if (file.Path.EndsWith(TypeMapFileSuffix, StringComparison.OrdinalIgnoreCase))
+                {
+                    foreach (var line in file.GetText()?.Lines ?? Enumerable.Empty<TextLine>())
+                    {
+                        var components = line.ToString().Trim().Split('\t');
```
twsouthwick, line 159: What about checking if it starts with `#` so people can do comments?

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/WebTypeReplacements.cs
```diff
@@ -0,0 +1,45 @@
...
+
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Text;
+
+namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers
+{
+    public class WebTypeReplacements : AdditionalText
```
twsouthwick, line 12: This seems kind of weird. Why not add it as as an actual additional document? If there's a good reason to do it this way, maybe add a comment as to the rationale.

We should have a props file that gets packaged with the analyzers that set things as additional files and then the project will automatically get them. We may need to update UA to ensure that's loaded correctly if we're not adding an analyzer to the actualy csproj.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorSourceUpdater.cs
twsouthwick: Why are you getting it via dependency injection? Is the main one the WebTypeReplacements? Why not use the extension model to allow additional files that get added in?

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/WebTypeReplacementsTests.cs
```diff
@@ -0,0 +1,60 @@
...
+            { "System.Web.Mvc.RedirectResult", "Microsoft.AspNetCore.Mvc.RedirectResult" },
+            { "System.Web.Mvc.PartialViewResult", "Microsoft.AspNetCore.Mvc.PartialViewResult" },
+            { "System.Web.Mvc.ViewResult", "Microsoft.AspNetCore.Mvc.ViewResult" },
+        };
+
+        [Fact]
+        public static void WebTypeReplacementContentTests()
+        {
+            // Arrange
+            var replacementsText = new WebTypeReplacements();
```
twsouthwick, line 43: I'm not sure we want to map them with a custom class like this. It's pretty hard-coded

twsouthwick (changes requested):

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/TypeUpgradeAnalyzer.cs
mjrousos: Good idea. I'll add that.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/TypeUpgradeAnalyzer.cs
mjrousos: Good point. I'll make it its own class.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/WebTypeReplacements.cs
mjrousos: Yea, I'm doing this because UA loads the analyzers directly (rather than via a csproj reference), and the AdditionalTexts are added programatically. 

I'll see about changing it to use a file on-disk. I'm not sure what the best way is to make sure the typemap file is packaged with the analyzers and then made available to the project file, but I'm sure there's a way.

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/WebTypeReplacementsTests.cs
KSchlobohm: loading data from the class is going to be much faster than loading the data from disk. These examples are built to run in VS so that optimization seemed like part of the design decision.

Agree that extending this would be more flexible if we could read from disk, but not sure we'd want that feature running in VS.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/WebTypeReplacements.cs
mjrousos: Ok, I have an update that loads type maps from on-disk type map configuration. It's a few more hoops to jump through for UA, but I think it's better overall because the same mapping files will work for UA and for the analyzers used as a standalone NuGet package.

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/WebTypeReplacementsTests.cs
mjrousos: I've re-worked this to load type mappings from configuration files so that they work better in the standalone analyzer package case. It probably is less performant, but I don't think it will matter. UA consumers can still supply AdditionalText files however they like (either by deriving from the class or using the `AdditionalFileText` I've added to load from disk).

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/TypeUpgradeAnalyzer.cs
KSchlobohm: I like the swap to pass-thru for _mappings but would like to ask more about the race condition.

@twsouthwick - are you indicating this analyzer is a singleton? Is there a scenario where an analyzer should be stateful or is that an edge case?

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Package/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Package.csproj
```diff
@@ -27,13 +27,18 @@
...
     <ItemGroup>
+      <TfmSpecificPackageFile Include="$(OutputPath)\Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common.dll" PackagePath="analyzers/dotnet/cs" />
       <TfmSpecificPackageFile Include="$(OutputPath)\Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.dll" PackagePath="analyzers/dotnet/cs" />
       <TfmSpecificPackageFile Include="$(OutputPath)\Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CodeFixes.dll" PackagePath="analyzers/dotnet/cs" />
+      <TfmSpecificPackageFile Include="$(OutputPath)\Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common.dll" PackagePath="analyzers/dotnet/vb" />
       <TfmSpecificPackageFile Include="$(OutputPath)\Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.dll" PackagePath="analyzers/dotnet/vb" />
       <TfmSpecificPackageFile Include="$(OutputPath)\Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CodeFixes.dll" PackagePath="analyzers/dotnet/vb" />
+      <TfmSpecificPackageFile Include="$(OutputPath)\TypeMaps.props" PackagePath="build" />
+      <TfmSpecificPackageFile Include="$(OutputPath)\WebTypeReplacements.typemap" PackagePath="build" />
```
KSchlobohm, line 41: why not a resource file? is this to promote the ability for users to edit the file?

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Extensions.Default/DefaultExtensionServiceProvider.cs
```diff
@@ -17,6 +19,10 @@ namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Default
     public class DefaultExtensionServiceProvider : IExtensionServiceProvider
     {
         private const string TryConvertProjectConverterStepOptionsSection = "TryConvertProjectConverter";
+        private static readonly string[] TypeMapFiles = new[]
+        {
+            "WebTypeReplacements.typemap"
+        };
```
KSchlobohm, line 25: nit: use configuration option so i can say where the file comes from and how many there are. would that be a future enhancement?

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/AdditionalFileText.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// </summary>
+        public override string Path { get; }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AdditionalFileText"/> class.
+        /// </summary>
+        /// <param name="filePath">The file path to read the text from.</param>
+        public AdditionalFileText(string filePath)
+        {
+            if (string.IsNullOrWhiteSpace(filePath))
```
KSchlobohm, line 28: At this point it seems like you should also do a `File.Exists` check

twsouthwick (approved):
A couple of comments, but otherwise LGTM

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Extensions.Default/DefaultExtensionServiceProvider.cs
mjrousos: Yep, good idea. I'll do it as a configuration option

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Package/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Package.csproj
mjrousos: Yes, when a user uses our analyzer package, I want the mechanism we use to include additional files to be something that they can easily extend with their own additional files.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/TypeMapLoader.cs
```diff
@@ -36,7 +36,7 @@ public static IEnumerable<TypeMapping> LoadMappings(ImmutableArray<AdditionalTex
                             continue;
                         }
 
-                        var components = trimmedLine.Split('\t');
+                        var components = trimmedLine.Split(new char[] { '\t', ' ' }, StringSplitOptions.RemoveEmptyEntries);
```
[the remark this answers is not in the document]
mjrousos: I ended up making it support tabs and spaces both because I was concerned about the scenario where an editor turns tabs into spaces or something like that. Type names shouldn't include spaces, so splitting on both of these should work. I remove empty entries in the split output so `A\t\t\t   \t\t\tB` will still only be a two-item array.

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Extensions.Default/DefaultExtensionServiceProvider.cs
mjrousos: I've updated this to use extension configuration options, so users should be able to add their own config settings listing other additional files they want UA to use.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/AdditionalFileText.cs
mjrousos: Makes sense. Added.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/TypeMaps.props
```diff
@@ -0,0 +1,8 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<Project>
+  <ItemGroup>
+    <AdditionalFiles Include="$(MSBuildThisFileDirectory)\WebTypeReplacements.typemap">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
```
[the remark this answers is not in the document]
mjrousos: Good point; this isn't needed. I'll remove it.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/TypeMaps.props
```diff
@@ -0,0 +1,8 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<Project>
+  <ItemGroup>
+    <AdditionalFiles Include="$(MSBuildThisFileDirectory)\WebTypeReplacements.typemap">
```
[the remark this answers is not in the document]
mjrousos: What is the analyzer package is used with an old-style csproj, though? *.typemap might not work in that case.

Issue #533, opened 2021-05-18
Language agnostic BinaryFormatter analyzer


KSchlobohm:
## Summary

Modify the existing *BinaryFormatterUnsafeDeserializeAnalyzer* (and code fixer) to be language agnostic.

## Motivation and goals

Add another example of a language agnostic approach to updating code that can be referenced by other developers. This category of code updating demonstrates how to update code that called one method to use another.

## In scope

Support for the *BinaryFormatterUnsafeDeserializeAnalyzer* in Visual Basic and C#

## Out of scope

Abstracting the approach to make a configurable/reusable map any method analyzer.

PR #544, opened 2021-05-24
Language indifferent BinaryFormatter analyzer

KSchlobohm:
**PR Title**
Convert BinaryFormatterAnalyzer to be language indifferent

**PR Description**
* Updated code analyzer and code fixer
* Adds vb unit test scenario
* Adds new _General_ types as wrappers around shared csharp and vb syntax types to keep code language indifferent

Addresses #533


src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/BinaryFormatterUnsafeDeserializeAnalyzer.cs
```diff
@@ -39,71 +38,81 @@ public override void Initialize(AnalysisContext context)
             context.EnableConcurrentExecution();
             context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
 
-            context.RegisterSyntaxNodeAction(AnalyzeMemberAccessExpressions, SyntaxKind.SimpleMemberAccessExpression);
+            context.RegisterSyntaxNodeAction(AnalyzeMemberAccessExpressions, CodeAnalysis.CSharp.SyntaxKind.SimpleMemberAccessExpression);
```
twsouthwick, line 41: Why not do an operation based? Then you get all the language agnostic features of operating there. I do not see a need to operate in a syntax level for this - it can be (and should be) done before moving to .net core and both APIs are available in .NET Framework

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/BinaryFormatterUnsafeDeserializeAnalyzer.cs
```diff
@@ -39,71 +38,81 @@ public override void Initialize(AnalysisContext context)
...
+            context.RegisterSyntaxNodeAction(AnalyzeMemberAccessExpressions, CodeAnalysis.VisualBasic.SyntaxKind.SimpleMemberAccessExpression);
```
twsouthwick, line 42: If we're going to do this as a pattern, let's create an extension method that will register both of these and handle as much book keeping as necessary.

I had put in a `RegisterMemberAccessExpression` kind of extension at some point (it operates at the symbol layer), but that kind of pattern could be used here to simplify all (regardless of symbol/syntax level).

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/BinaryFormatterUnsafeDeserializeAnalyzer.cs
```diff
@@ -39,71 +38,81 @@ public override void Initialize(AnalysisContext context)
...
-                if (!symbolText.StartsWith(TargetTypeSymbolNamespace, StringComparison.Ordinal)
-                    || !symbolText.EndsWith(TargetTypeSymbolName, StringComparison.Ordinal))
-                {
-                    return;
-                }
+                return false;
             }
-            else if (accessedSymbol != null)
+
+            return theSymbol.ContainingType.ToString().Equals(QualifiedTargetSymbolName, StringComparison.Ordinal);
```
twsouthwick, line 96: Once my PR for the HttpContext refactor gets in, I've got a `NameMatcher` helper that will do this in a better way (ie checks each part rather than calling tostring)

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/BinaryFormatterUnsafeDeserializeAnalyzer.cs
```diff
@@ -39,71 +38,81 @@ public override void Initialize(AnalysisContext context)
...
+        }
+
+        /// <summary>
+        /// Checks to see if the symbol is a reference to BinaryFormatter.
+        /// </summary>
+        /// <param name="theSymbol">any symbol.</param>
+        /// <returns>True when (e.g. new BinaryFormatter().UnsafeDeserialize and <paramref name="theSymbol"/> is the symbol representing BinaryFormatter).</returns>
+        private static bool IsSymbolAConstructorInstanceOfBinaryFormatter(ISymbol theSymbol)
+        {
+            if (theSymbol is null)
```
twsouthwick, line 91: nit: shouldn't be necessary as you don't accept nulls by the constructor

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CodeFixes/BinaryFormatterUnsafeDeserializeCodeFixer.cs
```diff
@@ -64,54 +64,47 @@ public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
...
-
-        private static void ReplaceMethodCall(MemberAccessExpressionSyntax? memberExpression, DocumentEditor docEditor)
-        {
-            if (memberExpression is null)
-            {
-                return;
-            }
+            // the codefixer should only be triggered when we are replacing "UnsafeDeserialize"
+            // so we can assume that the node has at least one descendant when invoking .First()
+            var newMemberAccessExpression = generator.MemberAccessExpression(node.DescendantNodes().First(), "Deserialize");
```
twsouthwick, line 99: nit: magic string - can you move to a const field?

twsouthwick (changes requested):

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/BinaryFormatterUnsafeDeserializeAnalyzer.cs
mjrousos: Right now, UA only runs code fixes after migration, though. We should add a second source update step so that we can have one before TFM change and one after.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/GeneralInvocationExpression.cs
```diff
@@ -0,0 +1,57 @@
...
+{
+    public class GeneralInvocationExpression
+    {
+        private readonly SyntaxNode _invocationExpression;
+
+        private GeneralInvocationExpression(SyntaxNode syntaxNode)
+        {
+            // the only valid way to construct one is to try parse
+            // this enables validation to assert we're providing a wrapper around a valid SyntaxNode
+            _invocationExpression = syntaxNode ?? throw new ArgumentNullException(nameof(syntaxNode));
```
mjrousos, line 22: Why not just use IsInvocationExpression to do the necessary validation in the ctor and make it public? That seems like it would be more direct than having the TryParse helper (which isn't really *parsing* anything, per se).

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/GeneralMemberAccessExpression.cs
```diff
@@ -0,0 +1,84 @@
...
+        {
+            return _memberAccessExpression.Language switch
+            {
+                LanguageNames.CSharp => GetCSharpNode().Expression.GetQualifiedName(),
+                LanguageNames.VisualBasic => GetVisualBasicNode().Expression.GetQualifiedName(),
+                _ => throw new NotSupportedException(nameof(_memberAccessExpression.Language))
+            };
+        }
+
+        public static bool TryParse(SyntaxNode node, [MaybeNullWhen(false)] out GeneralMemberAccessExpression expression)
```
mjrousos, line 62: As before, the ctor seems ok for this. Also, if we keep this method I feel like a different name might fit better. Maybe 'TryWrap' or something like that?

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/BinaryFormatterUnsafeDeserializeAnalyzer.cs
```diff
@@ -39,71 +38,81 @@ public override void Initialize(AnalysisContext context)
...
+            context.RegisterSyntaxNodeAction(AnalyzeMemberAccessExpressions, CodeAnalysis.VisualBasic.SyntaxKind.SimpleMemberAccessExpression);
         }
 
         private void AnalyzeMemberAccessExpressions(SyntaxNodeAnalysisContext context)
         {
-            var memberAccessExpression = (MemberAccessExpressionSyntax)context.Node;
-
-            // If the accessed member isn't named "UnsafeDeserialize" bail out
-            if (!TargetMember.Equals(memberAccessExpression.Name.ToString(), StringComparison.Ordinal))
+            var isMemberAccessExpresion = GeneralMemberAccessExpression.TryParse(context.Node, out var expression);
```
mjrousos, line 47: Nit-pick: This can be combined -

```
if (!GeneralMemberAccessExpression(context.Node, out var expression) || expression is null)
{
    return;
}
```

mjrousos (approved):
Looks good! A few comments inline.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/BinaryFormatterUnsafeDeserializeAnalyzer.cs
KSchlobohm: do we leave this as-is for now or do we add a second source update step?

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/BinaryFormatterUnsafeDeserializeAnalyzer.cs
KSchlobohm: i saw you remembered VB case insensitive, so this isn't actually correct as-is

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/GeneralInvocationExpression.cs
KSchlobohm: checks to see if a node is of the correct type before we allow constructing it. because the two are paired together, it seemed like the tryParse flow.

maybe another option is something like
```
if (!syntaxNode.IsInvocationExpression())
{
   return;
}
var generalExpression = syntaxNode.AsGeneralInvocationExpression();
```

KSchlobohm: I went with "General" since "Generic" in csharp has a significant meaning. Does it seem appropriate? Feels a bit off to me but I haven't come up with a better name yet. Maybe `ProxyInvocationExpression`
@mjrousos 
@twsouthwick 
@sunandabalu

mjrousos: General seems ok to me.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/BinaryFormatterUnsafeDeserializeAnalyzer.cs
KSchlobohm: created an extension method. I didn't see one available already.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/GeneralMemberAccessExpression.cs
KSchlobohm: switched to ctor

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/AnalysisContextExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+using Microsoft.CodeAnalysis.Diagnostics;
+
+using CS = Microsoft.CodeAnalysis.CSharp;
+using VB = Microsoft.CodeAnalysis.VisualBasic;
+
+namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common
+{
+    public static class AnalysisContextExtensions
+    {
+        public static void RegisterSimpleMemberAccessExpression(this AnalysisContext analysisContext, Action<SyntaxNodeAnalysisContext> analyze, Action<SyntaxNodeAnalysisContext>? vbAnalyze = null)
```
twsouthwick, line 15: I like this. Maybe the Action can also take in the string identifiers so that's readily available as well without needing to know about vb/cs?

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/GeneralInvocationExpression.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using CS = Microsoft.CodeAnalysis.CSharp;
+using VB = Microsoft.CodeAnalysis.VisualBasic;
+
+namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common
+{
+    public class GeneralInvocationExpression
```
twsouthwick, line 14: Since this may end up being used a ton, could this be made into a `readonly struct`?

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/GeneralInvocationExpression.cs
```diff
@@ -0,0 +1,50 @@
...
+            {
+                throw new ArgumentException("Invalid type of syntaxNode", nameof(syntaxNode));
+            }
+
+            // the only valid way to construct one is to try parse
+            // this enables validation to assert we're providing a wrapper around a valid SyntaxNode
+            _invocationExpression = syntaxNode ?? throw new ArgumentNullException(nameof(syntaxNode));
+        }
+
+        public IEnumerable<SyntaxNode> GetArguments()
```
twsouthwick, line 30: nit: IEnumerable<> ends up allocating a bit. Not sure where this is on a fast path, but it may be better to just return the SyntaxNodeOrList

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/GeneralInvocationExpression.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+            return _invocationExpression.Language switch
+            {
+                LanguageNames.CSharp => GetCSharpNode().ArgumentList.Arguments.Select(x => x.Expression),
+                LanguageNames.VisualBasic => GetVisualBasicNode().ArgumentList.Arguments.Select(x => x.GetExpression()),
+                _ => throw new NotSupportedException(nameof(_invocationExpression.Language))
+            };
+        }
+
+        private VB.Syntax.InvocationExpressionSyntax GetVisualBasicNode()
```
twsouthwick, line 40: Could this be done as a `bool TryGet...Node(out VB.Syntax....)` so that we can easily guard it?

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/GeneralMemberAccessExpression.cs
```diff
@@ -0,0 +1,77 @@
...
+
+using System;
+using System.Diagnostics.CodeAnalysis;
+using Microsoft.CodeAnalysis;
+using CS = Microsoft.CodeAnalysis.CSharp;
+using VB = Microsoft.CodeAnalysis.VisualBasic;
+
+namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common
+{
+    public class GeneralMemberAccessExpression
```
twsouthwick, line 12: Same comments as the general invocation expression

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/GeneralMemberAccessExpression.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            return _memberAccessExpression.Language switch
+            {
+                LanguageNames.CSharp => GetCSharpNode().Expression.GetQualifiedName(),
+                LanguageNames.VisualBasic => GetVisualBasicNode().Expression.GetQualifiedName(),
+                _ => throw new NotSupportedException(nameof(_memberAccessExpression.Language))
+            };
+        }
+
+        private VB.Syntax.MemberAccessExpressionSyntax GetVisualBasicNode()
```
twsouthwick, line 67: It may be nice to have an interface that defines some of these common methods so we can ensure the signature stays the same.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/SyntaxNodeExtensions.Predicates.cs
```diff
@@ -0,0 +1,105 @@
...
+using CS = Microsoft.CodeAnalysis.CSharp;
+using CSSyntax = Microsoft.CodeAnalysis.CSharp.Syntax;
+using VB = Microsoft.CodeAnalysis.VisualBasic;
+using VBSyntax = Microsoft.CodeAnalysis.VisualBasic.Syntax;
+
+namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Default
+{
+    public static partial class SyntaxNodeExtensions
+    {
+        public static bool IsCSharp(this SyntaxNode node) => node?.Language == LanguageNames.CSharp;
```
twsouthwick, line 14: Is the `.Predicates` because of the `Is` prefix?

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/BinaryFormatterUnsafeDeserializeAnalyzer.cs
```diff
@@ -39,71 +40,81 @@ public override void Initialize(AnalysisContext context)
...
-            // If the accessed member isn't named "UnsafeDeserialize" bail out
-            if (!TargetMember.Equals(memberAccessExpression.Name.ToString(), StringComparison.Ordinal))
+            if (!context.Node.IsMemberAccessExpression())
             {
                 return;
             }
 
-            // Continue, only if the call is to a method called UnsafeDeserialize
-            if (!(memberAccessExpression.Parent is InvocationExpressionSyntax))
+            var expression = new GeneralMemberAccessExpression(context.Node);
```
twsouthwick, line 53: Maybe pass this in if it is a valid expression so this method will only be hit if it's available

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/BinaryFormatterUnsafeDeserializeAnalyzer.cs
```diff
@@ -39,71 +40,81 @@ public override void Initialize(AnalysisContext context)
...
-                if (!symbolText.StartsWith(TargetTypeSymbolNamespace, StringComparison.Ordinal)
-                    || !symbolText.EndsWith(TargetTypeSymbolName, StringComparison.Ordinal))
-                {
-                    return;
-                }
+                return false;
             }
-            else if (accessedSymbol != null)
+
+            var methodSymbol = (IMethodSymbol)theSymbol;
```
twsouthwick, line 98: This may be a good method off of `NameMatcher` to check if it's a constructor

twsouthwick (approved):
I left a few comments, but nothing blocking.

We're getting to a point with a few different ways of tackling similar things, so it may be worth discussing which one we want to use going forward and consolidate the code base. I'm thinking of things like the `Is...` vs `generalized...Expression`, the `Microsoft.CodeAnalysis.Testing` vs our testing framework, etc. We have some good examples now so we can probably come up with a list of pros/cons to make the decision for going forward.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/SyntaxNodeExtensions.Predicates.cs
KSchlobohm: yes. Just splitting up the file, as it seemed to be getting large and unwieldy.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/GeneralInvocationExpression.cs
KSchlobohm: Will review as part of a larger conversation about how we handle roslyn abstractions

PR #558, opened 2021-05-27
Use assembly load context for extensions

twsouthwick:
If an extension contains multiple assemblies, they will now resolve from within the extension thanks to a private load context. We want to ensure that we don't load a different version of anything the default context loads, so this checks for that as well.


src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/Microsoft.DotNet.UpgradeAssistant.Abstractions.csproj
```diff
@@ -17,6 +17,7 @@
     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
     <PackageTags>Upgrade Assistant</PackageTags>
     <PackageIcon>icon.png</PackageIcon>
+    <DevelopmentDependency>true</DevelopmentDependency>
```
mjrousos, line 20: What's this for? I'm not familiar with this property.

src/components/Microsoft.DotNet.UpgradeAssistant.Extensions/ExtensionAssemblyLoadContext.cs
```diff
@@ -0,0 +1,55 @@
...
+
+            if (inDefault)
+            {
+                return null;
+            }
+
+            var dll = $"{assemblyName.Name}.dll";
+            var dllFile = _extension.FileProvider.GetFileInfo(dll);
+
+            var pdb = $"{assemblyName.Name}.pdb";
```
mjrousos, line 34: Super-nit-pick: You don't need to get the file info for the PDB if the dll doesn't exist, so these lines can move inside the first `if`.

src/components/Microsoft.DotNet.UpgradeAssistant.Extensions/ExtensionAssemblyLoadContext.cs
```diff
@@ -0,0 +1,55 @@
...
+            var pdbFile = _extension.FileProvider.GetFileInfo(pdb);
+
+            if (dllFile.Exists)
+            {
+                using var dllStream = dllFile.CreateReadStream();
+
+                if (pdbFile.Exists)
+                {
+                    using var pdbStream = pdbFile.CreateReadStream();
+                    return LoadFromStream(pdbStream);
```
mjrousos, line 44: Shouldn't this be `return LoadFromStream(dllStream, pdbStream)`?

mjrousos (changes requested):
Good change; thanks! Only real concern is that I think you might be returning the wrong stream in the case where an extension has a dll with symbols.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/Microsoft.DotNet.UpgradeAssistant.Abstractions.csproj
twsouthwick: This is so the resulting nupkg will automatically be added with `PrivateAssets=all` so they won't be in the output

mjrousos: Wouldn't devs depending on the abstractions package want these assets in their output? These are typically UA extension scenarios and the common libraries will need to be present at run time, not just at build time like analzyers.

Is this change because UA will be loading the abstractions library from its own binaries in the default load context rather than from the extensions?

twsouthwick:
@mjrousos This change became a bit bigger but also ensures we're dealing with the default extension in the *same* way as others. The build process will now move them to a subfolder and then at runtime we'll only load from there.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/Microsoft.DotNet.UpgradeAssistant.Abstractions.csproj
twsouthwick: Correct. It's a dev dependency precisely because we don't want it in the output. We need a unified type that we supply and don't want to accidently load another abstractions library in a different context. That was happening and causes weird type load exception.

We do guard against this explicitly by first checking if any type is already loaded in the main context, but since we don't need it there, it's just extra baggage.

mjrousos: Makes sense. I have some anxiety about scenarios where someone implements something on top of the abstractions library and then uses it outside of UA (in which case the abstractions binaries maybe should be present) but I can't think of a likely scenario where that would happen, so this is probably good.

twsouthwick: Sure - they can always remove the private assets then if they really want to build on it themselves. This just helps the main use case to be in a good place.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/Microsoft.DotNet.UpgradeAssistant.Abstractions.csproj
```diff
@@ -28,6 +29,15 @@
     <PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" />
     <PackageReference Include="Microsoft.Extensions.FileProviders.Abstractions" />
     <PackageReference Include="System.Linq.Async" />
+    <PackageReference Include="System.Text.Json" />
   </ItemGroup>
 
+  <Target Name="PrivateAssetsIfExtension" Condition=" '$(IsExtension)' == 'true' " BeforeTargets="CollectPackageReferences">
```
mjrousos, line 35: Is this updating extensions to not carry NuGet-referenced libraries with them? I would think they might need those assets at runtime if UA doesn't have the same NuGet dependencies.

src/components/Microsoft.DotNet.UpgradeAssistant.Extensions/ExtensionAssemblyLoadContext.cs
```diff
@@ -0,0 +1,57 @@
...
+namespace Microsoft.DotNet.UpgradeAssistant.Extensions
+{
+    internal class ExtensionAssemblyLoadContext : AssemblyLoadContext
+    {
+        private const string ALC_PREFIX = "UA_";
+
+        private readonly ExtensionInstance _extension;
+
+        public ExtensionAssemblyLoadContext(ExtensionInstance extension)
+            : base(ALC_PREFIX + extension.Name)
```
mjrousos, line 18: Interesting, were you hitting name conflicts here?

src/components/Microsoft.DotNet.UpgradeAssistant.Extensions/ExtensionAssemblyLoadContext.cs
```diff
@@ -0,0 +1,57 @@
...
+using System;
+using System.Linq;
+using System.Reflection;
+using System.Runtime.Loader;
+
+namespace Microsoft.DotNet.UpgradeAssistant.Extensions
+{
+    internal class ExtensionAssemblyLoadContext : AssemblyLoadContext
+    {
+        private const string ALC_PREFIX = "UA_";
```
mjrousos, line 13: Nit: `ALC_Prefix`, maybe, to match casing of our other private const variables?

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Package/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Package.csproj
```diff
@@ -4,7 +4,6 @@
     <IncludeBuildOutput>false</IncludeBuildOutput>
     <IsPackable>true</IsPackable>
     <SuppressDependenciesWhenPacking>true</SuppressDependenciesWhenPacking>
-    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
```
mjrousos, line 7: This isn't needed anymore?

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.ProjectFormat/TryConvertTool.cs
```diff
@@ -72,7 +73,7 @@ public Task<bool> RunAsync(IUpgradeContext context, IProject project, Cancellati
 
             return _runner.RunProcessAsync(new ProcessInfo
             {
-                Command = Path,
+                Command = DotNetCli,
```
mjrousos, line 76: Why the change to .NET CLI?

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/Microsoft.DotNet.UpgradeAssistant.Steps.Razor.csproj
```diff
@@ -12,8 +12,7 @@
     <PackageReference Include="DiffPlex" />
     <PackageReference Include="Microsoft.AspNetCore.Mvc.Razor.Extensions" />
     <PackageReference Include="Microsoft.AspNetCore.Razor.Language" />
-    <PackageReference Include="Microsoft.Bcl.Hashcode" />
     <PackageReference Include="Microsoft.CodeAnalysis.Razor" />
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" PrivateAssets="all" />
```
mjrousos, line 16: Why not copy these assets to output?

mjrousos (approved):
Looks good! A few questions inline about why certain changes were made, but looks good overall.

src/components/Microsoft.DotNet.UpgradeAssistant.Extensions/ExtensionAssemblyLoadContext.cs
twsouthwick: The name of our default extension is Default and the default alc is Default

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Package/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Package.csproj
twsouthwick: No - we doa separate pack command. I think this is in the template so people don't need to do both a build and pack. It isn't necessary and was causing some build errors on the ci

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.ProjectFormat/TryConvertTool.cs
twsouthwick: Because we got too long of a path with the shim that the dotnet tool installed now. I've changed it to grab just the try-convert bin directory now, which is just the managed stuff. Probably better any way now that we're going to support mac

sunandabalu: yeah the try-convert tool install path has become really long :)

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/Microsoft.DotNet.UpgradeAssistant.Steps.Razor.csproj
twsouthwick: It gets filtered out as it's in the output already, but reverting this

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/Microsoft.DotNet.UpgradeAssistant.Abstractions.csproj
twsouthwick: Not needed anymore. It made the abstractions libraries local as they'll be in the main output

PR #559, opened 2021-05-27
Add a Razor document updater to enabled upgrading @helper functions to local methods

### LINKED ISSUE #76: As a user, I'm alerted if I'm using @helper directives that will need updated

mjrousos:
Fixes #76


src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/HelperMatcher.cs
```diff
@@ -0,0 +1,170 @@
...
+namespace Microsoft.DotNet.UpgradeAssistant.Steps.Razor
+{
+    /// <summary>
+    /// Determines whether a Razor document contains @helper functions
+    /// and produces local method replacements for the @helpers.
+    /// </summary>
+    public class HelperMatcher : IHelperMatcher
+    {
+        // Regex used to identify @helper functions
+        private static readonly Regex HelperDeclRegex = new Regex(@"@helper\s+.*?\(.*?\)", RegexOptions.Compiled);
```
twsouthwick, line 23: Will this match if there are nested parentheses?

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/HelperMatcher.cs
```diff
@@ -0,0 +1,170 @@
...
+            var lineStart = index;
+
+            // Copy the helper function until the end of its body block
+            while ((!functionBodyBegun || braceCount > 0) && index < text.Length)
+            {
+                var nextChar = text[index];
+
+                // Match braces while moving through the function body so that
+                // the loop can end when all braces are matched
+                braceCount += nextChar switch
```
twsouthwick, line 128: What about a lone brace in a string?

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/HelperReplacement.cs
```diff
@@ -0,0 +1,40 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Text;
+
+namespace Microsoft.DotNet.UpgradeAssistant.Steps.Razor
+{
+    public class HelperReplacement
```
twsouthwick, line 9: This type seems like it could be a general purpose text replacement holder

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/IHelperMatcher.cs
```diff
@@ -0,0 +1,16 @@
...
+
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Razor.Language;
+
+namespace Microsoft.DotNet.UpgradeAssistant.Steps.Razor
+{
+    public interface IHelperMatcher
+    {
+        Task<IEnumerable<HelperReplacement>> GetHelperReplacementsAsync(RazorCodeDocument document);
```
twsouthwick, line 12: IAsyncEnumerable?

twsouthwick: Still not sure about when to use one over the other, but I like the idea of lazy calculations when possible

twsouthwick: Looking at usage, you could get rid of the `HasHelperAsync` with an IAsyncEnumerable and just do an `AnyAsync` on it

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/HelperMatcher.cs
```diff
@@ -0,0 +1,170 @@
...
+            if (text is null)
+            {
+                _logger.LogWarning("Failed to read Razor document {FilePath}", path);
+                return Enumerable.Empty<HelperReplacement>();
+            }
+
+            var helpers = HelperDeclRegex.Matches(text);
+
+            // This would be a lot more succinct with Linq,
+            // but MatchCollection doesn't implement IEnumerable<Match>.
```
twsouthwick, line 103: It does I believe in .NET 5 :)

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorHelperUpdater.cs
```diff
@@ -0,0 +1,153 @@
...
+            if (context is null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            // Track which documents are updated for inclusion in the updater result returned later
+            var updatedDocuments = new List<string>();
+            foreach (var document in inputs)
+            {
+                var path = document.Source.FilePath;
```
twsouthwick, line 109: I don't know if you can assume they all have a valid path you can read from.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorHelperUpdater.cs
```diff
@@ -0,0 +1,153 @@
...
+                    _logger.LogInformation("Replacing {HelperCount} @helper functions in {FilePath}", helperReplacements.Count(), path);
+                    foreach (var helper in helperReplacements.OrderByDescending(h => h.Offset))
+                    {
+                        helper.Apply(newText);
+                    }
+
+                    // Check whether the document has a using statement for HelperResult's namespace (Microsoft.AspNetCore.Mvc.Razor) and
+                    // add one if none is present. Again, this is done with string manipulation since there's no good API for updating
+                    // Razor syntax.
+                    if (!HasMvcRazorUsing(document))
```
twsouthwick, line 127: Could we check if there's an `_Imports.cshtml` that would have this?

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorHelperUpdater.cs
```diff
@@ -0,0 +1,153 @@
...
+                    {
+                        _logger.LogInformation("Adding an import statement for the '{namespace}' namespace to {FilePath}", HelperResultNamespace, path);
+                        newText.Insert(0, $"@using {HelperResultNamespace}{Environment.NewLine}");
+                    }
+
+                    _logger.LogDebug("Writing updates to {FilePath} to replace @helper functions", path);
+                    File.WriteAllText(path, newText.ToString());
```
twsouthwick, line 134: Can't we use a SourceText or something so we go through that abstraction rather than directly to the File IO?

tests/Microsoft.DotNet.UpgradeAssistant.TestHelpers/FileHelpers.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.DotNet.UpgradeAssistant
+{
+    public static class FileHelpers
+    {
+        private static readonly string[] DirsToIgnore = new[] { "bin", "obj" };
+
+        public static async Task CopyDirectoryAsync(string sourceDir, string destinationDir)
```
twsouthwick, line 15: Could we have this as an object that tracks the copied files and exposes an IDisposable to clean them up?

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor.Tests/RazorHelperUpdaterTests.cs
```diff
@@ -0,0 +1,193 @@
...
+                Assert.Equal(File.ReadAllText(e.FullName), File.ReadAllText(a.FullName));
+            }).ToArray());
+        }
+
+        public static IEnumerable<object[]> HelperTestViews =>
+            new List<object[]>
+            {
+                new object[]
+                {
+                    "RazorHelperUpdaterViews/OneHelper/Test.csproj",
```
twsouthwick, line 121: Can we mock this stuff without actually having full on projects? Something similar to the Microsoft.CodeAnalysis.Testing?

tests/tool/Integration.Tests/E2ETest.cs
```diff
@@ -50,15 +48,15 @@ public async Task UpgradeTest(string scenarioPath, string inputFileName, string
 
             // Copy the scenario files to the temporary directory
             var scenarioDir = Path.Combine(IntegrationTestAssetsPath, scenarioPath);
-            await CopyDirectoryAsync(Path.Combine(scenarioDir, OriginalProjectSubDir), workingDir).ConfigureAwait(false);
+            await FileHelpers.CopyDirectoryAsync(Path.Combine(scenarioDir, OriginalProjectSubDir), workingDir).ConfigureAwait(false);
```
twsouthwick, line 51: this pattern might be cleaner if instead of static methods it's an object that we optional clean up (IDisposable may not work as I like to have the files available on test failures)

twsouthwick (approved):
A left a number of comments, but nothing blockers. Take them or leave them, but otherwise, LGTM

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/HelperMatcher.cs
mjrousos: It actually doesn't need to worry about nested parentheses because we're actually only working with the `@helper (` part of the declaration. I'll change the regex to only match up to there to make that clearer.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/HelperMatcher.cs
mjrousos: Hm... that's a good point. This is where working with the Razor syntax tree would come in handy, but the .NET Core Razor engine doesn't understand @helper syntax so it isn't able to properly evaluate this code block. I may leave this as it is for now knowing that there are some corner cases it won't handle well and we can re-visit if they become problematic in practice.

Possible solutions would range from counting `'` and `"`, too, to give a best-effort attempt at only match real {} to (if we're really serious) creating Razor engine extensions to teach it about @helper syntax so it can parse this properly.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/HelperReplacement.cs
mjrousos: Yea, I considered that. I already have a `TextReplacement` type though for source replacement which functions a little differently from this one (it contains information on text in a generated source file that needs to be replaced on a particular line in a Razor document).

Maybe I'll rename that one to something else, though (MappedTextReplacement?) so that this can be TextReplacement.

mjrousos: Ok, I've updated the file names to be (hopefully) more sensible.

tests/Microsoft.DotNet.UpgradeAssistant.TestHelpers/FileHelpers.cs
mjrousos: Good idea. I'll do that.

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor.Tests/RazorHelperUpdaterTests.cs
mjrousos: That's a good question. The challenge will be that we need RazorCodeDocuments (and the generated C# that the Razor engine produces for them) and I'm not sure whether we can get that from any existing test helpers. We'll probably need to write our own framework that takes text for before and after razor docs and checks that code fixes apply as expected. I'm sure we could do that, but it would be a bit of an investment. I'll think about whether it's something I want to add now or later.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorHelperUpdater.cs
mjrousos: That will happen automatically. That's why `HasMvcRazorUsing` is looking at the generated C# - that source will include source from imported documents in addition to the source from the primary document.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/IHelperMatcher.cs
mjrousos: Yea, that's a better API shape. I'll make that update.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorHelperUpdater.cs
mjrousos: I'll check that FilePath is non-null and exists. I don't think there's a better way to get the text, though, because `RazorSourceDocument` doesn't have any public APIs for getting the contents of the document.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/IHelperMatcher.cs
mjrousos: I've updated IHelperMatcher to have a single method returning an enumerable.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Razor/RazorHelperUpdater.cs
mjrousos: That would be better. We'd still need a TextWriter to give the SourceText, though. We could maybe include a file IO abstraction in DI. I'm going to leave that for a future update, though, to keep this one focused.

PR #573, opened 2021-06-01
Remove central version management

twsouthwick:
This seems to be causing problems with the way we're building our
extensions. Somehow it's detecting that we have a version when no
PackageReferences do. Since this is a preview feature, it appears we may
have hit a bug with that. For now, swapping back to non-centrally
managed versions allows for a more seamless inner dev-loop.


mjrousos (approved):
LGTM. Hopefully we can add this back in the future.

PR #581, opened 2021-06-02
Add tests for extension options

twsouthwick:


tests/components/Microsoft.DotNet.UpgradeAssistant.Extensions.Tests/ExtensionConfigurationTests.cs
```diff
@@ -0,0 +1,140 @@
...
+            };
+
+            using var provider = CreateServiceProvider(manifests);
+
+            // Act
+            var options = provider.GetRequiredService<IOptions<OptionsWithArray>>();
+
+            // Assert
+            var o = options.Value;
+            Assert.Equal(new[] { 3, 4 }, options.Value.Array);
```
mjrousos, line 81: We should also confirm that `GetRequiredService<IOptions<ICollection<OptionsWithArray>>` gives both arrays back.

mjrousos (approved):
LGTM. One suggestion for an additional test case inline.

PR #582, opened 2021-06-02
Separate out web related items into its own extension

twsouthwick:
This removes most of the web-specific info to a separate extension so that it can have web-related things co-located with it. The analyzers/codefixers are not self-contained, but that may require a bit more thinking as we distribute them. This is a first step towards organizing extensions by technology areas.

The resulting output looks like this:

```
upgrade-assistant
|- Microsoft.DotNet.UpgradeAssistant.Cli.dll
|- ....
|- extensions
  |- default
    |- [files]
  |- web
    |- [files]
```


src/extensions/web/Microsoft.DotNet.UpgradeAssistant.Extensions.Web/WebExtension.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        public void AddServices(IExtensionServiceCollection services)
+        {
+            if (services is null)
+            {
+                throw new ArgumentNullException(nameof(services));
+            }
+
+            services.Services.AddTransient<IDependencyAnalyzer, NewtonsoftReferenceAnalyzer>();
+            services.Services.AddScoped<IUpdater<ConfigFile>, WebNamespaceConfigUpdater>();
```
mjrousos, line 20: I know you didn't change this, but the Transient/Scoped difference stands up when these are side-by-side here :) I don't think there's any reason the config updater needs to have a scoped lifetime. We may want to make it transient for consistency.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Configuration/ConfigUpdaterStep.cs
```diff
@@ -54,7 +57,7 @@ public ConfigUpdaterStep(IEnumerable<IUpdater<ConfigFile>> configUpdaters, IOpti
                 throw new ArgumentNullException(nameof(configUpdaterOptions));
             }
 
-            _configFilePaths = configUpdaterOptions.Value?.ConfigFilePaths ?? Array.Empty<string>();
+            _configFilePaths = configUpdaterOptions.Value.SelectMany(s => s.ConfigFilePaths).ToArray();
```
mjrousos, line 60: Nit-pick - we may want a .Distinct in case different extensions add the same config file.

mjrousos (approved):
LGTM; thanks for splitting this apart!

src/extensions/web/Microsoft.DotNet.UpgradeAssistant.Extensions.Web/WebExtension.cs
twsouthwick: Good call. I've created an issue to address the bigger picture here of how we decide to scope things.

PR #583, opened 2021-06-02
Update extension samples

mjrousos:
We've had some churn recently in our extension framework. This updates the samples to build and work with the new model and adds them to the solution so that they'll continue to be built by CI in the future.


samples/PackageMapSample/PackageMapSample.sln
```diff
@@ -0,0 +1,34 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
```
twsouthwick, line 2: Why is this solution needed?

twsouthwick (approved):
LGTM. Not sure what the extra sln file is for. If desired, you can create asolution filter that only has the samples and the abstractions.

samples/PackageMapSample/PackageMapSample.sln
mjrousos: I just made it to match the pattern of the other existing samples (which each have their own sln in addition to being part of the Microsoft.DotNet.UpgradeAssistant.sln. We may not even need those slns anymore, I just left them (and added this to match) because I didn't see a compelling reason to get rid of them.

mjrousos: Thinking more about it, we don't really need the sample-specific solutions anymore, so I may just get rid of those.

twsouthwick: Not sure the value either. If you want a samples solution, a solution filter would be a better option.

PR #597, opened 2021-06-03
Include web typemap in analyzers test

twsouthwick:
Not sure how the CI build was passing without this, but tests seem to be broken. A previous PR moved the web typemap so it wouldn't show up in the default extension list, which meant at runtime it was not in the test folder. This links the files so it'll show up.

### DIFF
--- a/tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test.csproj
+++ b/tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test.csproj
@@ -14,6 +14,9 @@
     <Content Include="assets\**">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
+    <None Include="..\..\..\..\..\src\extensions\default\analyzers\Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Package\build\WebTypeReplacements.typemap" Link="WebTypeReplacements.typemap">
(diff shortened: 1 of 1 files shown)


tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test.csproj
```diff
@@ -14,6 +14,9 @@
     <Content Include="assets\**">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
+    <None Include="..\..\..\..\..\src\extensions\default\analyzers\Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Package\build\WebTypeReplacements.typemap" Link="WebTypeReplacements.typemap">
```
mjrousos, line 17: If WebTypeReplacements.typemap lived in the analyzers project and the analyzer package project linked it from there, then it would show up in projects like this which depend on the analyzers automatically. Linking is ok, but it's nice to keep it to a minimum since it adds file path-specific dependencies outside of the current project.

mjrousos (approved):
LGTM, though it makes me wonder if WebTypeReplacements.typemap is living in the right project. Also, very weird that CI builds didn't fail without this.

tests/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Test.csproj
twsouthwick: Agreed. The reason it was done like this is so the webtypereplacements.typemap doesn't show up in the default extension.

PR #599, opened 2021-06-03
Update HttpContextHelper to use static instance

twsouthwick:


src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CodeFixes/Templates/HttpContextHelper.cs
```diff
@@ -1,29 +1,22 @@
...
-        }
+        private static readonly HttpContextAccessor HttpContextAccessor = new HttpContextAccessor();
 
         /// <summary>
-        /// Gets the current HttpContext. Returns null if there is no current HttpContext.
+        /// Gets the current <see cref="HttpContext"/>. Returns <c>null</c> if there is no current <see cref="HttpContext"/>.
         /// </summary>
-        public static HttpContext Current => HttpContextAccessor?.HttpContext;
+#if NET5_0_OR_GREATER
+        [Obsolete("Prefer accessing HttpContext via injection", error: false, DiagnosticId = "HttpContextCurrent", UrlFormat = "https://docs.microsoft.com/en-us/aspnet/core/fundamentals/http-context")]
```
mjrousos, line 18: Is there a way to make this informational instead of a warning? We want people to move away from it, but a lot of customers are going to want to use it in the short-term and I can imagine the extra warning would be frustrating.

mjrousos (approved):
LGTM

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CodeFixes/Templates/HttpContextHelper.cs
twsouthwick: Not with this pattern. It's their own code though so they can just remove it if they want. They can also set level in rulesets or editorconfig

mjrousos: Makes sense. We can keep an eye on this and remove the obsolete later if it's too much of a hassle for customers, but we can leave it like this for now and see how it goes.

PR #602, opened 2021-06-07
Redirect bin/obj folders for extensions to minimize file write races

twsouthwick:
When building on the CI, there are some race conditions that occur
periodically to write files to disk. This is because the extensions are
built separately for their own projects and for when they are being
included in the CLI. This change sets a custom OutDir and
IntermediateOutputPath when needed so there are separate places these
get written.

### DIFF
--- a/Extensions.targets
+++ b/Extensions.targets
@@ -5,8 +5,17 @@
     <IsCI Condition=" '$(BuildingInsideVisualStudio)' == '' ">true</IsCI>
   </PropertyGroup>
 
+  <!-- If building an extension with a _ExtensionIntermediateOutputPath, we want to redirect bin and obj folders to reduce file lock conflicts -->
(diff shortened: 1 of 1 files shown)


Extensions.targets
```diff
@@ -15,18 +24,19 @@
     </ItemGroup>
   </Target>
 
-  <Target Name="_AddMetadataToExtensions" BeforeTargets="RestoreExtensions;PublishUpgradeAssistantExtensions">
+  <Target Name="_AddMetadataToExtensions"
+          BeforeTargets="RestoreExtensions;PublishUpgradeAssistantExtensions">
     <!-- Add the relative directory the extension will be added to -->
```
sunandabalu, line 29: some weird formatting going on here, I don't see any other changes in this line.

sunandabalu (approved):

Extensions.targets
twsouthwick: It's intentional to align the attributes to read things better.

PR #603, opened 2021-06-07
Roslyn best practices doc

KSchlobohm:
**PR Title**
Documents best practices for building Roslyn Analyzers

**PR Description**
Adds link from code of conduct to new document that distills lessons learned while implementing Roslyn analyzers.


docs/roslyn_best_practices.md
```diff
@@ -0,0 +1,181 @@
+# Roslyn Best Practices
+
+We use [Roslyn Analyzers](https://docs.microsoft.com/en-us/visualstudio/code-quality/roslyn-analyzers-overview) highlight areas of code that will need to be refactored. By default our goal is to pair these Analyzers with Code Fixers to automate as much of the upgrade workflow as possible.
```
twsouthwick: It may be good to call out some definitions and describe what analyzers and codefixers are at a high level to level set the concept with potential new contributors (and send them off to the docs site for a more detailed description).

Maybe include terms like: Analyzer, Code Fixer, Language agnostic, etc

docs/roslyn_best_practices.md
```diff
@@ -0,0 +1,181 @@
...
+
+We support the following types of fixes:
```
twsouthwick, line 5: What's this list? Maybe more of an example list of things that can be done?

docs/roslyn_best_practices.md
```diff
@@ -0,0 +1,181 @@
...
+If you track multiple syntax nodes in your Analyzer then you can make it easier to write your Code Fixer by using the method overload of `Diagnostic.Create` that supports [additionalLocations](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.diagnostic.create#Microsoft_CodeAnalysis_Diagnostic_Create_Microsoft_CodeAnalysis_DiagnosticDescriptor_Microsoft_CodeAnalysis_Location_System_Collections_Generic_IEnumerable_Microsoft_CodeAnalysis_Location__System_Collections_Immutable_ImmutableDictionary_System_String_System_String__System_Object___)
+
+### 5. Beware of Trivia
+When searching for patterns of code it is common to perform string comparisons. These approaches can work successfully but you should beware that trivia, the spacing we use to make code more readable, can vary from team to team. Not everyone puts a space after an assignment operator and if there are two spaces, or a tab, after the assignment operator that code is still valid.
+
+**Should Not**
+* Try to avoid `ToFullString` and solutions that require awareness of trivia.
+
+### 6. Report the diagnostic only on parts of code that would be changed by a code fixer.
```
twsouthwick, line 88: What about focusing this practice as "Report the diagnostic only on part of the code that need to be changed". Whether or not that is manually or via a code fixer seems irrelevant.

docs/roslyn_best_practices.md
```diff
@@ -0,0 +1,181 @@
...
+https://docs.microsoft.com/en-us/dotnet/csharp/roslyn-sdk/syntax-visualizer
+
+The Syntax Visualizer is a tool window that helps you inspect and explore syntax trees. It's an essential tool to understand the models for code you want to analyze. It's also a debugging aid when you develop your own applications using the .NET Compiler Platform (“Roslyn”) SDK.
+
+### 2. Look at examples
+* [Roslyn-analyzers](https://github.com/dotnet/roslyn-analyzers)
+* [Roslynator](https://github.com/JosefPihrt/Roslynator)
+
+### 3. Video Training Courses
+* [Youtube: Learn Roslyn Now](https://www.youtube.com/watch?v=wXXHd8gYqVg) by Josh Varty
```
KSchlobohm, line 181: Do we have other courses that we would recommend?

This looked relevant but I have not finished this course:
https://www.pluralsight.com/courses/vs-2015-diagnostic-analyzers-first-look

docs/roslyn_best_practices.md
```diff
@@ -0,0 +1,181 @@
...
+
+**Do not**
+* Use `.Result` or `Wait()` in a Roslyn Analyzer. Forcing asynchronous code the behave synchronously can result in timing issues that are hard to debug and thread starvation.
+
+## Best Practices for Roslyn Code Fixers
+
+### 1. Code fixers handle well-known scenarios
+You may need to do a few final checks before registering your code fixer by calling `RegisterCodeFix` but you should lean into doing as many checks as possible in the analyzer.
+
+You can also accelerate your code fixer productivity by using the additionalLocations overload when reporting a diagnostic so that you can find multiple interesting locations in a document when working from a single diagnostic.
```
twsouthwick, line 102: nit: surround items like `additionalLocations` with a tick so it stands out as a constant concept

docs/roslyn_best_practices.md
```diff
@@ -0,0 +1,181 @@
...
+We support the following types of fixes:
+* Mapping from one type to another one common example is [System.Web.Http.ApiController](https://docs.microsoft.com/en-us/previous-versions/aspnet/hh834453(v=vs.118)) should become [Microsoft.AspNetCore.Mvc.ControllerBase](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.controllerbase)
+* Extracting HttpContext via method dependency injection
+* Replacing methods that are not available on .NET latest such as [BinaryFormatter.UnsafeDeserialize](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.serialization.formatters.binary.binaryformatter.unsafedeserialize)
+
+The following guidelines are a list of best practices that we use to guide the development of Analyzers and Code Fixers to support our goals.
+
+### Our Goals
+
+1. Write language agnostic analyzers with as much code reuse as possible.
```
KSchlobohm, line 14: Do we want to use the word `agnostic` or `indifferent`?

docs/roslyn_best_practices.md
```diff
@@ -0,0 +1,181 @@
...
+
+### 2. Symbols may not be resolvable
+`upgrade-assistant` will add, upgrade, and remove package references during the upgrade process. If your analyzer is looking for a specific symbol then you may need to [AddMetadataReference](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.project.addmetadatareference) to ensure the symbol is available.
+
+### 3. Add reference assemblies instead of the assembly implementation
+Reference assemblies are a special type of assembly that have only the smallest amount of metadata needed to represent the library's public API surface.
+
+If you need to [AddMetadataReference](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.project.addmetadatareference) then use a reference assembly when possible.
+
+Examples include:
```
KSchlobohm, line 165: Are there other metadata reference assemblies we could include?

docs/roslyn_best_practices.md
```diff
@@ -0,0 +1,181 @@
...
+
+Learn more about Async best practices by reading [Async/Await - Best Practices in Asynchronous Programming](https://docs.microsoft.com/en-us/archive/msdn-magazine/2013/march/async-await-best-practices-in-asynchronous-programming) by Stephen Cleary
+
+**Do**
+* Leverage the CancellationToken to interrupt processing when cancellation has been requested.
+
+**Do not**
+* Do not force async code to run synchronously. Avoid calling `.Result` and prefer to use await as needed when working in the code fixer construct.
+
+### 5. When using CodeAction.Create prefer Document over Solution
```
twsouthwick, line 126: Is this a best practice? What about limiting the smallest scope required?

KSchlobohm (commented):
Opened questions for reviewers

docs/roslyn_best_practices.md
```diff
@@ -0,0 +1,181 @@
...
+Your code fixer should override `GetFixAllProvider` to return a non-null instance of [FixAll Provider](https://github.com/dotnet/roslyn/blob/main/docs/analyzers/FixAllProvider.md).
+
+### 8. Treat the Code Fixer as a single unit of work
+Code fixers are applied as a one-step change and their order is not guaranteed. If the code you’re adding requires a namespace import then your code fixer should perform that change if necessary. Do not assume that another code fixer will be responsible for adding the correct namespace.
+
+## Considerations specific to upgrade-assistant
+
+There are many stopping points on the journey from .NET Framework to .NET latest and even more scenarios specific to our customers. Because of this, `upgrade-assistant` does not always assume that the code can be compiled. This section highlights some considerations specific to the "work in progress" state that all projects will pass through as they are upgraded.
+
+### 1. Do not assume the code will compile
```
twsouthwick, line 152: I'd also add consider if it's a change that can be done while still targeting .NET Framework. Those can assume things will compile. Only code that we want to function in that transition point should not.

Ie, migrating HttpContext to method injection should still compile and can be done before moving to .NET Core.

docs/roslyn_best_practices.md
```diff
@@ -0,0 +1,181 @@
...
+`upgrade-assistant` will make sweeping changes across the solution. These sweeping changes often result in manual changes that must be made after running upgrade assistant before the code will compile.
+
+As an example, `upgrade-assistant` will upgrade NuGet packages across major versions. `upgrade-assistant` does not evaluate if the newer package contains breaking changes.
+
+### 2. Symbols may not be resolvable
+`upgrade-assistant` will add, upgrade, and remove package references during the upgrade process. If your analyzer is looking for a specific symbol then you may need to [AddMetadataReference](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.project.addmetadatareference) to ensure the symbol is available.
+
+### 3. Add reference assemblies instead of the assembly implementation
```
twsouthwick, line 160: I bet we could come up with a way to have analzyer declare what reference assemblies they require to simplify this

docs/roslyn_best_practices.md
KSchlobohm: > Maybe more of an example list of things that can be done?

Yes. Seems we should reword this to clarify. How about `Example of changes that we can perform include:`

docs/roslyn_best_practices.md
KSchlobohm: That's the intent of this one, I'll reword the title

docs/roslyn_best_practices.md
KSchlobohm: Wanted to avoid an enumeration of reasons code would not compile here. There are a few, but this one seemed like one that everyone could get onboard with even with minimal awareness of .NET latest.

docs/roslyn_best_practices.md
KSchlobohm: Potentially, I'm not sure we actually want reference assemblies. Adding them certainly increases accuracy but I think we are able to reach a reasonable level of certainty without the reference.

Having said that, I wonder if we care about AddMetadataReference

docs/roslyn_best_practices.md
twsouthwick: I believe the line of thought I am going down deserves it's own issue and discussion so I created https://github.com/dotnet/upgrade-assistant/issues/604 to track that. Can punt it to there for now :)

docs/roslyn_best_practices.md
```diff
@@ -0,0 +1,189 @@
...
+Roslyn sees class files as rich trees of information. There will likely be millions of syntax nodes to evaluate across a large solution. You should limit the performance impact of running the analyzer by reducing the number of operations performed.
+
+**Do**
+* Use the [Syntax Visualizer](https://docs.microsoft.com/en-us/dotnet/csharp/roslyn-sdk/syntax-visualizer) to build syntax. Where possible use specific syntax options. As an example, if you are building an analyzer that will examine class inheritance then you should use `SyntaxKind.BaseList`, which will occur much less often than the `SyntaxKind.IdentifierName`.
+* Use information already available from `SyntaxNodeAnalysisContext` to quickly filter relevant information. As an example, if you build an analyzer that looks for a method then you would be evaluating `MemberAccessExpressionSyntax` nodes. In this scenario, you should also check the parent of the node to figure out if this Syntax is a method or a property.
+
+**Do Not**
+* Do not put string compare operations above other conditional checks that can be performed. String comparison is slower than checking for a type or examining an enumeration value.
```
twsouthwick, line 63: I think the important thing is not so much string comparison, but calling `.ToString()` and comparing that output. Comparing the ValueText of an identifier is probably cheaper than ISymbol.ToString

twsouthwick: Oh this may be blurring the next issue it seems

docs/roslyn_best_practices.md
```diff
@@ -0,0 +1,189 @@
...
+
+
+### 2. Beware of String manipulations
+Build your analyzer with the expectation that it will be invoked a million times. Look for, and replace, strings to prevent excessive garbage collection due to frequent executions of your analyzer.
+
+**Do**
+* Use string constants when evaluating string conditionals.
```
twsouthwick, line 70: Maybe a reformulation of the do not, but "cache string interpolation, concatenation, or format strings" could be good guidance

twsouthwick (approved):
This is a great starting point and should help as people start jumping in. Thanks!

docs/design/Pre-UpgradeAnalysis.md
```diff
@@ -77,9 +77,9 @@ Detailed instructions on how to set up the decentralized solution will be provid
 
 ## Phase III:
 Add loose assembly analysis capability to upgrade-assistant.
-- Chem integration.
+- Loose assembly analysis integration.
 
-**Cost** : L (This is L coz this requires quite a bit of experimental prototyping to know the best way to integrate Chem. This effort can happen in parallel to Phase I and Phase II)
+**Cost** : L (This is L coz this requires quite a bit of experimental prototyping to know the best way to integrate loose assembly analysis. This effort can happen in parallel to Phase I and Phase II)
```
CyrusNajmabadi, line 82: 'coz'? :)

docs/roslyn_best_practices.md
```diff
@@ -0,0 +1,184 @@
...
+
+### 1. Use the *Microsoft.CodeAnalysis.Testing* framework
+
+Separation of analyzer and code fix tests increases complexity and code duplication and tends to decrease the overall confidence in the test suite. If you're testing entire files at a time, you will either be quickly overwhelmed by the number of files per test scenario or be tempted to put multiple test scenarios into a single file which shifts from unit to integration testing. The *Microsoft.CodeAnalysis.Testing* framework addresses these concerns.
+
+**Do**
+* Read the testing overview: [Microsoft.CodeAnalysis.Testing](https://github.com/dotnet/roslyn-sdk/blob/main/src/Microsoft.CodeAnalysis.Testing/README.md)
+
+### 2. Avoid member variables and state management
+Expect your analyzer to be invoked repeatedly and asynchronously. Design your analyzer so that execution can start processing a 2nd call before processing finishes for the 1st call.
```
CyrusNajmabadi, line 39: we dont' invoke asynchronously.  we do invoke concurrently though.

docs/roslyn_best_practices.md
```diff
@@ -0,0 +1,184 @@
...
+* Read the testing overview: [Microsoft.CodeAnalysis.Testing](https://github.com/dotnet/roslyn-sdk/blob/main/src/Microsoft.CodeAnalysis.Testing/README.md)
+
+### 2. Avoid member variables and state management
+Expect your analyzer to be invoked repeatedly and asynchronously. Design your analyzer so that execution can start processing a 2nd call before processing finishes for the 1st call. 
+
+**Do**
+* Pass information between methods via method arguments.
+
+**Do Not**
+* Do not use member variables to store instance data.
```
CyrusNajmabadi, line 45: this probably needs more clarity.  For example, you shoudl not store instance data on your analyzer or fixer instance itself.  However, both of those types can (and often should) instantiate a new type when they're called into that will store state and perform logic as appropriate.  This way you can use standard classes and wahtnot to organize things.  For example:

`context.RegisterSyntaxNodeAction(new SyntaxAnalyzer().Analyze, ...)`

SyntaxAnalyzer can now be a rich class that can do these things.

docs/design/Pre-UpgradeAnalysis.md
KSchlobohm: rewording because it doesn't speak to the feature but rather an implementation detail.

PR #607, opened 2021-06-07
Add better support for zip extensions

twsouthwick:
- AssemblyLoadContext requires assemblies to be seekable, so we now will copy assemblies to byte arrays if they are not seekable
- We check for pattern that occurs when a zip file is created from a folder where the top level directory is a simple folder

This change also consolidates loading logic all into the assembly load context where it will attempt to load the assemblies from the manifest.

### DIFF
--- a/src/components/Microsoft.DotNet.UpgradeAssistant.Extensions/ExtensionInstance.cs
+++ b/src/components/Microsoft.DotNet.UpgradeAssistant.Extensions/ExtensionInstance.cs
@@ -2,21 +2,26 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System;
+using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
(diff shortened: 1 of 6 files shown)


src/components/Microsoft.DotNet.UpgradeAssistant.Extensions/ExtensionInstance.cs
```diff
@@ -74,7 +91,20 @@ public void Dispose()
                     {
                         var provider = new ZipFileProvider(e);
 
-                        return new ExtensionInstance(new ZipFileProvider(e), e);
+                        try
+                        {
+                            return new ExtensionInstance(provider, e);
+                        }
+
+                        // If the manifest file couldn't be found, let's try looking at one layer deep with the name
```
mjrousos, line 99: Good idea.

mjrousos (approved):
LGTM

PR #608, opened 2021-06-08
Make Package Analyzers TFM aware

sunandabalu:


src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/Dependencies/IDependencyAnalyzer.cs
```diff
@@ -15,8 +16,9 @@ public interface IDependencyAnalyzer
         /// the breaking change risk of this action.
         /// </summary>
         /// <param name="project">The project whose dependencies should be analyzed.</param>
+        /// <param name="targetframeworks">The targetframworkmonikers applicable for the project.</param>
         /// <param name="state">The current analysis state which will be updated.</param>
         /// <param name="token">The token used to gracefully cancel this request.</param>
-        Task AnalyzeAsync(IProject project, IDependencyAnalysisState state, CancellationToken token);
+        Task AnalyzeAsync(IProject project, IReadOnlyCollection<TargetFrameworkMoniker> targetframeworks, IDependencyAnalysisState state, CancellationToken token);
```
twsouthwick, line 22: Could this be a property on the `IDependencyAnalysisState` instead?

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/Dependencies/IDependencyAnalyzer.cs
```diff
@@ -15,8 +16,9 @@ public interface IDependencyAnalyzer
         /// the breaking change risk of this action.
         /// </summary>
         /// <param name="project">The project whose dependencies should be analyzed.</param>
+        /// <param name="targetframeworks">The targetframworkmonikers applicable for the project.</param>
```
twsouthwick, line 19: nit: spaces between targetframeworkmonikers or just TFMs

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/Dependencies/IDependencyAnalyzer.cs
twsouthwick: Does it have to be a collection? Could it just be IEnumerable?

sunandabalu: Added part of IDependencyAnalysisState

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/AnalyzePackageStatus.cs
```diff
@@ -35,9 +38,15 @@ public async Task AnalyzeAsync(AnalyzeContext analysis, CancellationToken token)
 
             foreach (var project in projects)
             {
+                var targetTfm = await _tfmSelector.SelectTargetFrameworkAsync(project, token).ConfigureAwait(false);
+                var targetframeworks = new List<TargetFrameworkMoniker>
```
twsouthwick, line 42: nit: can just use array

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/AnalyzePackageStatus.cs
```diff
@@ -35,9 +38,15 @@ public async Task AnalyzeAsync(AnalyzeContext analysis, CancellationToken token)
...
+                {
+                        targetTfm
+                };
+
                 try
                 {
-                    _analysisState = await _packageAnalyzer.AnalyzeAsync(context, project, token).ConfigureAwait(false);
+                    _analysisState = await _packageAnalyzer.AnalyzeAsync(context, project, targetframeworks.AsReadOnly(), token).ConfigureAwait(false);
```
twsouthwick, line 49: Shouldn't need the `AsReadOnly()`

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/AnalyzePackageStatus.cs
```diff
@@ -50,7 +59,7 @@ public async Task AnalyzeAsync(AnalyzeContext analysis, CancellationToken token)
                     Logger.LogCritical(exc, "Unexpected exception analyzing package references for: {ProjectPath}", context.CurrentProject.Required().FileInfo);
                 }
 
-                Logger.LogInformation("Package Analysis for {ProjectPath}", project.FileInfo.Name);
+                Logger.LogInformation("Package Analysis for {ProjectPath} for the target TFM of {TargetTFM}", new object[] { project.FileInfo.Name, targetTfm });
```
twsouthwick, line 62: You don't need to pass it as an array (ie there's a params overload)

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Analyzers/PackageMapReferenceAnalyzer.cs
```diff
@@ -101,14 +99,14 @@ private bool ContainsPackageReference(IEnumerable<NuGetReference> packages, stri
             return _comparer.Compare(version, reference.Version) <= 0;
         }
 
-        private async Task AddNetCoreReferences(NuGetPackageMap packageMap, IDependencyAnalysisState state, IProject project, CancellationToken token)
+        private async Task AddNetCoreReferences(NuGetPackageMap packageMap, IReadOnlyCollection<TargetFrameworkMoniker> targetframeworks, IDependencyAnalysisState state, IProject project, CancellationToken token)
```
twsouthwick, line 102: Isn't this now passed in the `state` object?

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageUpdaterStep.cs
```diff
@@ -155,7 +156,7 @@ protected override async Task<UpgradeStepApplyResult> ApplyImplAsync(IUpgradeCon
                         count++;
 
                         Logger.LogDebug("Re-running analysis to check whether additional changes are needed");
-                        _analysisState = await _packageAnalyzer.AnalyzeAsync(context, context.CurrentProject, token).ConfigureAwait(false);
+                        _analysisState = await _packageAnalyzer.AnalyzeAsync(context, project, project.TargetFrameworks, token).ConfigureAwait(false);
```
twsouthwick, line 159: Should this be passing in the selected TFM for the project?

sunandabalu: This is the same code that is being called for Pre-TFM step as well, so if we change this to use target TFM always then the package dependencies will be updated for target TFM before the TFM is updated, which is Ok I guess. We would then need to remove the Pre-TFM package update step probably.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/AnalyzePackageStatus.cs
sunandabalu: there isnt one - https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loggerextensions.loginformation?view=dotnet-plat-ext-5.0 - I can only see the object[] as parameter options under Microsoft.Extensions.Logging

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageUpdaterStep.cs
sunandabalu: That would probably be a breaking change here, coz this is applied during pre-TFM change too.

sunandabalu: FYI @mjrousos

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/AnalyzePackageStatus.cs
mjrousos: The one @twsouthwick is thinking of is an extension method: https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Logging.Abstractions/src/LoggerExtensions.cs#L172

This can become: `Logger.LogInformation("Package Analysis for {ProjectPath} for the target TFM of {TargetTFM}", project.FileInfo.Name, targetTfm);`

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/DependencyAnalysisState.cs
```diff
@@ -47,5 +49,7 @@ void SetRisk(BuildBreakRisk risk)
         public bool AreChangesRecommended => FrameworkReferences.HasChanges || Packages.HasChanges || References.HasChanges;
 
         public bool IsValid { get; set; }
+
+        public IReadOnlyCollection<TargetFrameworkMoniker> TargetFrameworks { get; set; }
```
mjrousos, line 53: This probably doesn't need a setter

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageUpdaterStep.cs
mjrousos: This is probably correct for now because it will preserve the tool's current behavior (we won't look for new Net5 versions of package the first time the step runs but will the second time (after updating the TFM)).

I suspect once we have a solution-wide package update step, we'll need to pass the new selected TFM in those cases but we can worry about that later.

mjrousos (approved):
LGTM; a couple minor suggestions inline, but looks good otherwise.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/AnalyzePackageStatus.cs
sunandabalu: changed

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageUpdaterStep.cs
sunandabalu: leaving it as is for now

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/DependencyAnalysisState.cs
sunandabalu: removed

PR #609, opened 2021-06-08
Make the VS interop lib private

twsouthwick:
This package is a build time component only and provides COM types that
are used to identify where VS is.


mjrousos (approved):
LGTM

KSchlobohm (approved):

PR #615, opened 2021-06-11
Simple implementation to expose Imports in IProjectFile

sunandabalu:

### DIFF
--- a/src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IProjectFile.cs
+++ b/src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IProjectFile.cs
@@ -25,6 +25,10 @@ namespace Microsoft.DotNet.UpgradeAssistant
 
         void RemoveReferences(IEnumerable<Reference> references);
 
+        void AddImports(string item);
+
+        void RemoveImports(string item);
+
         ValueTask SaveAsync(CancellationToken token);
 
         void Simplify();
@@ -41,6 +45,6 @@ namespace Microsoft.DotNet.UpgradeAssistant
 
         void SetTFM(TargetFrameworkMoniker targetTFM);
 
-        IEnumerable<string> Imports { get; }
+        ICollection<string> Imports { get; }
     }
 }
--- a/src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IProjectFile.cs
+++ b/src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IProjectFile.cs
@@ -25,10 +25,6 @@ namespace Microsoft.DotNet.UpgradeAssistant
 
         void RemoveReferences(IEnumerable<Reference> references);
 
-        void AddImports(string item);
-
-        void RemoveImports(string item);
-
         ValueTask SaveAsync(CancellationToken token);
 
         void Simplify();
--- a/src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/ImportsCollection.cs
+++ b/src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/ImportsCollection.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using Microsoft.Build.Construction;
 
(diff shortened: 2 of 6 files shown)


src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IProjectFile.cs
```diff
@@ -41,6 +45,6 @@ public interface IProjectFile
 
         void SetTFM(TargetFrameworkMoniker targetTFM);
 
-        IEnumerable<string> Imports { get; }
+        ICollection<string> Imports { get; }
```
mjrousos, line 48: This should stay IEnumerable. Otherwise, there appear to be two ways to add Imports (with the AddImports method or adding to the collection). We should limit it to the one recommended way (following the pattern of references would mean using Add* methods).

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.File.cs
```diff
@@ -56,7 +56,7 @@ public string Sdk
         public bool IsSdk =>
             ProjectRoot.Sdk is not null && ProjectRoot.Sdk.Contains(MSBuildConstants.DefaultSDK, StringComparison.OrdinalIgnoreCase);
 
-        public IEnumerable<string> Imports => ProjectRoot.Imports.Select(p => Path.GetFileName(p.Project));
+        public ICollection<string> Imports => ProjectRoot.Imports.Select(p => Path.GetFileName(p.Project)).ToList();
```
mjrousos, line 59: As mentioned above, this can probably stay an enumerable.

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.File.cs
```diff
@@ -194,5 +194,21 @@ public void SetPropertyValue(string propertyName, string propertyValue)
             Path.IsPathFullyQualified(path)
             ? path
             : Path.Combine(projectDir, path);
+
+        public void AddImports(string item)
```
mjrousos, line 198: This won't change the project file on disk. We need `AddImports` and `RemoveImports` to actually update the project file. `ProjectRoot.AddImport` should work for the add case, but I'm not sure if there's a good helper like that for removing. For that, we might need to find child nodes that match the import and use RemoveChild.

mjrousos (changes requested):

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IProjectFile.cs
sunandabalu: Removed the Add/Remove methods and added a custom collection to support the add/remove operation

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.File.cs
sunandabalu: Removed the Add/Remove methods and added a custom collection to support the add/remove operation

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.File.cs
sunandabalu: Changed to use extension method to removelement similar to what removereferences is doing.

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/ImportsCollection.cs
```diff
@@ -0,0 +1,85 @@
...
+                element.RemoveElement();
+                return true;
+            }
+
+            return false;
+        }
+
+        public void CopyTo(string[] array, int arrayIndex)
+        {
+            throw new NotImplementedException();
```
twsouthwick, line 72: For completeness could you implement this?

twsouthwick (approved):
Lgtm modulo the missing implementation

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/ImportsCollection.cs
sunandabalu: Added implementation

PR #623, opened 2021-06-14
Lazily load extensions when needed

twsouthwick:
This change moves the extension loading into a child scope. This allows for things like logging to be available while we are loading extensions. The main value of this change is it allows a bit better organization of some of the logic.


src/components/Microsoft.DotNet.UpgradeAssistant.Extensions/ExtensionManager.cs
```diff
@@ -0,0 +1,79 @@
...
+                return list;
+            });
+
+            ExtensionInstance? LoadExtension(string path)
+            {
+                foreach (var loader in loaders)
+                {
+                    try
+                    {
+                        if (loader.LoadExtension(path) is ExtensionInstance instance)
```
mjrousos, line 48: Not sure if this is worth changing, but rather than trying all the loaders, `IExtensionLoader` could have an API for determining whether a path matches their expected format (*.zip, manifest.json, etc.) and you could try loading the extension with only the loaders that indicate they can handle the path.

src/components/Microsoft.DotNet.UpgradeAssistant.Extensions/ExtensionProviderExtensions.cs
```diff
@@ -35,60 +36,63 @@ public static void AddExtensions(this IServiceCollection services, IConfiguratio
...
-                    sp.AddServices(new ExtensionServiceCollection(services, extension));
-                }
-            }
-        }
+                    foreach (var path in additionalExtensionPaths)
+                    {
+                        options.ExtensionPaths.Add(path);
+                    }
+                })
+                .AddDefaultExtensions(configuration)
```
mjrousos, line 47: The order extensions are added matters (if two extensions add template files with the same path, the later one wins) so we need to make sure we're consistently registering extensions in the same order. To match old behavior, I think the order needs to be default extensions loaded first, then from env vars, then from command line.

mjrousos (approved):
Looks good. Left a couple comments inline, but my only real concern is making sure that the order extensions are registered isn't changed (since later extensions can override earlier ones in some cases).

Also, wouldn't hurt to add some unit tests for some of the extension loading infrastructure.

src/components/Microsoft.DotNet.UpgradeAssistant.Extensions/ExtensionManager.cs
twsouthwick: Created an issue to track that change

PR #636, opened 2021-06-16
Add option to search for unlisted packages

twsouthwick:
This change exposes the options for searching in a more centralized way so that we can add more if we need to without breaking the API.

### DIFF
--- a/src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Analyzers/PackageMapReferenceAnalyzer.cs
+++ b/src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Analyzers/PackageMapReferenceAnalyzer.cs
@@ -59,7 +59,7 @@ namespace Microsoft.DotNet.UpgradeAssistant.Steps.Packages.Analyzers
                 {
                     _logger.LogInformation("Marking package {PackageName} for removal based on package mapping configuration {PackageMapSet}", packageReference.Name, map.PackageSetName);
                     state.Packages.Remove(packageReference, BuildBreakRisk.Medium);
-                    await AddNetCoreReferences(map, state, project, token).ConfigureAwait(false);
+                    await AddNetCoreReferences(map, state, token).ConfigureAwait(false);
                 }
             }
 
@@ -69,7 +69,7 @@ namespace Microsoft.DotNet.UpgradeAssistant.Steps.Packages.Analyzers
                 {
                     _logger.LogInformation("Marking assembly reference {ReferenceName} for removal based on package mapping configuration {PackageMapSet}", reference.Name, map.PackageSetName);
                     state.References.Remove(reference, BuildBreakRisk.Medium);
-                    await AddNetCoreReferences(map, state, project, token).ConfigureAwait(false);
+                    await AddNetCoreReferences(map, state, token).ConfigureAwait(false);
                 }
             }
         }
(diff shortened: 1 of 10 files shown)


twsouthwick:
@sunandabalu The builds don't seem to be showing up as checks for my PR. Any ideas?

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Analyzers/TargetCompatibilityReferenceAnalyzer.cs
```diff
@@ -51,7 +51,7 @@ public async Task AnalyzeAsync(IProject project, IDependencyAnalysisState state,
                 else
                 {
                     // If the package won't work on the target Framework, check newer versions of the package
-                    var newerVersions = await _packageLoader.GetNewerVersionsAsync(packageReference, state.TargetFrameworks, true, token).ConfigureAwait(false);
+                    var newerVersions = await _packageLoader.GetNewerVersionsAsync(packageReference, state.TargetFrameworks, new() { Prerelease = true }, token).ConfigureAwait(false);
```
mjrousos, line 54: I think the `true` that was passed here previously corresponded to `latestMinorAndBuildOnly`, not `prerelease`.

mjrousos (changes requested):
Good change. Only concern is that I've think you've set the wrong option property in TargetCompatibilityReferenceAnalyzer.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Analyzers/TargetCompatibilityReferenceAnalyzer.cs
sunandabalu: Mike is right, it was for latestMinorAndBuildOnly

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Analyzers/PackageMapReferenceAnalyzer.cs
```diff
@@ -99,14 +99,14 @@ private bool ContainsPackageReference(IEnumerable<NuGetReference> packages, stri
...
-        private async Task AddNetCoreReferences(NuGetPackageMap packageMap, IDependencyAnalysisState state, IProject project, CancellationToken token)
+        private async Task AddNetCoreReferences(NuGetPackageMap packageMap, IDependencyAnalysisState state, CancellationToken token)
         {
             foreach (var newPackage in packageMap.NetCorePackages)
             {
                 var packageToAdd = newPackage;
                 if (packageToAdd.HasWildcardVersion)
                 {
-                    var reference = await _packageLoader.GetLatestVersionAsync(packageToAdd.Name, state.TargetFrameworks, false, token).ConfigureAwait(false);
+                    var reference = await _packageLoader.GetLatestVersionAsync(packageToAdd.Name, state.TargetFrameworks, new(), token).ConfigureAwait(false);
```
sunandabalu, line 109: no need to set Prerelease false to the options object here? will it default to false?

sunandabalu:
> @sunandabalu The builds don't seem to be showing up as checks for my PR. Any ideas?

Looks like there is an AzDO outage happening, they are working on mitigating it.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Analyzers/PackageMapReferenceAnalyzer.cs
twsouthwick: yes, that's the default for bool values

mjrousos (approved):
LGTM

PR #637, opened 2021-06-16
Add support for a minimum required version in extension

twsouthwick:
This change allows an extension to say what it's minimum required version for Upgrade Assistant is.

### DIFF
--- a/src/components/Microsoft.DotNet.UpgradeAssistant.Extensions/ExtensionManager.cs
+++ b/src/components/Microsoft.DotNet.UpgradeAssistant.Extensions/ExtensionManager.cs
@@ -26,7 +26,14 @@ namespace Microsoft.DotNet.UpgradeAssistant.Extensions
                 {
                     if (LoadExtension(path) is ExtensionInstance extension)
                     {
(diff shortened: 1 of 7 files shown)


src/components/Microsoft.DotNet.UpgradeAssistant.Extensions/ExtensionManager.cs
```diff
@@ -26,7 +26,14 @@ internal sealed class ExtensionManager : IDisposable, IEnumerable<ExtensionInsta
                 {
                     if (LoadExtension(path) is ExtensionInstance extension)
                     {
-                        list.Add(extension);
+                        if (extension.MinUpgradeAssistantVersion is Version minVersion && minVersion < options.Value.CurrentVersion)
+                        {
+                            logger.LogWarning("Could not load extension from {Path}", path);
```
mjrousos, line 31: Would be good to call out in the warning that it couldn't be loaded because UA's version was too low and what the min version is.

mjrousos (approved):
LGTM

PR #641, opened 2021-06-17
Add analyzer/code fix for updating or removing old attribute types

### LINKED ISSUE #493: Replace System.Web.Mvc.AuthorizeAttribute with Microsoft.AspNetCore.Authorization.AuthorizeAttribute
This should be a simple translation to make in a code fix that will help ASP.NET controllers upgrade to be ready for ASP.NET Core.

### LINKED ISSUE #499: Replace System.Web.Mvc.OutputCacheAttribute with Microsoft.AspNetCore.Mvc.ResponseCacheAttribute

### LINKED ISSUE #501: Replace System.Web.Mvc.BindAttribute with Microsoft.AspNetCore.Mvc.BindAttribute

### LINKED ISSUE #504: Remove System.Web.Mvc.ValidateInputAttribute
This attribute is no longer needed

### LINKED ISSUE #638: TypeUpgradeAnalyzer shouldn't create diagnostics for attributes
Currently, this the TypeUpgradeAnalyzer and code fix provider translate this code:

```CSharp
[ApiController]
public class MyClass {
```

Into:

```CSharp
[ControllerBase]
public class MyClass {
```

The analyzer thinks this is right because ApiController -> ControllerBase but doesn't recognize that in this context [ApiController] is really ApiControllerAttribute. TypeUpgradeAnalyzer should be updated to only replace attribute types if the *attribute* types are mapped in type mappings (e.g. ApiController*Attribute*). Alternatively, the type upgrade analyzer can completely ignore attributes and allow a more specialized analyzer to handle them.

mjrousos:
This replaces the AllowHtmlAttribute analyzer and code fixer with a more general purpose analzyers/code fix provider that will update or remove any attribute types listed in typemaps.

Fixes #493 
Fixes #499 
Fixes #501
Fixes #504 
Fixes #638


src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/AttributeUpgradeAnalyzer.cs
```diff
@@ -0,0 +1,127 @@
...
+            // Store the new attribute type that this identifier should be replaced with (or null to remove the attribute)
+            // for use by the code fix provider.
+            var properties = ImmutableDictionary.Create<string, string?>().Add(NewTypeKey, mapping.NewName);
+
+            // Create and report the diagnostic
+            var diagnostic = Diagnostic.Create(Rule, context.Node.GetLocation(), properties, mapping.OldName, mapping.NewName);
+            context.ReportDiagnostic(diagnostic);
+        }
+
+        private static bool MatchesPartiallyQualifiedName(string fullyQualifiedName, string partiallyQualifiedName)
```
twsouthwick, line 106: This would be a great addition for the NameMatcher class

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CodeFixes/AttributeUpgradeCodeFixer.cs
```diff
@@ -0,0 +1,153 @@
...
+            var qualifier = namespaceDelimiterIndex >= 0
+                ? attributeType.Substring(0, namespaceDelimiterIndex)
+                : null;
+
+            // The simplifier does not recognize using statements within namespaces, so this
+            // checks whether the necessary using statement is present or not and adds the
+            // AddImportsAnnotation only if it's absent.
+            if (qualifier is not null && !node.HasAccessToNamespace(qualifier))
+            {
+                updatedNode = updatedNode.WithAdditionalAnnotations(Simplifier.AddImportsAnnotation);
```
twsouthwick, line 130: This is a new API I believe. We had been targeting support in VS 16.8 (which supports Roslyn 3.8) due to the requirement for .NET 5. However, looks like we've updated to 3.9. Was this intentional? We should update a requirements then.

twsouthwick (changes requested):
Blocking to ensure we discuss the potential versioning issue

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CodeFixes/AttributeUpgradeCodeFixer.cs
mjrousos: I just checked and this is in 3.8.0. Good point about not wanting to be on a newer version, though. I opened a separate PR to roll that back (#653).

KSchlobohm:
changelog++

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/AttributeUpgradeAnalyzer.cs
mjrousos: Good point. I've moved it there.

mjrousos:
Feedback has been addressed and I've fixed up the way we add namespaces to work now, so unit tests pass.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/SyntaxNodeExtensions.cs
```diff
@@ -173,44 +197,74 @@ public static bool HasUsingStatement(this SyntaxNode node, string namespaceName)
                 .Concat(children.OfType<VBSyntax.SimpleImportsClauseSyntax>().Select(i => i.Name.ToString()))
                 .Concat(children.OfType<VBSyntax.ImportsStatementSyntax>().SelectMany(i => i.ImportsClauses.OfType<VBSyntax.SimpleImportsClauseSyntax>().Select(i => i.Name.ToString())));
 
-            return usings.Any(n => n.Equals(namespaceName, StringComparison.Ordinal));
+            return usings.Any(n => n.Equals(namespaceName, node.Language == LanguageNames.CSharp ? StringComparison.Ordinal : StringComparison.OrdinalIgnoreCase));
```
KSchlobohm, line 200: nice, always appreciate when someone reminds me to be case insensitive for VB

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/SyntaxNodeExtensions.cs
```diff
@@ -173,44 +197,74 @@ public static bool HasUsingStatement(this SyntaxNode node, string namespaceName)
...
+        /// Determines if a document root element contains a using/import statement for a given namespace.
         /// </summary>
-        /// <param name="documentRoot">The document to add the directive to.</param>
-        /// <param name="namespaceName">The namespace to reference with the using directive.</param>
-        /// <returns>An updated document root with the specific using directive.</returns>
-        public static CSSyntax.CompilationUnitSyntax AddUsingIfMissing(this CSSyntax.CompilationUnitSyntax documentRoot, string namespaceName)
+        /// <param name="documentRoot">The document root to analyze.</param>
+        /// <param name="namespaceName">The namespace name to look for an import for.</param>
+        /// <returns>True if the documentRoot is a root node and contains a top-level using/import statement for the specified namespace name.</returns>
+        private static bool RootIncludesImport(this SyntaxNode documentRoot, string namespaceName)
```
KSchlobohm, line 253: nit: this process of determining if a class includes an import seems to be its own unit of work. seeing there are a few private helper methods has me wanting a new class to isolate these helpers as part of a single unit of work.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/NameMatcher.cs
```diff
@@ -19,10 +19,37 @@ private NameMatcher(string typeName, string? memberName = null)
             _memberName = memberName;
         }
 
+        public static NameMatcher MatchType(string typeName) => new NameMatcher(typeName);
+
         public static NameMatcher MatchPropertyAccess(string typeName, string memberName) => new(typeName, memberName);
 
         public string TypeName => _typeName[_typeName.Length - 1];
 
+        public bool MatchesPartiallyQualifiedType(string partiallyQualifiedTypeName)
```
KSchlobohm, line 28: this is invoked from the analyzer for each evaluation. Is there a way we can avoid doing a string.Split('.') on every evaluation?

KSchlobohm (changes requested):
question about nameMatcher method

Directory.Build.targets
```diff
@@ -9,7 +9,7 @@
     </PackageReference>
   </ItemGroup>
 
-  <ItemGroup>
+  <ItemGroup Condition="'$(Language)'=='C#'">
```
twsouthwick, line 12: :)

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/SyntaxNodeExtensions.cs
twsouthwick: I believe we have an extension method to get a comparison based on the node....

twsouthwick (approved):

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/SyntaxNodeExtensions.cs
mjrousos: It looks like we have an extension for symbols but not syntax. I'll add one and use it.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/NameMatcher.cs
mjrousos: Actually, I've changed my mind. I'll just add the caching now. It's unclear to me whether it will matter, but it's easy enough to add.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/SyntaxNodeExtensions.cs
mjrousos: I refactored this a bit to try keep it more condensed. If it grows any more, I'll look into moving it into its own class.

PR #642, opened 2021-06-17
Expose SDK as an ICollection to enable extention

sunandabalu:


src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.File.cs
```diff
@@ -38,26 +38,13 @@ public ProjectRootElement ProjectRoot
...
-                if (sdk is null)
-                {
-                    throw new ArgumentOutOfRangeException("Should check IsSdk property first");
-                }
-
-                return sdk;
-            }
-        }
-
         public bool IsSdk =>
```
twsouthwick, line 41: I don't know if this is necessary anymore. Shouldn't it be equivalent to `Sdk.Any()`?

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.File.cs
```diff
@@ -38,26 +38,13 @@ public ProjectRootElement ProjectRoot
...
-            ProjectRoot.Sdk is not null && ProjectRoot.Sdk.Contains(MSBuildConstants.DefaultSDK, StringComparison.OrdinalIgnoreCase);
+            Sdk.Contains(MSBuildConstants.DefaultSDK) || Sdk.Contains(MSBuildConstants.WebSdk);
 
         public ICollection<string> Imports => new ImportsCollection(ProjectRoot);
 
+        public ICollection<string> Sdk => ProjectRoot.Sdk.Split(';').ToList();
```
twsouthwick, line 46: You should implement the ICollection against ProjectRoot so you can add/remove stuff

src/extensions/web/Microsoft.DotNet.UpgradeAssistant.Extensions.Web/WebSdkCleanupAnalyzer.cs
```diff
@@ -37,7 +36,7 @@ public Task AnalyzeAsync(IProject project, IDependencyAnalysisState state, Cance
             // Check SDK directly (instead of using project.Components) since having the FrameworkReference counts as
             // having the AspNetCore component and this analyzer is specifically insterested in cases where both the SDK
             // and the framework reference are present.
-            if (!projectRoot.IsSdk || !projectRoot.Sdk.Equals(WebSdk, StringComparison.OrdinalIgnoreCase))
+            if (projectRoot.Sdk != null && !projectRoot.Sdk.Contains(WebSdk))
```
twsouthwick, line 39: Sdk should never be null

twsouthwick (changes requested):

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.File.cs
sunandabalu: Changed to sdk.Any().

twsouthwick: Can we remove it? A caller could call `Sdk.Any()`

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.File.cs
```diff
@@ -194,5 +181,11 @@ public void SetPropertyValue(string propertyName, string propertyValue)
             Path.IsPathFullyQualified(path)
             ? path
             : Path.Combine(projectDir, path);
+
+        public void SetSDK(string sdk)
```
twsouthwick, line 185: Do we need this?

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/SdkCollection.cs
```diff
@@ -0,0 +1,85 @@
...
+    public class SdkCollection : ICollection<string>
+    {
+        private readonly ProjectRootElement _projectRoot;
+
+        public SdkCollection(ProjectRootElement projectRoot)
+        {
+            _projectRoot = projectRoot;
+        }
+
+        public HashSet<string> GetSdks() => _projectRoot.Sdk.Split(';').ToHashSet(StringComparer.OrdinalIgnoreCase);
```
twsouthwick, line 22: this can be private

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/SdkCollection.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        public ICollection<string> GetSdkLists() => _projectRoot.Sdk.Split(';').ToList();
```
twsouthwick, line 24: this can be private and List<string>

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/SdkCollection.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        public bool Contains(string item)
+        {
+            return GetSdks().Contains(item, StringComparer.OrdinalIgnoreCase);
```
twsouthwick, line 28: Why the comparer again? The GetSdks is returning hashset that has the right comparer. You're actually using a LINQ API that will iterate over everything

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/SdkCollection.cs
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        public bool IsReadOnly
+        {
+            get { return false; }
+        }
+
+        public bool Remove(string item)
+        {
+            if (Contains(item))
```
twsouthwick, line 59: This is going to create the hashset and then a list. Can it be simplified?

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/SdkCollection.cs
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        public void CopyTo(string[] array, int arrayIndex)
+        {
+            ((ICollection<string>)this).CopyTo(array, arrayIndex);
+        }
+
+        public IEnumerator<string> GetEnumerator()
+        {
+            return _projectRoot.Sdk.Split(';').ToList().GetEnumerator();
```
twsouthwick, line 77: You can return the GetSdkList().GetEnumerator here

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/SdkCollection.cs
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        IEnumerator IEnumerable.GetEnumerator()
+        {
+            return _projectRoot.Sdk.Split(';').ToArray().GetEnumerator();
```
twsouthwick, line 82: You can just do `IEnumerator IEnumerable.GetEnumerator() => GetEnumerator()` here to reduce duplicat8ion

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/SdkCollection.cs
twsouthwick: Also, why is it a list? Is there a problem with using the array returned here?

twsouthwick (changes requested):
Can you add some tests for the SdkCollection?

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/SdkCollection.cs
sunandabalu: I got a warning for declaring this as List<string> so had to make this a ICollection

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/SdkCollection.cs
sunandabalu: I can just use the Hashset to remove but it wont retain the order which might be Ok. But is there a better way of doing this without getting into string manipulation?

sunandabalu: Tried with LINQ

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.File.cs
sunandabalu: Removed

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/ComponentIdentifier.cs
```diff
@@ -53,7 +53,7 @@ private static ProjectComponents GetSDKProjectComponents(IProject project, IProj
                 components |= ProjectComponents.WindowsDesktop;
             }
 
-            if (file.Sdk.Equals(MSBuildConstants.DesktopSdk, StringComparison.OrdinalIgnoreCase))
+            if (file.Sdk.Contains(MSBuildConstants.DesktopSdk, StringComparer.OrdinalIgnoreCase))
```
twsouthwick, line 56: These stringcompareres are making it so it's using LINQ rather than you're actual contains method

sunandabalu: removed

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.File.cs
twsouthwick, line 41: This should be Sdk.Any() as this is a check if we're using Sdk style. Not sure why it was checking for specific things below.

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/SdkCollection.cs
```diff
@@ -0,0 +1,78 @@
...
+        private readonly ProjectRootElement _projectRoot;
+
+        public SdkCollection(ProjectRootElement projectRoot)
+        {
+            _projectRoot = projectRoot;
+        }
+
+        private HashSet<string> GetSdks() => _projectRoot.Sdk.Split(';').ToHashSet(StringComparer.OrdinalIgnoreCase);
+
+        private ICollection<string> GetSdkLists() => _projectRoot.Sdk.Split(';').ToList();
```
twsouthwick, line 24: Can't this just be the array? Why copy it to another list?

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/SdkCollection.cs
twsouthwick: This doesn't work - you're still creating a copy twice, but now it's the hashset twice. The hashset doesn't maintain the order (I'm assuming that matters). If you're going to us linq, do it on the GetSdksList

src/extensions/web/Microsoft.DotNet.UpgradeAssistant.Extensions.Web/WebSdkCleanupAnalyzer.cs
```diff
@@ -37,7 +36,7 @@ public Task AnalyzeAsync(IProject project, IDependencyAnalysisState state, Cance
             // Check SDK directly (instead of using project.Components) since having the FrameworkReference counts as
             // having the AspNetCore component and this analyzer is specifically insterested in cases where both the SDK
             // and the framework reference are present.
-            if (!projectRoot.IsSdk || !projectRoot.Sdk.Equals(WebSdk, StringComparison.OrdinalIgnoreCase))
+            if (!projectRoot.Sdk.Contains(WebSdk, StringComparer.OrdinalIgnoreCase))
```
twsouthwick, line 39: Again, you should remove all the stringcomparer args here - you handle that in the collection

twsouthwick (changes requested):
Can you add some tests for this?

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/SdkCollection.cs
```diff
@@ -0,0 +1,74 @@
...
+    public class SdkCollection : ICollection<string>
+    {
+        private readonly ProjectRootElement _projectRoot;
+
+        public SdkCollection(ProjectRootElement projectRoot)
+        {
+            _projectRoot = projectRoot;
+        }
+
+        private string[] GetSdks() => _projectRoot.Sdk.Split(";", StringSplitOptions.RemoveEmptyEntries);
```
twsouthwick, line 22: nit: Might be better to name it like `SplitSdk()` or something like that

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/SdkCollection.cs
```diff
@@ -0,0 +1,74 @@
...
+        public bool IsReadOnly
+        {
+            get { return false; }
+        }
+
+        public bool Remove(string item)
+        {
+            var sdkList = GetSdks().Where(p => p != item);
+            _projectRoot.Sdk = string.Join(";", sdkList);
+            return true;
```
twsouthwick, line 59: should only return false if the item wasn't removed

tests/components/Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests/SdkCollectionTests.cs
```diff
@@ -0,0 +1,71 @@
...
+    [Collection(MSBuildStepTestCollection.Name)]
+    public class SdkCollectionTests
+    {
+        private const string DefaultSDK = "Microsoft.NET.Sdk";
+
+        [InlineData(DefaultSDK)]
+        [Theory]
+        public void SdkCollectionContainsTest(string sdk)
+        {
+            using var mock = AutoMock.GetLoose();
```
twsouthwick, line 19: nit: include an //arrange/act/assert

tests/components/Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests/SdkCollectionTests.cs
```diff
@@ -0,0 +1,71 @@
...
+namespace Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests
+{
+    [Collection(MSBuildStepTestCollection.Name)]
+    public class SdkCollectionTests
+    {
+        private const string DefaultSDK = "Microsoft.NET.Sdk";
+
+        [InlineData(DefaultSDK)]
+        [Theory]
+        public void SdkCollectionContainsTest(string sdk)
```
twsouthwick, line 17: What about tests for empty strings, for removals, etc. What's the code coverage of this look like (for the SdkCollection type)?

sunandabalu: There is a test for remove, I'll add an entry for empty string.

twsouthwick: You may want to add a test for remove that doesn't have the entry - that should return false

sunandabalu: Added

tests/components/Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests/SdkCollectionTests.cs
```diff
@@ -0,0 +1,102 @@
...
+            // Arrange
+            using var mock = AutoMock.GetLoose();
+
+            var projectRoot = ProjectRootElement.Create();
+            projectRoot.Sdk = string.Concat(DefaultSDK, ";", WebSDK);
+
+            var sdkCollection = new SdkCollection(projectRoot);
+
+            // Act
+            sdkCollection.Remove(sdkToRemove);
```
twsouthwick, line 94: Should assert result here

twsouthwick (approved):
A little nit, but otherwise LGTM

tests/components/Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests/SdkCollectionTests.cs
sunandabalu: Added an assert here

PR #645, opened 2021-06-18
ExtensionManager should be a singleton

twsouthwick:
This was making it so that extensions were loaded twice. Only one of them was used, but the ExtensionManager should be a singleton to only load extensions once.


KSchlobohm (approved):
LGTM

PR #665, opened 2021-06-24
Organize host builder patterns

twsouthwick:
This separates out the commands to separate classes to clean up what is done in Program.cs


mjrousos (approved):
LGTM

PR #666, opened 2021-06-25
Move UpgradeOptions into CLI class

twsouthwick:
UpgradeOptions should be an implementation detail of the CLI as it is tightly bound to the command line parsing. The names on it cannot be changed without breaking that binding, and has ended up being a bag of a bunch of options.

This change introduces new options for the parts of the system that needed it and flows the values to them via:

- Directly configuring the options in the ConfigureServices
- Setting them up as extension options for values that are needed in extensions and cannot be directly set


src/components/Microsoft.DotNet.UpgradeAssistant.Extensions/ExtensionProviderExtensions.cs
```diff
@@ -96,6 +97,27 @@ public static OptionsBuilder<ExtensionOptions> AddDefaultExtensions(this Options
...
+        {
+            if (builder is null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            var json = JsonSerializer.SerializeToUtf8Bytes(option);
+            using var stream = new MemoryStream(json);
+
+            var config = new ConfigurationBuilder()
```
mjrousos, line 110: Should we also include configuration from appsettings.json since that's a logical place people might put default extension settings at some point?

mjrousos (approved):

src/components/Microsoft.DotNet.UpgradeAssistant.Extensions/ExtensionProviderExtensions.cs
twsouthwick: I don't think so - they should be putting anything like that into the ExtensionManfiest.json. This is to pass through command line args.

mjrousos: Ok, that makes sense. I'm still in the mindset of when we had a default extension mixed in with the tool (and appsettings.json served as the extensionmanifest.json). You're right that now the extensions all have their own manifests that should be used.

PR #668, opened 2021-06-25
Make SetProperty Remove an element if passed as Null or Empty

sunandabalu:

### DIFF
--- a/src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.File.cs
+++ b/src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.File.cs
@@ -174,7 +174,21 @@ namespace Microsoft.DotNet.UpgradeAssistant.MSBuild
             => Project.GetPropertyValue(propertyName);
 
         public void SetPropertyValue(string propertyName, string propertyValue)
-            => Project.SetProperty(propertyName, propertyValue);
+        {
+            var property = Project.GetProperty(propertyName);
+
+            if (property != null)
+            {
+                if (!string.IsNullOrEmpty(propertyValue))
+                {
+                    Project.SetProperty(propertyName, propertyValue);
+                }
+                else
+                {
+                    Project.RemoveProperty(property);
+                }
+            }
+        }
 
         private static string GetPathRelativeToProject(string path, string projectDir) =>
             Path.IsPathFullyQualified(path)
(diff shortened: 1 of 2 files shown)


src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.File.cs
```diff
@@ -174,7 +174,21 @@ public string GetPropertyValue(string propertyName)
             => Project.GetPropertyValue(propertyName);
 
         public void SetPropertyValue(string propertyName, string propertyValue)
-            => Project.SetProperty(propertyName, propertyValue);
+        {
+            var property = Project.GetProperty(propertyName);
+
+            if (property != null)
```
mjrousos, line 180: Would the old implementation have set a not-yet-existent property? I haven't tested it out, but it seems like that API might have allowed setting properties that didn't exist whereas this one will treat those as a no-op.

mjrousos (changes requested):
What does `Property.SetProperty` do if no property exists with the right name? I worry we might be changing the behavior in that case.

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.File.cs
twsouthwick: If `propertyValue is null` is the only time we should find the property. Otherwise, just remove it. It shouldn't be contingent on the property not existing, but on what the user passes in.

@mjrousos does that address it?

mjrousos: Right, if `propertyValue` is null (or maybe empty), then we expect `property` to exist. If `propertyValue` exists, though, I'm not sure `propety` needs to.

I'm thinking this should look more like:

```CSharp
if (string.IsNullOrEmpty(propertyValue))
{
    var property = Project.GetProperty(propertyName);
    if (property != null)
    {
        Project.RemoveProperty(property);
    }
}
else
{
    Project.SetProperty(propertyName, propertyValue);
}
```

sunandabalu: It will Set or add a property with the specified name and value. I'll remove the check on property existing and separate out the remove functionality to make it cleaner.

mjrousos (approved):
LGTM

PR #685, opened 2021-07-02
Add infrastructure for creating diagnostics for general API usage

### LINKED ISSUE #65: As a user, I am alerted if I use bundle/minify patterns that will need to change

### LINKED ISSUE #69: As a user, I am alerted if I have HttpModules or message handlers that will need to be replaced with middleware

### LINKED ISSUE #75: As a user, I'm alerted if I'm using profile/identity APIs that will need updated

### LINKED ISSUE #502: Flag usage of System.Web.Mvc.ChildActionOnlyAttribute
Recommend replacing with view components

### LINKED ISSUE #621: Create an analyzer to identify usage of APIs that UA can't upgrade
This will involve identifying namespaces, types, or members that users need to upgrade manually and displaying relevant information and links to documentation.

The analyzer should use config files so that users and extension authors can easily add their own APIs mapped to messages that the analyzer will use.

mjrousos:
This adds an analyzer (without code fix provider) for identifying and flagging usage of configured namespaces, types, or members and creating diagnostics with associated messages.

This will allow easily notifying users if they're using namespace, types, or members that are known to require manual changes.

Fixes #621 
Fixes #502 
Fixes #65 
Fixes #69 
Fixes #75


src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/TargetSyntaxMessageLoader.cs
```diff
@@ -0,0 +1,65 @@
...
+    public static class TargetSyntaxMessageLoader
+    {
+        /// <summary>
+        /// The suffix that api/message map files are expected to use.
+        /// </summary>
+        private const string TargetSyntaxMessageFileSuffix = ".apitargets";
+
+        private static JsonSerializerOptions GetJsonSerializationOptions()
+        {
+            var jsonSerializerOptions = new JsonSerializerOptions
```
twsouthwick, line 22: Is this available to the analyzer? Dependencies (such as System.Text.Json) are notoriously difficult in roslyn extensions (ie analyzers/source code generators/etc)

twsouthwick (changes requested):

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/SyntaxNodeExtensions.cs
```diff
@@ -278,6 +275,14 @@ private static bool RootIncludesImport(this SyntaxNode documentRoot, string name
             throw new NotImplementedException(Resources.UnknownLanguage);
         }
 
+        public static SyntaxNode? GetMAEExpressionSyntax(this SyntaxNode memberAccessSyntax) =>
```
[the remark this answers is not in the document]
mjrousos: The problem is that it gets the MemberAccessExpression's expression member. And GetMemberAccessExpressionSyntaxExpression or GetMemberAccessExpressionExpression are pretty long and confusing. :-/

Maybe I'll call it `GetChildExpressionSyntax` and mention that it only works on MAE syntaxes in doc comments?

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/ApiAlertAnalyzer.cs
```diff
@@ -0,0 +1,237 @@
...
+                    return targetSyntax.NameMatcher.Matches(symbol);
+                }
+            }
+
+            // If the node's full type can't be determined (either by symbol or fully qualified syntax),
+            // return true for the partial match only if ambiguous matching is enabled
+            return targetSyntax.AlertOnAmbiguousMatch;
+        }
+
+        private static string GetFullName(SyntaxNode node)
```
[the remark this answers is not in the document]
mjrousos: Good idea.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/DefaultApiAlerts.json
```diff
@@ -0,0 +1,120 @@
...
+        "SyntaxType": "Type",
+        "AlertOnAmbiguousMatch": true
+      },
+      {
+        "FullName": "System.Web.IHttpHandler",
+        "SyntaxType": "Type",
+        "AlertOnAmbiguousMatch": true
+      }
+    ],
+    "Message": "HTTP modules and HTTP handlers should be rewritten as middleware in ASP.NET Core. https://docs.microsoft.com/aspnet/core/migration/http-modulesIHttpModule",
```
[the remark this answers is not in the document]
mjrousos: There's a lot of localization work that's needed in UA. I'll leave this as-is for now, but create an issue to track localization.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/TargetSyntaxMessageLoader.cs
mjrousos: Yes, this is used to deserialize embedded resources and additional files in the project that define which APIs to create diagnostics for. I thought that I saw this working in VS, but I'm not certain, so I'll double-check.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/TargetSyntax.cs
```diff
@@ -0,0 +1,70 @@
...
+using System;
+
+namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers
+{
+    /// <summary>
+    /// Describes a syntax (namspace, type, or member) users might reference in their projects.
+    /// </summary>
+    public class TargetSyntax
+    {
+        private string _fullName = default!;
```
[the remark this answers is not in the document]
mjrousos: Good point; fixed.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/ApiAlertAnalyzer.cs
```diff
@@ -0,0 +1,237 @@
...
+        {
+            var fullName = node.GetQualifiedName().ToString();
+
+            // Ignore generic parameters for now to simplify matching
+            var genericParameterIndex = fullName.IndexOfAny(new[] { '<', '(' });
+            if (genericParameterIndex > 0)
+            {
+                fullName = fullName.Substring(0, genericParameterIndex);
```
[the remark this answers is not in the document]
mjrousos: This pretty quickly returns as a string, though, so we'd still have to allocate. We could have the method return an ROS instead of a string, but that gets to be pretty viral. I'll leave this as-is for now, but ping me when you're back if you think this should be changed and we can re-visit.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/NameMatcher.cs
```diff
@@ -25,11 +25,27 @@ private NameMatcher(string typeName, string? memberName = null)
...
+                throw new ArgumentNullException(nameof(memberName));
+            }
+
+            var idx = memberName.LastIndexOf('.');
+            if (idx < 0)
+            {
+                throw new ArgumentException("Member name must be fully qualified");
+            }
+
+            return new NameMatcher(memberName.Substring(0, idx), memberName.Substring(idx + 1));
```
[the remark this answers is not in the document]
mjrousos: We'd still have to allocate strings to pass to NameMatcher. We could look into making NameMatcher work with ROS instead of strings, but that would be a larger change that we should probably track separately.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/TargetSyntaxMessageLoader.cs
mjrousos: Quick update - I investigated this and as you suspected, the System.Text.Json dependency is problematic depending on how the analyzer is hosted. Basically the same problem as dotnet/roslyn#41785. Things work in UA but it's important that the analyzers work outside of UA, too, so I'll re-work this to not have the System.Text.Json dependency. Ug. :( Good catch, though.

mjrousos: Ok, I've reworked this to not use JSON deserialization and confirmed that the analyzers do, in fact, work in VS now.

mjrousos:
@twsouthwick, I've updated the PR based on your feedback (thanks!). Can you take another look? 

Also, I removed launchSettings.json which it looks like you added last week but I think is unnecessary since I don't want to have a git change (or, worse, conflict) every time I f5 UA against a different project locally. Let me know if there's a reason you prefer committing the file.

twsouthwick:
The launchsettings.json change was so it would get launched with "development" enviornment so it won't send telemetry on dev work. Happy to explore other options there

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/DefaultApiAlerts.apitargets
```diff
@@ -0,0 +1,33 @@
+﻿# Target syntax messages are defined as an ID and message (formatted as `ID: Message`) followed by APIs that the message applies to.
+# APIs are described by a three-tuple of API type, API name, and a boolean indicating whether the alert should be given in cases where
+# the API is only ambiguously matched.
+
+A: HTTP modules and HTTP handlers should be rewritten as middleware in ASP.NET Core. https://docs.microsoft.com/aspnet/core/migration/http-modulesIHttpModule
+Type, System.Web.IHttpModule, true
+Type, System.Web.IHttpHandler, true
+
+B: WCF server APIs are unsupported on .NET Core. Consider rewriting to use gRPC (https://docs.microsoft.com/dotnet/architecture/grpc-for-wcf-developers), ASP.NET Core, or CoreWCF (https://github.com/CoreWCF/CoreWCF) instead.
+Type, System.ServiceModel.ServiceHost, true
```
twsouthwick, line 10: comma delimited could break with generic types, right?

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/TargetSyntaxMessageLoader.cs
```diff
@@ -0,0 +1,124 @@
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// Load target syntax messages from a serialized string.
+        /// </summary>
+        /// <param name="serializedContents">A collection of target syntax messages stored as strings.</param>
+        /// <returns>An enumerable of TargetSyntaxMessages corresponding to those serialized in the serializedContents argument.</returns>
+        public static IEnumerable<TargetSyntaxMessage>? LoadMappings(string serializedContents)
```
twsouthwick, line 52: Can this just return an empty enumerable if it fails?

twsouthwick (approved):
A few comments, but otherwise LGTM

As for the localizability, I'm fine punting that, but we should identify any architectural choices to make sure we won't cause breaking changes to introduce it.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/TargetSyntaxMessageLoader.cs
mjrousos: It could, but I feel like throwing might be preferable here because otherwise the analyzer might silently just not do anything in the case of a malformed file. Throwing will make the failure more visible but, since it's in an analyzer, I think it still won't block anything.

twsouthwick: In that case, there should be an analyze to detect malformed files.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/DefaultApiAlerts.apitargets
mjrousos: Currently, when we get the name of the type in the analyzer, we're leaving off any generic parameters. It would probably be more correct to get the name of the open/unbound generic type to reduce ambiguity, but currently we just get the generic type's identifier so there won't be any commas. 

I can change this to semi-colon delimiting, though, so that it doesn't break if we change things in the future. Also, I'll add a comment about how generics are handled.

mjrousos: Actually, thinking about it more I'd like to just handle generics properly. I'll update the way we get the name for the generic to use "ildasm" style generics List`1, Dictionary`2, etc. since we don't know what generic parameter names make sense. Then commas will still be good delimiters. I'll include a sample and comment in the apitargets file.

mjrousos:
Yea, I see why you changed launchsettings.json. I'll change it back to how you had it. I don't like having that file be always modified (and never committed) on my dev machine, but I can't think of a better alternative.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/TargetSyntaxMessageLoader.cs
mjrousos: I'll have it return an empty enumerable for now and we can look into adding an analyzer for malformed apitargets files in the future if it seems like it would help.

PR #688, opened 2021-07-08
Separate command options for upgrade and analyze

twsouthwick:
This change abstracts the command line options so that additional commands don't have to rely on the same options object.


sunandabalu (approved):
LGTM

PR #696, opened 2021-07-09
Add validation diagram and analyzers

twsouthwick:
This change adds a validation diagram to help maintain the architectural decisions so far and codify it into the build process.

PR #697, opened 2021-07-09
Remove abstractions library from analyzers

twsouthwick:
Since we don't ship the abstractions with the analyzers, they will fail to load in situations where it's not available (ie VS, msbuild, etc). This links the only dependencies in since we already match based on shape rather than direct matching.

### DIFF
--- a/src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ApplicableComponentsAttribute.cs
+++ b/src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ApplicableComponentsAttribute.cs
@@ -9,7 +9,12 @@ namespace Microsoft.DotNet.UpgradeAssistant
     /// the project being upgraded contains certain components.
     /// </summary>
     [AttributeUsage(AttributeTargets.Class, AllowMultiple = false, Inherited = false)]
-    public sealed class ApplicableComponentsAttribute : Attribute
+#if ABSTRACTIONS
+    public
+#else
+    internal
+#endif
+        sealed class ApplicableComponentsAttribute : Attribute
     {
         /// <summary>
         /// Gets the components that a project must contain for the attributed code fix provider to apply.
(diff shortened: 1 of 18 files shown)


twsouthwick:
Once #696 is available, we can look at how to encode this into the architecture. I'm wondering if we could separate out the analyzers as not part of the extension. Instead, we could have a configuration such as

```json
{
  "AnalyzerId": [
"UA...."
],
"AnalyzerPackages":[
...
]
}
```

This would then pull in the analyzers dynamically based on the enabled diagnostics. I'll create a separate issue to track this.

mjrousos:
The only thing ApplicableComponentsAttribute is used for is to filter which code fix providers are shown as sub-steps in different scenarios (so that someone upgrading a console app doesn't have to see a dozen web-specific analyzer/code-fix steps).

It might be ok to just remove this altogether since, even though users with console/library projects would see a lot more source updater sub-steps, they would still all show as 'complete' as soon as they got to the source updater step.

Another approach would be to (optionally) use ProjectComponents when registering a code fix provider with DI. We'd have to make some sort of holder class that holds a code fix provider and any associated components. Then, the source updater could get those objects instead of code fix providers directly and choose which to use/show based on the project components. I'm not sure that filtering the list of source updater sub-steps is even important enough to do that work for, though.

My preference would be to remove ApplicableComponentsAttirbute for now and, in the future, if we want to not list all the code fix providers as source updater sub-steps in the future, consider accomplishing that differently (by registering a UA-specific holder class that associates code fix providers with project components).

mjrousos (changes requested):
The only thing ApplicableComponentsAttribute is used for is to filter which code fix providers are shown as sub-steps in different scenarios (so that someone upgrading a console app doesn't have to see a dozen web-specific analyzer/code-fix steps).

It might be ok to just remove this altogether since, even though users with console/library projects would see a lot more source updater sub-steps, they would still all show as 'complete' as soon as they got to the source updater step.

Another approach would be to (optionally) use ProjectComponents when registering a code fix provider with DI. We'd have to make some sort of holder class that holds a code fix provider and any associated components. Then, the source updater could get those objects instead of code fix providers directly and choose which to use/show based on the project components. I'm not sure that filtering the list of source updater sub-steps is even important enough to do that work for, though.

My preference would be to remove ApplicableComponentsAttirbute for now and, in the future, if we want to not list all the code fix providers as source updater sub-steps in the future, consider accomplishing that differently (by registering a UA-specific holder class that associates code fix providers with project components).

twsouthwick:
@mjrousos I ended up just removing the applicable components. We can explore other options later.

mjrousos:
Sounds good; thanks!

mjrousos (approved):

Issue #729, opened 2021-07-19
Analyze report fails validation


twsouthwick:
After running the analyze command on baby smash, it failed to validate correctly at https://sarifweb.azurewebsites.net/Validation. 

- [ ] `SARIF2005: runs[0].tool.driver: The tool 'Dependency Analysis' does not provide 'informationUri'. This property helps the developer responsible for addessing a result by providing a way to learn more about the tool.`
- [ ] `SARIF2005: runs[0].tool.driver: The tool 'Dependency Analysis' does not provide any of the version-related properties 'version', 'semanticVersion', 'dottedQuadFileVersion'. Providing version information enables the log file consumer to determine whether the file was produced by an up to date version, and to avoid accidentally comparing log files produced by different tool versions.`
- [ ] `SARIF1010: runs[0].results[0]: This result contains neither of the properties 'ruleId' or 'rule.id'. The SARIF specification ([§3.27.5](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317643)) requires at least one of these properties to be present.`

sunandabalu:
Closed via #730

PR #730, opened 2021-07-20
Fix missing items in the sarif report

sunandabalu:
Addresses #729

### DIFF
--- a/src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/AnalyzePackageStatus.cs
+++ b/src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/AnalyzePackageStatus.cs
@@ -17,7 +17,11 @@ namespace Microsoft.DotNet.UpgradeAssistant.Steps.Packages
 
         private ILogger Logger { get; }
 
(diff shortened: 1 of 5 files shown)


src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/AnalyzePackageStatus.cs
```diff
@@ -17,7 +17,11 @@ public class AnalyzePackageStatus : IAnalyzeResultProvider
 
         private ILogger Logger { get; }
 
-        public string AnalysisTypeName => "Dependency Analysis";
+        public string Id => "UA101";
```
twsouthwick, line 20: We may want to document rule scheme, but ok for now to get this through

twsouthwick (approved):

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/AnalyzePackageStatus.cs
sunandabalu: I will add some documentation on rule scheme

PR #734, opened 2021-07-20
Plumb extension version through for analysis reports

twsouthwick:
This change does the following:

- Create an internal abstractions library that is used for abstractions that we don't want to ship publicly
- Move the extension management interfaces there
- Add a method to retrieve an extension for a service
- Add call in analyze command to get version of analyze providers


src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Commands/Analyze/ConsoleAnalyze.cs
```diff
@@ -40,11 +44,12 @@ public async Task RunAsync(CancellationToken token)
             await _stateManager.LoadStateAsync(context, token);
             var analzyerContext = new AnalyzeContext(context);
             var analyzeResultMap = new List<AnalyzeResultDefinition>();
+
             foreach (var provider in _providers)
             {
                 analyzeResultMap.Add(new()
                 {
-                    Version = UpgradeVersion.Current.FullVersion,
+                    Version = GetProviderVersion(provider),
```
sunandabalu, line 52: I think the sarif log expects a value for the version, could you add a default value in case the GetProviderVersion returns empty?

sunandabalu (changes requested):

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Commands/Analyze/ConsoleAnalyze.cs
twsouthwick: Good call. Defaulting to 0.0.0 in that case

sunandabalu (approved):

PR #741, opened 2021-07-23
Update logging in TFM selector filters to be more informative

### LINKED ISSUE #718: The target framework selector should log more details about why a particular TFM was chosen
As pointed out in #692, it's not always clear why Upgrade Assistant prefers .NET Standard, .NET 5, etc. for different projects. It's expected that different TFMs are used in different scenarios, but the target framework selector should log more messages (at informational level) explaining why a given TFM was chosen.

mjrousos:
Fixes #718


sunandabalu (approved):
LGTM

PR #742, opened 2021-07-26
Move from XliffTasks to Microsoft.DotNet.XliffTasks

premun:
The package name was changed to start with a reserved name (https://github.com/dotnet/xliff-tasks/issues/412)


sunandabalu (approved):
Thanks @premun  for making this change :)

PR #744, opened 2021-07-26
Add initial implementation of IExtensionManager APIs

### LINKED ISSUE #710: As a user, I can remove extensions

twsouthwick:
This change adds a config file for Upgrade Assistant that can hold items
that are expected to be persisted, starting with extensions. The file created will look like this:

```
upgrade-assistant.json:

{
  "Extensions": [
    {
      "Name": "test",
      "Source": "https://api.nuget.org/v3/index.json",
      "Version": null
    }
  ]
}
```

The values of `source`/`version`  will be validated/etc in a subsequent PR

Fixes #710

### DIFF
--- a/src/components/Microsoft.DotNet.UpgradeAssistant.Extensions/ExtensionOptions.cs
+++ b/src/components/Microsoft.DotNet.UpgradeAssistant.Extensions/ExtensionOptions.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 
 namespace Microsoft.DotNet.UpgradeAssistant.Extensions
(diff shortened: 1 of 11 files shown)


src/components/Microsoft.DotNet.UpgradeAssistant.Extensions/ExtensionOptions.cs
```diff
@@ -18,5 +19,7 @@ public class ExtensionOptions
         public ICollection<ExtensionInstance> Extensions { get; } = new List<ExtensionInstance>();
 
         public Version CurrentVersion { get; set; } = null!;
+
+        public string ConfigurationFilePath { get; } = Path.Combine(Environment.CurrentDirectory, "upgrade-assistant.json");
```
sunandabalu, line 23: Would there be a need to add this as a user option?

sunandabalu (approved):

src/components/Microsoft.DotNet.UpgradeAssistant.Extensions/ExtensionOptions.cs
twsouthwick: I don't see any scenario that would make sense. We may want to pivot it off of the directory of the input file, though

PR #749, opened 2021-07-27
Reference test packages in Directory.Build.targets

twsouthwick:
This change moves all PackageReferences for test utilities to the Directory.Build.targets. This will ensure all tests have the same mocking/fixture/etc library.


sunandabalu (approved):
Nice!

PR #753, opened 2021-07-27
Add commandline to chose VS version

### LINKED ISSUE #733: Allow selection of VS version
## Summary

A user should be able to select the version of VS they want UA to use for loading project files.

## Motivation and goals

VS can be installed side by side and multiple versions can exist on the same machine at a time. Since UA will use VS locations to locate target files, this can pose problems when only one version has a required workload.

## In scope

- Pass in VS path via command line
- Provide selection step for VS version if not available and multiple exist

## Out of scope

- None known

## Risks / unknowns

- Many users appreciate the non-interactive mode. There should be a way to continue using UA in non-interactive mode with multiple VS instances available

## Examples

`upgrade-assistant upgrade [path] --vs-path [path]`

twsouthwick:
This adds a commandline `--vs-path` to pass a VS instance through, as well as throwing a useful exception when a target is not available.

Fixes #733


src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Commands/Upgrade/ConsoleUpgradeCommand.cs
```diff
@@ -6,6 +6,8 @@
 using System.CommandLine.Invocation;
 using System.CommandLine.Parsing;
 using System.Threading;
+using Microsoft.DotNet.UpgradeAssistant.MSBuild;
+using Microsoft.Extensions.DependencyInjection;
```
sunandabalu, line 10: Are these imports needed? I don't see a corresponding change for usage

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Commands/UpgradeAssistantCommand.cs
```diff
@@ -44,6 +45,8 @@ protected class CommandOptions : IUpgradeAssistantOptions
             public UpgradeTarget TargetTfmSupport { get; set; } = UpgradeTarget.Current;
 
             public IEnumerable<AdditionalOption> AdditionalOptions => Option.ParseOptions();
+
+            public DirectoryInfo? VsPath { get; set; }
```
sunandabalu, line 49: This should be VSPath

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/IUpgradeAssistantOptions.cs
```diff
@@ -20,5 +20,7 @@ public interface IUpgradeAssistantOptions
         IReadOnlyCollection<string> Extension { get; }
 
         IEnumerable<AdditionalOption> AdditionalOptions { get; }
+
+        DirectoryInfo? VsPath { get; set; }
```
sunandabalu, line 24: Same, should be VSPath

sunandabalu:
Is there a way we can throw info about what components are needed if it fails to load the project?

twsouthwick:
Not at this time. It's a bootstrapping issue - in order to identify components, we need to load the project. In order to load the project, we need to have the workloads available.

sunandabalu (approved):

PR #756, opened 2021-07-28
Add implementation of extension creation/updating/adding/restoring

twsouthwick:
This change adds a NuGet backed way of creating/adding/restoring extensions.

NOTE: With this change, you can add/remove/list/use extensions. However, in order to restore on demand things that are currently synchronous need to be made async. That will be done in a subsequent change to minimize the amount of churn.


src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Commands/ExtensionManagement/ExtensionManagementCommand.cs
```diff
@@ -64,7 +88,7 @@ public AddExtensionCommand()
             {
                 AddHandler<AddExtensionAppCommand>();
                 AddArgument(new Argument<string>("name", LocalizedStrings.ExtensionManagementName));
-                AddOption(new Option<string>(new[] { "--source" }, () => DefaultSource, LocalizedStrings.ExtensionManagementSource));
+                AddOption(new Option<string>(new[] { "--source" }, LocalizedStrings.ExtensionManagementSource));
```
sunandabalu, line 91: Default source is not needed anymore?

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions.Internal/Extensions/IExtensionInstance.cs
```diff
@@ -20,6 +21,10 @@ public interface IExtensionInstance
 
         string Name { get; }
 
+        string Description { get; }
+
+        IReadOnlyCollection<string> Authors { get; }
+
```
sunandabalu, line 27: 👍

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions.Internal/IUpgradeAssistantConfigurationLoader.cs
```diff
@@ -1,15 +1,12 @@
...
-
 namespace Microsoft.DotNet.UpgradeAssistant
 {
     public interface IUpgradeAssistantConfigurationLoader
     {
-        Task SaveAsync(UpgradeAssistantConfiguration configuration, CancellationToken token);
+        UpgradeAssistantConfiguration Load();
 
-        Task<UpgradeAssistantConfiguration> LoadAsync(CancellationToken token);
+        void Save(UpgradeAssistantConfiguration configuration);
```
sunandabalu, line 10: Is this one of the instances that will be made async in the next iteration? Is there any perf hits in making this to be sync now?

src/components/Microsoft.DotNet.UpgradeAssistant.Extensions/ExtensionManager.cs
```diff
@@ -55,6 +65,17 @@ public sealed class ExtensionManager : IDisposable, IExtensionManager
                     LoadPath(path, isDefault: true);
                 }
 
+                try
+                {
+                    foreach (var path in Registered.Select(_extensionDownloader.GetInstallPath))
```
sunandabalu, line 70: Probably a good idea to move the `Registered.Select(_extensionDownloader.GetInstallPath)` out of the forloop and be a variable so it doesnot need to resolve it in every loop

src/components/Microsoft.DotNet.UpgradeAssistant.Extensions/ExtensionManager.cs
```diff
@@ -55,6 +65,17 @@ public sealed class ExtensionManager : IDisposable, IExtensionManager
...
+                    {
+                        LoadPath(path, isDefault: false);
+                    }
+                }
+                catch (OperationCanceledException)
+                {
+                }
+
```
sunandabalu, line 78: Why an empty catch? can we log something here?

src/components/Microsoft.DotNet.UpgradeAssistant.Extensions/ExtensionManager.cs
```diff
@@ -143,30 +135,59 @@ private static ExtensionInstance LoadOptionsExtension(IEnumerable<AdditionalOpti
...
+        public Task<bool> RemoveAsync(string name, CancellationToken token)
         {
-            var config = await _configurationLoader.LoadAsync(token).ConfigureAwait(false);
+            var config = _configurationLoader.Load();
             var existing = config.Extensions.FirstOrDefault(e => string.Equals(e.Name, name, StringComparison.OrdinalIgnoreCase));
 
             if (existing is null)
             {
-                return false;
+                return Task.FromResult(false);
```
sunandabalu, line 145: I only see sync method calls in here , why is this needed?

sunandabalu: Also why is the Remove an async if none of the method calls are async ?

src/components/Microsoft.DotNet.UpgradeAssistant.Extensions/ExtensionOptions.cs
```diff
@@ -20,6 +22,10 @@ public class ExtensionOptions
 
         public Version CurrentVersion { get; set; } = null!;
 
+        public bool LoadExtensions { get; set; } = true;
+
         public string ConfigurationFilePath { get; } = Path.Combine(Environment.CurrentDirectory, "upgrade-assistant.json");
+
+        public string ExtensionCachePath { get; } = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), "dotnet", "upgrade-assisistant", "extensions");
```
sunandabalu, line 29: What is a special folder?

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/PackageLoader.cs
```diff
@@ -240,53 +251,60 @@ private async Task<IEnumerable<NuGetFramework>> GetTargetFrameworksAsync(Package
...
+        private async Task<Stream> GetPackageStream(NuGetReference packageReference, string source, CancellationToken token)
+        {
+            var ms = new MemoryStream();
+            var packageSource = new[] { new PackageSource(source) };
+
+            await DownloadPackageToStreamAsync(packageReference, ms, packageSource, token).ConfigureAwait(false);
+
+            ms.Position = 0;
+
+            return ms;
```
sunandabalu, line 283: Where does this the memorystream object get disposed?

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/PackageLoader.cs
```diff
@@ -333,5 +379,125 @@ private SourceRepository GetSourceRepository(PackageSource source)
...
+                zip.ExtractToDirectory(path);
+
+                return true;
+            }
+            catch (InvalidDataException)
+            {
+            }
+            catch (IOException)
+            {
+            }
```
sunandabalu, line 416: Why empty catches? need to log something here.

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Commands/ExtensionManagement/ExtensionManagementCommand.cs
twsouthwick: It's not a constant any more, but a setting brought in via appsettings.json

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions.Internal/IUpgradeAssistantConfigurationLoader.cs
twsouthwick: No, this will stay sync (maybe go back to async). However, the config file is small, so we can just load it into memory and then parse it via the sync patterns

src/components/Microsoft.DotNet.UpgradeAssistant.Extensions/ExtensionManager.cs
twsouthwick: Which loop are you refering to? The initializer is only invoked once (see https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/statements/iteration-statements#the-for-statement) if that's what you're referring to

src/components/Microsoft.DotNet.UpgradeAssistant.Extensions/ExtensionManager.cs
twsouthwick: This is unnecessary and I'll just remove it.

src/components/Microsoft.DotNet.UpgradeAssistant.Extensions/ExtensionManager.cs
twsouthwick: To remain consistent with the other signatures. Also, it may end up being async and reduces change (for instance, one of these methods went from sync->async and it caused a bit of annoying churn)

src/components/Microsoft.DotNet.UpgradeAssistant.Extensions/ExtensionOptions.cs
twsouthwick: They're well-defined system folders. See https://docs.microsoft.com/en-us/dotnet/api/system.environment.specialfolder?view=net-5.0 for some details

src/components/Microsoft.DotNet.UpgradeAssistant.Extensions/ExtensionManager.cs
sunandabalu: `var s = Registered.Select(_extensionDownloader.GetInstallPath);
foreach (var path in s)
{
...
}`

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/PackageLoader.cs
twsouthwick: There's a generalized error below these. If successful, nothing beyond the try will be hit; on the caught exceptions it does. That's so there's less duplication as we just want to let them know it's amalformed package.

src/components/Microsoft.DotNet.UpgradeAssistant.Extensions/ExtensionManager.cs
twsouthwick: Not needed - initializers are only evaluated once (see the linked spec)

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/PackageLoader.cs
twsouthwick: The receiver takes ownership. I wish we had a good way of defining that in code, but we don't :(

sunandabalu (approved):

twsouthwick:
This change adds a NuGet backed way of creating/adding/restoring extensions.

With this change, you can add/remove/list/use extensions. However, in order to restore on demand things that are currently synchronous need to be made async. That will be done in a subsequent change to minimize the amount of churn.

PR #764, opened 2021-08-02
Add AnalysisProvider for diagnostic analyzers

### LINKED ISSUE #19: Surface unsupported API categories
Reduce noise by excluding the unavailable APIs - https://docs.microsoft.com/en-us/dotnet/core/porting/net-framework-tech-unavailable?toc=/dotnet/fundamentals/toc.json&bc=/dotnet/breadcrumb/toc.json

sunandabalu:
Fixes #19


sunandabalu:
PR Build - https://dev.azure.com/dnceng/public/_build/results?buildId=1272256&view=results

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source/AnalyzeDiagnostics.cs
```diff
@@ -0,0 +1,139 @@
...
+                {
+                    Diagnostics = Enumerable.Empty<Diagnostic>();
+                }
+                else
+                {
+                    var compilationWithAnalyzer = compilation
+                        .WithAnalyzers(ImmutableArray.CreateRange(applicableAnalyzers), new CompilationWithAnalyzersOptions(new AnalyzerOptions(_additionalTexts), ProcessAnalyzerException, true, true));
+
+                    // Find all diagnostics that registered analyzers produce
+                    Diagnostics = (await compilationWithAnalyzer.GetAnalyzerDiagnosticsAsync(token).ConfigureAwait(false)).Where(d => d.Location.IsInSource);
```
twsouthwick, line 124: Why keep state? Why not return this as the diagnostics used?

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source/AnalyzeDiagnostics.cs
```diff
@@ -0,0 +1,139 @@
...
+
+namespace Microsoft.DotNet.UpgradeAssistant.Steps.Source
+{
+    public class AnalyzeDiagnostics : IAnalyzeResultProvider
+    {
+        private readonly IEnumerable<DiagnosticAnalyzer> _allAnalyzers;
+        private readonly ImmutableArray<AdditionalText> _additionalTexts;
+        private readonly ILogger _logger;
+
+        internal IEnumerable<Diagnostic> Diagnostics { get; set; } = Enumerable.Empty<Diagnostic>();
```
twsouthwick, line 23: Why is this internal? What's it used for? Does it need to maintain state of the result provider?

twsouthwick (changes requested):

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/AnalyzeResult.cs
```diff
@@ -8,8 +8,15 @@ namespace Microsoft.DotNet.UpgradeAssistant.Analysis
...
+    }
+
+    public record ResultObj
+    {
+        public int LineNumber { get; init; }
+
         public string FileLocation { get; init; } = string.Empty;
 
-        public HashSet<string> Results { get; init; } = new HashSet<string>();
+        public string ResultMessage { get; init; } = string.Empty;
```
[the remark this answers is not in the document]
sunandabalu: Its conflcting with the Result object from Sarif, so had to rename these.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source/AnalyzeDiagnostics.cs
```diff
@@ -0,0 +1,139 @@
...
+                throw new ArgumentNullException(nameof(additionalTexts));
+            }
+
+            if (logger is null)
+            {
+                throw new ArgumentNullException(nameof(logger));
+            }
+
+            _logger = logger;
+            _allAnalyzers = analyzers.OrderBy(a => a.SupportedDiagnostics.First().Id);
```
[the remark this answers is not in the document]
sunandabalu: Oops, let me change that. I do need all the diagnostics.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source/AnalyzeDiagnostics.cs
```diff
@@ -0,0 +1,139 @@
...
+    {
+        private readonly IEnumerable<DiagnosticAnalyzer> _allAnalyzers;
+        private readonly ImmutableArray<AdditionalText> _additionalTexts;
+        private readonly ILogger _logger;
+
+        internal IEnumerable<Diagnostic> Diagnostics { get; set; } = Enumerable.Empty<Diagnostic>();
+
+        public string Id => "UA102";
+
+        public string Name => "API Upgradability";
```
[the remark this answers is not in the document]
sunandabalu: This is the name for the category of analysis.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/AnalyzeResult.cs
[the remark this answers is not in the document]
sunandabalu: The change was to support the line number needing to be associated with a message and a file location for diagnostic analyzers.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source/AnalyzeDiagnostics.cs
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        private HashSet<ResultObj> ProcessDiagnostics()
+        {
+            var results = new HashSet<ResultObj>();
+            foreach (var diag in Diagnostics)
+            {
+                results.Add(new()
+                {
+                    // Diagnostic line numbers are zero-based so offsetting by one
```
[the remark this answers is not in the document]
sunandabalu: VS has 1-based line numbers whereas the diagnostic GetLifeSpan has 0-based line numbers, there is comments on the GetLifeSpan method on the same.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source/AnalyzeDiagnostics.cs
sunandabalu: I was following the pattern in the SourceUpdaterStep.cs

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source/AnalyzeDiagnostics.cs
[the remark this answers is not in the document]
sunandabalu: https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.text.lineposition.line?view=roslyn-dotnet-3.10.0

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source/AnalyzeDiagnostics.cs
sunandabalu: Changed

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source/AnalyzeDiagnostics.cs
sunandabalu: removed

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source/AnalyzeDiagnostics.cs
```diff
@@ -0,0 +1,133 @@
...
+                    LineNumber = diag.Location.GetLineSpan().Span.End.Line + 1,
+                    FileLocation = diag.Location.GetLineSpan().Path,
+                    ResultMessage = diag.Descriptor.Description.ToString(System.Globalization.CultureInfo.InvariantCulture),
+                });
+            }
+
+            return results;
+        }
+
+        private static IAsyncEnumerable<DiagnosticAnalyzer> GetApplicableAnalyzersAsync(IEnumerable<DiagnosticAnalyzer> analyzers, IProject project)
```
twsouthwick, line 85: This doesn't do anything - none of the analyzers have the components anymore. If you'd like to see that, we can add an issue, but this logic here should be centralized rather than a private method here.

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source/AnalyzeDiagnostics.cs
```diff
@@ -0,0 +1,133 @@
...
+            => analyzers.ToAsyncEnumerable()
+                        .WhereAwaitWithCancellation((a, token) => a.GetType().AppliesToProjectAsync(project, token));
+
+        private async Task<IEnumerable<Diagnostic>> GetDiagnosticsAsync(IProject project, CancellationToken token)
+        {
+            var diagnostics = Enumerable.Empty<Diagnostic>();
```
twsouthwick, line 91: Do you really need a local variable here? You can just return the `Enumerable.Empty` where needed, and then return the diagnostics directly from where they are retrieved

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source/AnalyzeDiagnostics.cs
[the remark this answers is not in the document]
twsouthwick: So maybe there's a meta discussion we should have as to what the id/name here represent. If I were to get this report, I would expect this to show me an id that I can correlate back to a roslyn analyzer

twsouthwick (changes requested):

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source/AnalyzeDiagnostics.cs
```diff
@@ -0,0 +1,139 @@
...
+{
+    public class AnalyzeDiagnostics : IAnalyzeResultProvider
+    {
+        private readonly IEnumerable<DiagnosticAnalyzer> _allAnalyzers;
+        private readonly ImmutableArray<AdditionalText> _additionalTexts;
+        private readonly ILogger _logger;
+
+        internal IEnumerable<Diagnostic> Diagnostics { get; set; } = Enumerable.Empty<Diagnostic>();
+
+        public string Id => "UA102";
```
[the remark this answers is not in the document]
sunandabalu: Changed to use the diagnostic id for diagnostic analyzers

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source/AnalyzeDiagnostics.cs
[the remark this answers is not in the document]
sunandabalu: Diagnostic description is added to results, this will be used for Tool name only.

src/components/Microsoft.DotNet.UpgradeAssistant.Analysis/AnalyzeResultWriter.cs
```diff
@@ -51,40 +52,27 @@ await foreach (var ar in analyzeResults)
...
                         },
                     },
-                    Results = await ExtractResultsAsync(ar).ConfigureAwait(false),
+                    Results = ExtractResults(analyzeResults),
                 };
             }
         }
 
-        private static async Task<IList<Result>> ExtractResultsAsync(AnalyzeResultDefinition result)
+        private static IList<Result> ExtractResults(IList<AnalyzeResult> analyzeResults)
```
twsouthwick, line 67: nit: do you need IList<>? Since this is private, can it just take List<>? Same for return type

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source/AnalyzeDiagnostics.cs
```diff
@@ -0,0 +1,134 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.DotNet.UpgradeAssistant.Analysis;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.DotNet.UpgradeAssistant.Steps.Source
+{
+    public class AnalyzeDiagnostics : IAnalyzeResultProvider
```
twsouthwick, line 17: nit: this name is pretty vague. Maybe `DiagnosticAnalyzerResultProvider`?

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source/AnalyzeDiagnostics.cs
```diff
@@ -0,0 +1,134 @@
...
+    {
+        private readonly IEnumerable<DiagnosticAnalyzer> _allAnalyzers;
+        private readonly ImmutableArray<AdditionalText> _additionalTexts;
+        private readonly ILogger _logger;
+
+        public string Name => "API Upgradability";
+
+        public Uri InformationURI => new("https://docs.microsoft.com/en-us/dotnet/core/porting/upgrade-assistant-overview");
+
+        public AnalyzeDiagnostics(IEnumerable<DiagnosticAnalyzer> analyzers, IEnumerable<AdditionalText> additionalTexts, ILogger<AnalyzeDiagnostics> logger)
```
twsouthwick, line 27: This appears to duplicate a lot of logic from the source updaters. Can a new service be created that provides the diagnostics to minimize the duplication?

twsouthwick (changes requested):
I don't see any tests for anything. Can you add some?

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source/AnalyzeDiagnostics.cs
sunandabalu: Removed the filter

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source/AnalyzeDiagnostics.cs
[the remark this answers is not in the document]
sunandabalu: Talking to @mjrousos this is just to order the list of diagnostics, this does not filter the list. putting this line back in.

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source.Tests/DiagnosticAnalysisRunnerTests.cs
```diff
@@ -0,0 +1,140 @@
...
+            // Arrange
+            using var mock = AutoMock.GetLoose();
+            var loggerMock = new Mock<ILogger<DiagnosticAnalysisRunner>>();
+
+            var project = CreateProject(mock);
+            project.Setup(p => p.GetRoslynProject()).Returns(() =>
+            {
+                var ws = new AdhocWorkspace();
+                var name = Path.GetFileNameWithoutExtension(TestProjectPath)!;
+                return ws.AddProject(ProjectInfo.Create(ProjectId.CreateNewId(), VersionStamp.Default, name, name, "C#", filePath: TestProjectPath));
```
twsouthwick, line 54: Do you need this TestProjectPath?

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source.Tests/DiagnosticAnalysisRunnerTests.cs
```diff
@@ -0,0 +1,140 @@
...
+        public async Task DiagnosticsWithEmptyProjectTest()
+        {
+            // Arrange
+            using var mock = AutoMock.GetLoose();
+            var loggerMock = new Mock<ILogger<DiagnosticAnalysisRunner>>();
+
+            var project = CreateProject(mock);
+            project.Setup(p => p.GetRoslynProject()).Returns(() =>
+            {
+                var ws = new AdhocWorkspace();
```
twsouthwick, line 52: I usually create these kind of things outside of the returns call. This *may be* fine, just verify that if it gets called twice, will it get the same one?

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source.Tests/DiagnosticAnalysisRunnerTests.cs
```diff
@@ -0,0 +1,140 @@
...
+                var name = Path.GetFileNameWithoutExtension(TestProjectPath)!;
+                return ws.AddProject(ProjectInfo.Create(ProjectId.CreateNewId(), VersionStamp.Default, name, name, "C#", filePath: TestProjectPath));
+            });
+
+            var context = mock.Mock<IUpgradeContext>();
+            context.Setup(c => c.Projects).Returns(new[] { project.Object });
+            var diagnosticAnalysisRunner = new DiagnosticAnalysisRunner(GetAnalyzers(false), AdditionalTexts, loggerMock.Object);
```
twsouthwick, line 59: You shouldn't have to create it like this. We use Automock so you can just do `mock.Create<DiagnosticAnalysisRunner>()` and it will compose it correctly

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source.Tests/DiagnosticAnalysisRunnerTests.cs
```diff
@@ -0,0 +1,140 @@
...
+
+        private Mock<IProject> CreateProject(AutoMock mock)
+        {
+            var project = mock.Mock<IProject>();
+            var nugetReferences = mock.Mock<INuGetReferences>();
+            nugetReferences.Setup(n => n.IsTransitivelyAvailableAsync(It.IsAny<string>(), default))
+                .Returns(new ValueTask<bool>(false));
+            var projectFile = mock.Mock<IProjectFile>();
+            projectFile.Setup(f => f.IsSdk).Returns(true);
+            projectFile.Setup(f => f.Sdk).Returns(new HashSet<string>(new[] { "Microsoft.NET.Sdk.Web" }));
```
twsouthwick, line 107: Are all these properties needed for your test? These look like they had been copied from a different test that needed different properties set. This should mock a minimal IProject

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source.Tests/DiagnosticAnalysisRunnerTests.cs
```diff
@@ -0,0 +1,140 @@
...
+            project.Setup(p => p.NuGetReferences).Returns(nugetReferences.Object);
+            project.Setup(p => p.FileInfo).Returns(new FileInfo(_fixture.Create<string>()));
+
+            return project;
+        }
+
+        private IEnumerable<DiagnosticAnalyzer> GetAnalyzers(bool includeLocation)
+        {
+            var analyzers = new List<DiagnosticAnalyzer>();
+            for (int i = 0; i < 3; i++)
```
twsouthwick, line 123: Do you need 3 of them?

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source.Tests/DiagnosticAnalysisRunnerTests.cs
```diff
@@ -0,0 +1,140 @@
...
+using Xunit;
+
+namespace Microsoft.DotNet.UpgradeAssistant.Steps.Source.Tests
+{
+    public class DiagnosticAnalysisRunnerTests
+    {
+        private readonly Fixture _fixture;
+        internal const string TestProjectPath = @"assets\TestProject.csproj";
+
+        private static IEnumerable<AdditionalText> AdditionalTexts
```
twsouthwick, line 29: Why this? Can you just use `Enumerable.Empty<AdditionalText>()` where necessary?

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source.Tests/DiagnosticAnalysisRunnerTests.cs
```diff
@@ -0,0 +1,140 @@
...
+
+            var context = mock.Mock<IUpgradeContext>();
+            context.Setup(c => c.Projects).Returns(new[] { project.Object });
+            var diagnosticAnalysisRunner = new DiagnosticAnalysisRunner(GetAnalyzers(includeLocation), AdditionalTexts, loggerMock.Object);
+
+            // Act
+            var diagnostics = await diagnosticAnalysisRunner.GetDiagnosticsAsync(project.Object, CancellationToken.None).ConfigureAwait(false);
+
+            // Assert
+            Assert.True(diagnostics.Count() == expected);
```
twsouthwick, line 96: You should verify they are the ones expected, ie by diagnostic id

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source.Tests/DiagnosticAnalysisRunnerTests.cs
sunandabalu: I tried that first, but it was not able to pass in the parameters for `DiagnosticAnalysisRunner` - Analyzers. AdditionalTexts, etc.

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source.Tests/DiagnosticAnalysisRunnerTests.cs
sunandabalu: Following @mjrousos's pattern here - https://github.com/dotnet/upgrade-assistant/blob/main/tests/extensions/web/Microsoft.DotNet.UpgradeAssistant.Steps.Razor.Tests/RazorSourceUpdaterTests.cs#L364

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source/SourceUpdaterStep.cs
```diff
@@ -51,22 +51,22 @@ public class SourceUpdaterStep : UpgradeStep
...
             }
 
-            if (logger is null)
+            if (additionalTexts is null)
             {
-                throw new ArgumentNullException(nameof(logger));
+                throw new ArgumentNullException(nameof(additionalTexts));
             }
 
             _allAnalyzers = analyzers.OrderBy(a => a.SupportedDiagnostics.First().Id);
```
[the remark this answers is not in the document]
sunandabalu: Makes sense, I'll defer this change for a later PR. Issue to track it - https://github.com/dotnet/upgrade-assistant/issues/776

src/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source/DiagnosticAnalysisRunner.cs
```diff
@@ -0,0 +1,88 @@
...
+            {
+                throw new ArgumentNullException(nameof(logger));
+            }
+
+            _logger = logger;
+            _allAnalyzers = analyzers.OrderBy(a => a.SupportedDiagnostics.First().Id);
+            _additionalTexts = ImmutableArray.CreateRange(additionalTexts);
+        }
+
+        public async Task<IEnumerable<Diagnostic>> GetDiagnosticsAsync(IProject project, CancellationToken token)
```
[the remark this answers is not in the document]
sunandabalu: Not sure how much value IAsyncEnumerable will add here as the method call `GetAnalyzerDiagnosticsAsync` returns a `Task<ImmutableArray<Diagnostic>>` which we are then returning.

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source.Tests/DiagnosticAnalysisRunnerTests.cs
twsouthwick: Sure, but it's a pattern that I'm not sure it'll have the intended consequence. It may not matter here, but does it matter if the call is idempotent?

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source.Tests/DiagnosticAnalysisRunnerTests.cs
```diff
@@ -0,0 +1,116 @@
...
+        {
+            var project = mock.Mock<IProject>();
+            var projectFile = mock.Mock<IProjectFile>();
+            project.Setup(p => p.GetFile()).Returns(projectFile.Object);
+            project.Setup(p => p.FileInfo).Returns(new FileInfo(_fixture.Create<string>()));
+
+            return project;
+        }
+
+        private IEnumerable<DiagnosticAnalyzer> GetAnalyzers(bool includeLocation)
```
twsouthwick, line 101: Why process a lsit?

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source.Tests/DiagnosticAnalysisRunnerTests.cs
```diff
@@ -0,0 +1,116 @@
...
+                var name = Path.GetFileNameWithoutExtension("TestProject.csproj")!;
+                var projectInfo = ProjectInfo.Create(ProjectId.CreateNewId(), VersionStamp.Default, name, name, "C#", filePath: "TestProject.csproj");
+                var documents = new List<DocumentInfo>() { DocumentInfo.Create(DocumentId.CreateNewId(projectInfo.Id), "Foo.cs") };
+                var proj = ws.AddProject(projectInfo.WithDocuments(documents));
+                return proj;
+            });
+
+            var context = mock.Mock<IUpgradeContext>();
+            context.Setup(c => c.Projects).Returns(new[] { project.Object });
+            var diagnosticAnalysisRunner = new RoslynDiagnosticProvider(GetAnalyzers(includeLocation), Array.Empty<AdditionalText>(), loggerMock.Object);
```
twsouthwick, line 78: You should call `mock.Create<RoslynDiagnosticProvider>()` here instead of newing it up yourself.

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source.Tests/DiagnosticAnalysisRunnerTests.cs
```diff
@@ -0,0 +1,116 @@
...
+            // Arrange
+            using var mock = AutoMock.GetLoose();
+            var loggerMock = new Mock<ILogger<RoslynDiagnosticProvider>>();
+
+            var project = CreateProject(mock);
+            project.Setup(p => p.GetRoslynProject()).Returns(() =>
+            {
+                var ws = new AdhocWorkspace();
+                var name = Path.GetFileNameWithoutExtension("TestProject.csproj")!;
+                return ws.AddProject(ProjectInfo.Create(ProjectId.CreateNewId(), VersionStamp.Default, name, name, "C#", filePath: "TestProject.csproj"));
```
twsouthwick, line 42: Is this filepath required?

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source.Tests/DiagnosticAnalysisRunnerTests.cs
```diff
@@ -0,0 +1,116 @@
...
+
+            var context = mock.Mock<IUpgradeContext>();
+            context.Setup(c => c.Projects).Returns(new[] { project.Object });
+            var diagnosticAnalysisRunner = new RoslynDiagnosticProvider(GetAnalyzers(includeLocation), Array.Empty<AdditionalText>(), loggerMock.Object);
+
+            // Act
+            var diagnostics = await diagnosticAnalysisRunner.GetDiagnosticsAsync(project.Object, CancellationToken.None).ConfigureAwait(false);
+
+            // Assert
+            Assert.True(diagnostics.Count() == expectedDiagnosticCount);
```
twsouthwick, line 84: Instead of this pattern, if you do a collection, it's better to do:

```cs
Assert.Collection(diagnostics,
  r => Assert....,
  .... [for each item]);
```

twsouthwick (approved):
A few comments to consider

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source.Tests/Microsoft.DotNet.UpgradeAssistant.Steps.Source.Tests.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.Workspaces.MSBuild" Version="3.8.0" />
```
twsouthwick, line 8: You shouldn't need this

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source.Tests/DiagnosticAnalysisRunnerTests.cs
sunandabalu: One of the testdata is to check for empty diagnostics, in that case the Assert.Collection was failing which is why I had to put it under `if(diagnostics.Any())`.

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source.Tests/DiagnosticAnalysisRunnerTests.cs
sunandabalu: As I mentioned before, I did try mocking the `RoslynDiagnosticProvider` but it was unable to take in parameters like Analyzers, AdditionalTexts and Logger - `System.NullReferenceException : Object reference not set to an instance of an object.
` 
So I decided to new up instead.

tests/steps/Microsoft.DotNet.UpgradeAssistant.Steps.Source.Tests/DiagnosticAnalysisRunnerTests.cs
twsouthwick: Yup, when `expectedDiagnosticCount == 0`. It's ok as is, but the collection ends up being a bit better for readability and understanding what went wrong when something fails

sunandabalu: Agreed, `Assert.Collection` provides better readability than the plain old `Assert.True / False`

PR #791, opened 2021-08-11
Add project component filtering to codefixers

twsouthwick:
This adds them back by linking the file in and then the applies to will
verify based on the shape of the API and not the exact instance.


tests/common/Microsoft.DotNet.UpgradeAssistant.Abstractions.Tests/ProjectExtensionsTests.cs
```diff
@@ -85,5 +85,17 @@ public async Task CanConfigUpdaterApplyToMultipleLanguagesAsync(ProjectComponent
...
+        [ApplicableLanguage(Language.CSharp, Language.FSharp)]
+        private class TestObjectMultipleLanguage
+        {
+        }
+
+        [ApplicableComponents(ProjectComponents.AspNetCore)]
+        [ExportCodeFixProvider(LanguageNames.CSharp, LanguageNames.FSharp, Name = "UHOH1 CodeFix Provider")]
+        private class TestCodeFixer
+        {
+        }
```
sunandabalu, line 99: Why empty classes?

sunandabalu (approved):

tests/common/Microsoft.DotNet.UpgradeAssistant.Abstractions.Tests/ProjectExtensionsTests.cs
twsouthwick: For test purposes - it verifies attributes extraction

PR #796, opened 2021-08-16
Update loose binary idenfication docs for current UX

twsouthwick:


docs/design/Loose_binary_identification.md
```diff
@@ -18,21 +18,33 @@ The main scenario this will focus on is one where a project has downloaded an as
...
+
+1. Set an environment variable for `UA_FEATURES=EXTENSION_MANAGEMENT`. Extension management is currently under preview, but is required to test this feature.
+2. Restore the extensions installed
+   ```
+   upgrade-assistant extensions restore
+   ```
+
+In order to update to the latest versions, the extensions may be updated:
+
+3. Set an environment variable for `UA_FEATURES=EXTENSION_MANAGEMENT`. Extension management is currently under preview, but is required to test this feature.
```
sunandabalu, line 43: Do the customers need to set this environment variable 3 times? looks that way from the documentation here. Could you make it clear if the setting will persist?

sunandabalu (changes requested):
This section needs to be a separate usage scenario doc including command to run to analyze and upgrade using the extensions, along with the pre-setup for the commands to work (set env variable, download extensions etc.)

docs/design/Loose_binary_identification.md
twsouthwick: That's up to the user. It's an environment variable and they can manage that however they want

twsouthwick:
So you'd like a separate doc for usage of the extension without the context of what it's supposed to do?

sunandabalu:
> So you'd like a separate doc for usage of the extension without the context of what it's supposed to do?

you can always have "For more context on how extensions work see here..." at the beginning of the doc. That way the usage doc is specific to the scenario - "how to resolve loose assemblies in the project" , using extensions is an implementation detail that customers may not be interested in.

twsouthwick:
The doc right now focuses on the the experience of identifying loose dependencies. Can you provide an outline of what you would like to see?

sunandabalu:
With the `How to use` section in the Loose-Assembly doc, I think its mostly there. I'm thinking about a line under https://github.com/dotnet/upgrade-assistant#analyze-solution-prior-to-upgrade to say "Loose-Assembly resolution to a corresponding package (if extensions are added as described here (link to "how to use" section)" Also a similar line talking about upgrade under https://github.com/dotnet/upgrade-assistant#upgrade-solution. This way we are only sharing one top level section to customers.

twsouthwick:
So the section in `Experimental Features` you want duplicated in each of those sections? How do you propose to clearly mark that it's experimental and not GA'd?

sunandabalu (approved):
Fine for now.

PR #800, opened 2021-08-17
Alert on deprecated controls in Winforms

sunandabalu:
Fixes 
#782
#783

PR #802, opened 2021-08-17
Add command option for --msbuild-path

twsouthwick:
This change adds a way to pass a custom path for msbuild through to the engine. This allows for users to select preview versions if they would like until we figure out a better way to automatically select MSBuild versions.

This change separates out the MSBuild path into the `WorkspaceOptions` object and refactors the logic into `IConfigureOptions<>` implementations so that they will be cached automatically by the system.


twsouthwick:
Blocked by #800

sunandabalu (approved):

PR #816, opened 2021-08-24
Host try-convert in process instead of calling exe

twsouthwick:
This change incorporates try-convert assemblies into the build process
and the default ITryConvertTool implementation. A feature flag
TRY_CONVERT_EXE is used to switch back to using the exe. This will be
removed in a future release (along with the packaged try-convert) if it
is working well.


twsouthwick:
This change also includes some changes to the `IdentifierUpgradeCodeFixer` so that the imports and simplification are only annotated. The actual changes to add imports, simplify etc, should be done outside of the codefixer itself which limits potential conflicts in the diffing engine.

twsouthwick:
FYI This builds on #819 and will be simplified once that merges in

PR #850, opened 2021-09-15
Fix to avoid operation on the collection being iterated upon

### LINKED ISSUE #836: Package analysis failed
- [ ] Include the log file `upgrade-assistant.clef` that is produced in the working directory

### Describe the bug

> FTL] Package analysis failed (analyzer Target compatibility reference analyzer: Collection was modified; enumeration operation may not execute.

Bug believed to be here
https://github.com/dotnet/upgrade-assistant/blob/09731053296437986b252866f024ff9e4ce8ee08/src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Analyzers/TargetCompatibilityReferenceAnalyzer.cs#L45
https://github.com/dotnet/upgrade-assistant/blob/09731053296437986b252866f024ff9e4ce8ee08/src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Analyzers/TargetCompatibilityReferenceAnalyzer.cs#L88

### Exceptions (if any)

Likely `System.InvalidOperationException`

### Further technical details
`systeminfo | findstr /B /C:"OS Name" /C:"OS Version"`
OS Name:                   Microsoft Windows 10 Pro
OS Version:                10.0.19043 N/A Build 19043

`upgrade-assistant --version`
0.2.241603+99dc6c7bf12364ae199f7acd8d0d42c631508de0

`dotnet --list-sdks`
2.2.207 [C:\Program Files\dotnet\sdk]
3.1.101 [C:\Program Files\dotnet\sdk]
3.1.401 [C:\Program Files\dotnet\sdk]
5.0.302 [C:\Program Files\dotnet\sdk]
5.0.400 [C:\Program Files\dotnet\sdk]

sunandabalu:
Fixes #836

### DIFF
--- a/src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Analyzers/TargetCompatibilityReferenceAnalyzer.cs
+++ b/src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Analyzers/TargetCompatibilityReferenceAnalyzer.cs
@@ -42,7 +42,7 @@ namespace Microsoft.DotNet.UpgradeAssistant.Steps.Packages.Analyzers
                 throw new ArgumentNullException(nameof(state));
             }
 
-            IDependencyCollection<NuGetReference> packages = state.Packages;
+            var packages = state.Packages.ToList();
 
             foreach (var packageReference in packages)
             {
--- a/src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Analyzers/TargetCompatibilityReferenceAnalyzer.cs
+++ b/src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Analyzers/TargetCompatibilityReferenceAnalyzer.cs
@@ -42,6 +42,7 @@ namespace Microsoft.DotNet.UpgradeAssistant.Steps.Packages.Analyzers
                 throw new ArgumentNullException(nameof(state));
             }
 
+            // Making a copy of state.Packages collection to avoid modifying a collection that is being enumerated.
             var packages = state.Packages.ToList();
 
             foreach (var packageReference in packages)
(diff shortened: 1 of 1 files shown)


src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Analyzers/TargetCompatibilityReferenceAnalyzer.cs
```diff
@@ -42,7 +42,9 @@ public async Task AnalyzeAsync(IProject project, IDependencyAnalysisState state,
                 throw new ArgumentNullException(nameof(state));
             }
 
-            foreach (var packageReference in state.Packages)
+            IDependencyCollection<NuGetReference> packages = state.Packages;
```
mjrousos, line 45: Does this actually fix the issue? `packages` and `state.Packages` will still refer to the same collection, right? So wouldn't the issue of mutating a collection while foreach-ing over it still happen?

ryanliang88: I thought that this wouldn't work and would for example require an explicit for loop

sunandabalu: Added an explicit toList so a new copy will get created.

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Analyzers/TargetCompatibilityReferenceAnalyzer.cs
```diff
@@ -42,7 +42,9 @@ public async Task AnalyzeAsync(IProject project, IDependencyAnalysisState state,
                 throw new ArgumentNullException(nameof(state));
             }
 
-            foreach (var packageReference in state.Packages)
+            var packages = state.Packages.ToList();
```
twsouthwick, line 45: I wonder if we were to back the packages with an immutable collection if it would work without copying. For now, this is probably the easiest fix. Can you add a comment why we're doing a `.ToList()` here?

twsouthwick (approved):

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Analyzers/TargetCompatibilityReferenceAnalyzer.cs
sunandabalu: Added

mjrousos (commented):
LGTM

mjrousos (approved):

PR #855, opened 2021-09-17
Add loose assembly infrastructure

twsouthwick:


twsouthwick:
@marklio There's a bit of churn in the indexing client code in order to fit the style constraints of this project. We can discuss later how that affects other users of the code so it doesn't get out of sync.

twsouthwick:
@Sweekriti91 This change brings in some helpers to create nupkg redistributables of the extensions. We can adapt the MAUI one to do this as well

src/extensions/looseassembly/LooseAssembly/NuGetPackageLookup.cs
```diff
@@ -70,7 +71,9 @@ public async IAsyncEnumerable<NuGetReference> SearchAsync(string path, IEnumerab
         {
             foreach (var index in _indexes.Value)
             {
-                index.FindNuGetPackageInfoForFile(path, out var ownerPackageId, out var containingPackage);
+                using var stream = File.OpenRead(path);
```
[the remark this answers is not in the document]
twsouthwick: Yup I had that thought as well! I'll see what I can do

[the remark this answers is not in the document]
twsouthwick: Added a `LookupToken` that will calculate it and store it for future use

src/extensions/looseassembly/Client/HashToken.cs
```diff
@@ -15,6 +15,8 @@ namespace Microsoft.DotNet.UpgradeAssistant.Extensions.LooseAssembly.Client
     /// </summary>
     public readonly struct HashToken : IEquatable<HashToken>, IComparable<HashToken>
     {
+        private static readonly SHA256 Sha256 = SHA256.Create();
```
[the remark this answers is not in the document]
twsouthwick: done

src/extensions/looseassembly/Client/StrongNameInfo.cs
```diff
@@ -0,0 +1,83 @@
...
+                }
+
+                var reader = peReader.GetMetadataReader();
+
+                if (!reader.IsAssembly)
+                {
+                    return default;
+                }
+
+                var assemblyName = reader.GetAssemblyDefinition().GetAssemblyName();
```
marklio, line 63: It's worth noting that GetAssemblyDefinition can fail (return null?) for modules. I haven't looked at the behavior here. It's rare, but it's probably worth hardening for.

twsouthwick: Would the call to `IsAssembly` be sufficient ( a few lines up)?

twsouthwick: Per the code (https://source.dot.net/#System.Reflection.Metadata/System/Reflection/Metadata/MetadataReader.cs,0fb3765222daa266) it looks like that's the check that would throw.

I've already got that. Is there any other hardening you're thinking of here?

marklio: Yeah, you got it. I didn't notice the IsAssembly check. That's sufficient here. And IsAssembly is doing the equivalent number of rows check that I was previously doing. Seems fine.

azure-pipeline.yml
```diff
@@ -98,6 +100,11 @@ stages:
             -prepareMachine
             $(_BuildArgs)
           displayName: Windows Build / Publish
+        - task: PowerShell@2
+          inputs:
```
sunandabalu, line 104: I think this powershell task need to be before the build/publish task coz the build/publish task is the one that signs files. Did you test if the package created here is being signed?

twsouthwick:
Semantic versioning isn't support on extensions, so right now it will just truncate the semantic part. Here's an issue to address the issue fully: https://github.com/dotnet/upgrade-assistant/issues/859

azure-pipeline.yml
twsouthwick: Oh - so build/publish does build/publish/sign? How do I hook into that?

sunandabalu: You just need to put the package in the folder that it picks for signing from, so basically putting your powershell task before the build/publish should sign your package along with others. you can verify that when its running your PR.

sunandabalu (approved):
Approving for the client code, the signing and publishing in this PR will not work and needs to be re-worked.

twsouthwick:
Issue to track the signing/publishing: https://github.com/dotnet/upgrade-assistant/issues/860

PR #869, opened 2021-09-30
Pack non-default extensions in pack target

twsouthwick:
This change enables normal `/t:pack` to be used to generate the
extension nupkg packages when `PackExtension=true` is included in the
csproj.


src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/build/Microsoft.DotNet.UpgradeAssistant.Abstractions.props
```diff
@@ -1,5 +1 @@
-﻿<Project>
-  <PropertyGroup>
-    <ZipExtensionMethod>false</ZipExtensionMethod>
-  </PropertyGroup>
-</Project>
```
sunandabalu, line 5: Do you need this file? seems like the reference to the props file has been removed from everywhere

sunandabalu:
From the build logs I can see that the looseassembly nuget packages are being created, there are 5 packages being constructed whereas only 4 are being signed, looks one of the looseassembly package is not getting signed, is that intentional?

`2021-09-30T00:45:22.4328163Z   Repacking 4 containers.
2021-09-30T00:45:22.7812325Z   Round 1: Signing 4 files.
`

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/build/Microsoft.DotNet.UpgradeAssistant.Abstractions.props
twsouthwick: Yeah - it's for the NuGet package. I moved the properties into the props so we can just reference it once

twsouthwick:
Not sure how to diagnose this. The package is being created in the same way. There are no assemblies in the `.NuGet` one, so we just need the signing to sign the nupkg itself.

sunandabalu (approved):

Issue #805, opened 2021-08-18
Analyze command should give more details in console output


mjrousos:
I've been experimenting with the Analyze command. The Sarif output file is really cool, but it would be great if all the updates needed were also listed in the console so that I could easily get a sense of needed changes as soon as the tool executes.

Currently, necessary package updates are listed but for diagnostics, the Analyze command only says `Identified 16 diagnostics in project eShopLegacyMVC`. For upgrade scenarios this is ok because the specific diagnostics are walked through in source updater step/sub-steps but, for the Analyze command, it would be useful to have all of the diagnostics listed out (or at least summarized by diagnostic ID and count).

sunandabalu:
Seems reasonable to me, we could just log to console when we are processing each of the diagnostics in the `ProcessDiagnostics` method - https://github.com/dotnet/upgrade-assistant/blob/main/src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Source/DiagnosticAnalyzerResultProvider.cs#L60 that way this happens only for analyze and not for upgrade.

sunandabalu:
Addressed by #877

PR #877, opened 2021-10-04
Winforms source updater for HighDPISetting

### LINKED ISSUE #785: Analyzer and CodeFixer for HighDPIMode

sunandabalu:
Fixes #785 
Also addresses #805

### DIFF
--- a/src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/StreamHelpers.cs
+++ b/src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/StreamHelpers.cs
@@ -1,0 +1,51 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.IO;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Windows
+{
+    public static class StreamHelpers
+    {
+        private const int DefaultBufferSize = 1024;
+
+        public static async Task CopyStreamWithNewLineAdded(string[] inputLines, Stream outputStream, string existingLine, string newLine)
+        {
+            if (inputLines is null)
+            {
+                throw new ArgumentNullException(nameof(inputLines));
+            }
+
+            if (outputStream is null)
+            {
+                throw new ArgumentNullException(nameof(outputStream));
+            }
+
+            if (existingLine is null)
+            {
+                throw new ArgumentNullException(nameof(existingLine));
+            }
+
+            if (newLine is null)
+            {
+                throw new ArgumentNullException(nameof(newLine));
+            }
+
+            using var output = new StreamWriter(outputStream, Encoding.UTF8, DefaultBufferSize, leaveOpen: true);
+            foreach (var i in inputLines)
+            {
+                var line = new StringBuilder(i);
+                if (i.Contains(existingLine))
+                {
+                    line.AppendLine();
+                    line.Append(string.Concat("\t", newLine));
+                }
+
+                await output.WriteLineAsync(line.ToString()).ConfigureAwait(false);
+            }
+        }
+    }
+}
--- a/src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/WinformsDefaultFontUpdater.cs
+++ b/src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/WinformsDefaultFontUpdater.cs
@@ -7,8 +7,10 @@ using System.Collections.Immutable;
 using System.IO;
 using System.Linq;
 using System.Text;
+using System.Text.Json;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Xml.XPath;
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Windows
@@ -16,14 +18,18 @@ namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Windows
     [ApplicableComponents(ProjectComponents.WinForms)]
     public class WinformsDefaultFontUpdater : IUpdater<IProject>
     {
+        private const int BufferSize = 65536;
+        private const string HighDPIConfiguration = "/configuration/System.Windows.Forms.ApplicationConfigurationSection";
+
         private readonly WindowsUtilities _utilities = new();
         private readonly ILogger<WinformsDefaultFontUpdater> _logger;
+        private readonly string _winformsSourceUpdateInfo = "Default font in Windows Forms has been changed from Microsoft Sans Serif to Seg Segoe UI, in order to change the default font use the API - Application.SetDefaultFont(Font font). For more details see here - https://devblogs.microsoft.com/dotnet/whats-new-in-windows-forms-in-net-6-0-preview-5/#application-wide-default-font. \nIts recommended to set HighDpiMode to be SystemAware for better results in Main() - Application.SetHighDpiMode(HighDpiMode.SystemAware)";
 
         public string Id => typeof(WinformsDefaultFontUpdater).FullName;
 
-        public string Title => "Default Font API Alert";
+        public string Title => "Winforms Source Updater";
 
-        public string Description => "Default font in Windows Forms has been changed from Microsoft Sans Serif to Seg Segoe UI, in order to change the default font use the API - Application.SetDefaultFont(Font font). For more details see here - https://devblogs.microsoft.com/dotnet/whats-new-in-windows-forms-in-net-6-0-preview-5/#application-wide-default-font";
+        public string Description => "Update code for Winforms project documents";
 
         public BuildBreakRisk Risk => BuildBreakRisk.Low;
 
@@ -32,9 +38,76 @@ namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Windows
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
         }
 
-        public Task<IUpdaterResult> ApplyAsync(IUpgradeContext context, ImmutableArray<IProject> inputs, CancellationToken token)
+        private string ProcessAppConfigFile(IProject project)
         {
-            return this.IsApplicableAsync(context, inputs, token);
+            var hdpiValue = "SystemAware";
+            var appConfigFilePath = project.FindFiles("App.config").FirstOrDefault();
+            ConfigFile appConfigFile;
+            if (File.Exists(appConfigFilePath))
+            {
+                appConfigFile = new ConfigFile(appConfigFilePath);
+                var highDPISetting = appConfigFile.Contents.XPathSelectElement(HighDPIConfiguration);
+                if (highDPISetting is not null)
+                {
+                    var hdpi = highDPISetting.Elements("add").Where(e => e.Attribute("key").Value == "DpiAwareness").FirstOrDefault();
+                    if (hdpi is not null)
+                    {
+                        hdpiValue = hdpi.Attribute("value").Value;
+                        _logger.LogDebug("Found DpiAwareness Setting with {value}", hdpiValue);
+                    }
+                }
+            }
+
+            return hdpiValue;
+        }
+
+        private async Task UpdateHighDPISetting(IProject project)
+        {
+            var programFile = project.FindFiles("Program.cs").FirstOrDefault();
+            var programFileContent = File.ReadAllLines(programFile);
+
+            if (!programFileContent.Contains("Application.SetHighDpiMode"))
+            {
+                var hdpiValue = ProcessAppConfigFile(project);
+                try
+                {
+                    using var outputStream = File.Create(programFile, BufferSize, FileOptions.Asynchronous | FileOptions.SequentialScan);
+
+                    await StreamHelpers.CopyStreamWithNewLineAdded(programFileContent, outputStream, "Application.EnableVisualStyles()", SR.Format("Application.SetHighDpiMode(HighDpiMode.{0});", hdpiValue)).ConfigureAwait(false);
+                }
+                catch (IOException exc)
+                {
+                    _logger.LogCritical(exc, "Error while editing Program.cs {Path}", programFile);
+                }
+
+                _logger.LogInformation("Updated Program.cs file at {Path} with HighDPISetting set to {hdpi}", programFile, hdpiValue);
+            }
+            else
+            {
+                _logger.LogInformation("Program.cs file at {Path} already contains HighDPISetting, no need to edit.", programFile);
+            }
+        }
+
+        public async Task<IUpdaterResult> ApplyAsync(IUpgradeContext context, ImmutableArray<IProject> inputs, CancellationToken token)
+        {
+            if (context is null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            var fileLocations = new List<string>();
+
+            foreach (var project in inputs)
+            {
+                if (await _utilities.IsWinFormsProjectAsync(project, token))
+                {
+                    _logger.LogWarning(_winformsSourceUpdateInfo);
+                    fileLocations.Add(Path.Combine(project.FileInfo.DirectoryName, project.FileInfo.Name));
+                    await UpdateHighDPISetting(project);
+                }
+            }
+
+            return new WinformsUpdaterResult(fileLocations.Any(), _winformsSourceUpdateInfo, fileLocations);
         }
 
         public async Task<IUpdaterResult> IsApplicableAsync(IUpgradeContext context, ImmutableArray<IProject> inputs, CancellationToken token)
@@ -50,12 +123,12 @@ namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Windows
             {
                 if (await _utilities.IsWinFormsProjectAsync(project, token))
                 {
-                    _logger.LogWarning(this.Description);
+                    _logger.LogWarning(_winformsSourceUpdateInfo);
                     fileLocations.Add(Path.Combine(project.FileInfo.DirectoryName, project.FileInfo.Name));
                 }
             }
 
-            return new WinformsUpdaterResult(fileLocations.Any(), this.Description, fileLocations);
+            return new WinformsUpdaterResult(fileLocations.Any(), _winformsSourceUpdateInfo, fileLocations);
         }
     }
 }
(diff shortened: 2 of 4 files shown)


src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/WinformsDefaultFontUpdater.cs
```diff
@@ -7,23 +7,29 @@
...
 {
     [ApplicableComponents(ProjectComponents.WinForms)]
     public class WinformsDefaultFontUpdater : IUpdater<IProject>
     {
+        private const int BufferSize = 65536;
+        private const string HighDPIConfiguration = "/configuration/System.Windows.Forms.ApplicationConfigurationSection";
+
         private readonly WindowsUtilities _utilities = new();
         private readonly ILogger<WinformsDefaultFontUpdater> _logger;
+        private readonly string _winformsSourceUpdateInfo = "Default font in Windows Forms has been changed from Microsoft Sans Serif to Seg Segoe UI, in order to change the default font use the API - Application.SetDefaultFont(Font font). For more details see here - https://devblogs.microsoft.com/dotnet/whats-new-in-windows-forms-in-net-6-0-preview-5/#application-wide-default-font. \nIts recommended to set HighDpiMode to be SystemAware for better results in Main() - Application.SetHighDpiMode(HighDpiMode.SystemAware)";
```
twsouthwick, line 26: This should be a resource string

twsouthwick (changes requested):
This looks like you're doing source updating but outside of Roslyn. Is there a reason for that? Also, there should be some tests to cover this change.

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/StreamHelpers.cs
```diff
@@ -0,0 +1,51 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.IO;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Windows
+{
+    public static class StreamHelpers
```
twsouthwick, line 11: nit: does this need to be public?

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/WinformsDefaultFontUpdater.cs
```diff
@@ -32,9 +38,79 @@ public WinformsDefaultFontUpdater(ILogger<WinformsDefaultFontUpdater> logger)
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
         }
 
-        public Task<IUpdaterResult> ApplyAsync(IUpgradeContext context, ImmutableArray<IProject> inputs, CancellationToken token)
+        private string ProcessAppConfigFile(IProject project)
+        {
+            var hdpiValue = "SystemAware";
+            var appConfigFilePath = project.FindFiles("App.config").FirstOrDefault();
+            ConfigFile appConfigFile;
```
twsouthwick, line 45: why declare here and define below?

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/WinformsDefaultFontUpdater.cs
```diff
@@ -32,9 +38,79 @@ public WinformsDefaultFontUpdater(ILogger<WinformsDefaultFontUpdater> logger)
...
+                var programFileContent = File.ReadAllLines(programFile);
+
+                if (!programFileContent.Contains("Application.SetHighDpiMode"))
+                {
+                    var hdpiValue = ProcessAppConfigFile(project);
+                    try
+                    {
+                        using var outputStream = File.Create(programFile, BufferSize, FileOptions.Asynchronous | FileOptions.SequentialScan);
+
+                        await StreamHelpers.CopyStreamWithNewLineAdded(programFileContent, outputStream, "Application.EnableVisualStyles()", SR.Format("Application.SetHighDpiMode(HighDpiMode.{0});", hdpiValue)).ConfigureAwait(false);
```
twsouthwick, line 78: This method call is interesting. Can you move the string to constants with descriptive names so it's clear what the purpose of each is?

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/StreamHelpers.cs
```diff
@@ -0,0 +1,51 @@
...
+            }
+
+            using var output = new StreamWriter(outputStream, Encoding.UTF8, DefaultBufferSize, leaveOpen: true);
+            foreach (var i in inputLines)
+            {
+                var line = new StringBuilder(i);
+                if (i.Contains(existingLine))
+                {
+                    line.AppendLine();
+                    line.Append(string.Concat("\t", newLine));
```
mjrousos, line 44: Nit: `\t` won't always be the right leading trivia. It might be better to concat whatever whitespace begins existingLine so that they match.

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/StreamHelpers.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Windows
+{
+    public static class StreamHelpers
+    {
+        private const int DefaultBufferSize = 1024;
+
+        public static async Task CopyStreamWithNewLineAdded(string[] inputLines, Stream outputStream, string existingLine, string newLine)
```
mjrousos, line 15: It would be good to have a doc comment here explaining the relationship between existingLine and newLine (newLine is added into the inputLines after any line containing existingLine).

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/WinformsDefaultFontUpdater.cs
mjrousos: Also, should it be two separate resource strings since these are two separate messages we want to give users?

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/WinformsDefaultFontUpdater.cs
```diff
@@ -32,9 +38,79 @@ public WinformsDefaultFontUpdater(ILogger<WinformsDefaultFontUpdater> logger)
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
         }
 
-        public Task<IUpdaterResult> ApplyAsync(IUpgradeContext context, ImmutableArray<IProject> inputs, CancellationToken token)
+        private string ProcessAppConfigFile(IProject project)
+        {
+            var hdpiValue = "SystemAware";
```
mjrousos, line 43: Nit: Should this and App.config be consts similar to `HighDPIConfiguration`?

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/WinformsDefaultFontUpdater.cs
```diff
@@ -32,9 +38,79 @@ public WinformsDefaultFontUpdater(ILogger<WinformsDefaultFontUpdater> logger)
...
+        }
+
+        private async Task UpdateHighDPISetting(IProject project)
+        {
+            var programFile = project.FindFiles("Program.cs").FirstOrDefault();
+            if (programFile is not null && File.Exists(programFile))
+            {
+                var programFileContent = File.ReadAllLines(programFile);
+
+                if (!programFileContent.Contains("Application.SetHighDpiMode"))
```
mjrousos, line 71: Wouldn't `Contains` look for a line that's an exact match for this string? I don't think that's what we want since something like `Application.SetHighDpiMode(HighDpiMode.Foo)` should be considered a match but I suspect wouldn't be since programFileContent wouldn't contain exactly `Application.SetHighDpiMode`.

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/WinformsDefaultFontUpdater.cs
mjrousos: Will WinForms apps always have a call to Application.EnableVisualStyles() that we can key off of?

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/WinformsDefaultFontUpdater.cs
```diff
@@ -32,9 +38,79 @@ public WinformsDefaultFontUpdater(ILogger<WinformsDefaultFontUpdater> logger)
...
+                {
+                    var hdpiValue = ProcessAppConfigFile(project);
```
mjrousos, line 73: I see we're setting the HighDpiMode value to whatever was in the app.config file previously. Is the app.config not respected on its own anymore?

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/WinformsDefaultFontUpdater.cs
```diff
@@ -32,9 +38,79 @@ public WinformsDefaultFontUpdater(ILogger<WinformsDefaultFontUpdater> logger)
...
+                        using var outputStream = File.Create(programFile, BufferSize, FileOptions.Asynchronous | FileOptions.SequentialScan);
+
+                        await StreamHelpers.CopyStreamWithNewLineAdded(programFileContent, outputStream, "Application.EnableVisualStyles()", SR.Format("Application.SetHighDpiMode(HighDpiMode.{0});", hdpiValue)).ConfigureAwait(false);
+                    }
+                    catch (IOException exc)
+                    {
+                        _logger.LogCritical(exc, "Error while editing Program.cs {Path}", programFile);
+                    }
+
+                    _logger.LogInformation("Updated Program.cs file at {Path} with HighDPISetting set to {hdpi}", programFile, hdpiValue);
```
mjrousos, line 85: Nit: This may belong in the try block since it's likely not true if you encounter an exception.

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/WinformsDefaultFontUpdater.cs
```diff
@@ -7,23 +7,29 @@
...
+        private const string HighDPIConfiguration = "/configuration/System.Windows.Forms.ApplicationConfigurationSection";
+
         private readonly WindowsUtilities _utilities = new();
         private readonly ILogger<WinformsDefaultFontUpdater> _logger;
+        private readonly string _winformsSourceUpdateInfo = "Default font in Windows Forms has been changed from Microsoft Sans Serif to Seg Segoe UI, in order to change the default font use the API - Application.SetDefaultFont(Font font). For more details see here - https://devblogs.microsoft.com/dotnet/whats-new-in-windows-forms-in-net-6-0-preview-5/#application-wide-default-font. \nIts recommended to set HighDpiMode to be SystemAware for better results in Main() - Application.SetHighDpiMode(HighDpiMode.SystemAware)";
 
         public string Id => typeof(WinformsDefaultFontUpdater).FullName;
 
-        public string Title => "Default Font API Alert";
+        public string Title => "Winforms Source Updater";
```
mjrousos, line 30: It would probably be good to rename this class and file to match the new broader scope of the type. Or just have two separate `IUpdater<IProject>` implementations.

mjrousos (changes requested):
Looks good. I left a few comments inline, but my only major concerns are:

1. I'm not sure we're checking for "Application.SetHighDpiMode" already being present in program.cs correctly.
2. We should have some unit tests.

mjrousos:
@twsouthwick I think the reason for not using Roslyn here is because some of the updates depend on non-source files (like app.config) that are harder to reach from Roslyn. I agree there should be unit tests, though.

twsouthwick:
That makes sense about not easily accessible. However, we *can* work around that fairly easily if we want to keep source manipulation to source analyzers (I'd prefer that as a matter of principal, but fine with well defined exceptions). Some thoughts on how to enable it with roslyn:

1. For the diagnostic to be available when using just the nupkg, we'd want to add a targets file that would add any config files to additional files (this is a pattern a number of analyzers do to ensure files are available)
2. For the diagnostic to work as a source analyzer in the context of UA, we'd want to update the runner to be aware of globbing patterns to add as additional files.

Regardless, I'd like to see tests to catch any regressions. If you continue down this path of non-Roslyn based, it would be nice to introduce some abstractions to the `IProject` to read/replace files (I've seen requests for that anyway) so you're not using `System.IO.File` static methods directly (this will help a ton with testing).

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/WinformsDefaultFontUpdater.cs
twsouthwick: I'd go with separate updaters - keep each one scoped so it's clear to a user what they're opting into

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/WinformsDefaultFontUpdater.cs
sunandabalu: Yes the app.config is not respected on it own, the expectation is to have the HighDPIMode setting in the main()

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/ProgramFileSpec.cs
```diff
@@ -0,0 +1,35 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+
+namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Windows
+{
+    public class ProgramFileSpec
```
twsouthwick, line 12: can this be a record?

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/WinformsDefaultFontUpdater.cs
```diff
@@ -23,7 +23,7 @@ public class WinformsDefaultFontUpdater : IUpdater<IProject>
 
         public string Title => "Default Font API Alert";
 
-        public string Description => "Default font in Windows Forms has been changed from Microsoft Sans Serif to Seg Segoe UI, in order to change the default font use the API - Application.SetDefaultFont(Font font). For more details see here - https://devblogs.microsoft.com/dotnet/whats-new-in-windows-forms-in-net-6-0-preview-5/#application-wide-default-font";
+        public string Description => Resources.DefFontMessage;
```
twsouthwick, line 26: nit: can you spell out the full name? i.e. `DefaultFontMessage`. no need to shorten things

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/WinformsDpiSettingUpdater.cs
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        private string ProcessAppConfigFile(IProject project)
+        {
+            var hdpiValue = "SystemAware";
+            var appConfigFilePath = project.FindFiles("App.config").FirstOrDefault();
+            if (File.Exists(appConfigFilePath))
+            {
+                var appConfigFile = new ConfigFile(appConfigFilePath);
+                var highDPISetting = appConfigFile.Contents.XPathSelectElement(HighDPIConfiguration);
```
twsouthwick, line 48: It would be nice to abstract the retrieval of keys so that you can just ask for them rather than using xpath...

tests/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows.Tests/WinformsDpiSettingUpdaterTests.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.Threading.Tasks;
+using Autofac.Extras.Moq;
+using Microsoft.Extensions.Logging;
+using Xunit;
+
+namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Windows.Tests
+{
+    public class WinformsDpiSettingUpdaterTests
+    {
+        [InlineData(new string[] { "Application.EnableVisualStyles();", "Application.SetCompatibleTextRenderingDefault(false);" }, false, true, new string[] { "Application.EnableVisualStyles();", "Application.SetHighDpiMode(HighDpiMode.SystemAware);", "Application.SetCompatibleTextRenderingDefault(false);" })] // Happy Path
```
twsouthwick, line 14: Can you set this up so that it's the actual expected file?

tests/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows.Tests/WinformsDpiSettingUpdaterTests.cs
```diff
@@ -0,0 +1,57 @@
...
+            projectFile.Setup(f => f.IsSdk).Returns(true);
+
+            var project = mock.Mock<IProject>();
+            project.Setup(p => p.GetFile()).Returns(projectFile.Object);
+            project.Setup(p => p.FileInfo).Returns(new FileInfo("./test"));
+
+            var context = mock.Mock<IUpgradeContext>();
+            context.Setup(c => c.Projects).Returns(new[] { project.Object });
+
+            var programFilePath = Path.Combine(Directory.GetCurrentDirectory(), "Program.cs");
```
twsouthwick, line 38: Can you use a temporary directory instead? Probably would be helpful to create a helper for this, but you'll have conflicts potentially if run in parallel. Probably a random name as well (does it matter what the name is)?

twsouthwick (approved):

PR #881, opened 2021-10-07
Simplify extension building

twsouthwick:
This change reduces the number of msbuild calls that need to occur so
that the CLI can just use the output of already built assemblies to grab
as extensions. This should help with performance as well as incremental
builds.

### DIFF
--- a/Extensions.targets
+++ b/Extensions.targets
@@ -65,7 +65,7 @@
       </Content>
     </ItemGroup>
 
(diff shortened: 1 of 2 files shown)


sunandabalu:
Looks like the PR build is still failing @twsouthwick

twsouthwick:
@sunandabalu That error was a transient one downloading the nuget looseassembly data file. I reorganized the targets to be smaller and clearer to understand as well. Once approved, I'll merge it in if you don't have other concerns.

sunandabalu (approved):

PR #910, opened 2021-11-03
Remove test abstraction and use base classes

twsouthwick:
Uses the AnalyzerTest<> and CodeFixTest<> types to build extension
methods rather than the customer interfaces.


KSchlobohm (approved):
looks like we just moved the extension methods around to new classes

PR #939, opened 2021-11-15
Improving SARIF with HelpUri, full description, specific rule name

eddynaka:
- [x] You've included unit or integration tests for your change, where applicable.

**PR Title**
Improving SARIF with HelpUri, full description, specific rule name

**PR Description**
This PR will include a few improvements in the SARIF, exposing a specific rule name, rule description, and help uri.
This PR will only change classes that uses the native provider. All the others will be changed in another iteration.


src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Commands/Analyze/ConsoleAnalyze.cs
```diff
@@ -46,8 +46,8 @@ await foreach (var provider in _providers.ToAsyncEnumerable().WhereAwait(async i
                 analyzeResultMap.Add(new()
                 {
                     Version = GetProviderVersion(provider),
-                    Name = provider.Name,
-                    InformationURI = provider.InformationURI,
+                    Name = provider.ToolName,
```
eddynaka, line 49: To make things clear, I'm renaming it from `Name` to `ToolName`.

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageAnalyzerResultProvider.cs
```diff
@@ -16,14 +16,17 @@ namespace Microsoft.DotNet.UpgradeAssistant.Steps.Packages
 {
     public class PackageAnalyzerResultProvider : IAnalyzeResultProvider
     {
-        private const string Id = "UA101";
+        private const string RuleId = "UA101";
+        private const string RuleName = "SomeRuleName";
+        private const string FullDescription = "Some full description";
+        private readonly Uri _helpUri = new("https://docs.microsoft.com/en-us/dotnet/core/porting/upgrade-assistant-overview");
```
eddynaka, line 22: This is one example where we could create a specific `RuleName`, `FullDescription`, and `HelpUri`.
This will make the SARIF more complete.

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageAnalyzerResultProvider.cs
```diff
@@ -86,24 +89,14 @@ private HashSet<AnalyzeResult> ExtractAnalysisResult(string fileLocation, IDepen
...
-            if (analysisState is null || !analysisState.AreChangesRecommended)
+            if (analysisState is not null && analysisState.AreChangesRecommended)
             {
-                results.Add(new()
-                {
-                    RuleId = Id,
-                    RuleName = this.Name,
-                    FileLocation = fileLocation,
-                    ResultMessage = "No package updates needed",
-                });
```
eddynaka, line 97: We do not need to emit a Result that is correct.

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageAnalyzerResultProvider.cs
```diff
@@ -86,24 +89,14 @@ private HashSet<AnalyzeResult> ExtractAnalysisResult(string fileLocation, IDepen
...
-            }
-            else
-            {
-                GetResults("Reference to ", " needs to be deleted ", analysisState.References.Deletions);
-                GetResults("Reference to ", " needs to be added ", analysisState.References.Additions);
-                GetResults("Package ", " needs to be deleted ", analysisState.Packages.Deletions);
-                GetResults("Package ", " needs to be added ", analysisState.Packages.Additions);
-                GetResults("Framework Reference to ", " needs to be deleted ", analysisState.FrameworkReferences.Deletions);
-                GetResults("Framework Reference to ", " needs to be added ", analysisState.FrameworkReferences.Additions);
+                GetResults("Reference to ", " needs to be deleted.", analysisState.References.Deletions);
```
eddynaka, line 94: Added a `.` making it a complete sentence.

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageAnalyzerResultProvider.cs
sunandabalu: It works in the scenario of PackageAnalysis but likely won't for DiagnosticAnalysis as the the Rule Name is taken from the Diagnostic Analyzers.

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageAnalyzerResultProvider.cs
sunandabalu: We do need to provide this information to the users that there are no updates needed for them in the current analysis. That is a customer requirement.

src/components/Microsoft.DotNet.UpgradeAssistant.Analysis/AnalyzeResultWriter.cs
```diff
@@ -49,19 +49,38 @@ await foreach (var ar in analyzeResultDefinitions)
...
                         },
                     },
                     Results = ExtractResults(analyzeResults),
                 };
             }
         }
 
+        private static ReportingDescriptor ExtractRule(IGrouping<string, AnalyzeResult> analyzeResults)
+        {
+            var analyzeResult = analyzeResults.First();
```
sunandabalu, line 63: Why only `.First()`? We do need all the distinct rules to be assigned to `Rules`.

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageAnalyzerResultProvider.cs
```diff
@@ -16,16 +16,19 @@ namespace Microsoft.DotNet.UpgradeAssistant.Steps.Packages
 {
     public class PackageAnalyzerResultProvider : IAnalyzeResultProvider
     {
-        private const string Id = "UA101";
+        private const string RuleId = "UA101";
+        private const string RuleName = "SomeRuleName";
+        private const string FullDescription = "Some full description";
+        private readonly Uri _helpUri = new("about:blank");
```
sunandabalu, line 22: For now the RuleName and Full Description can be the same as ToolName and the helpUri can be the same as InformationUri.

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageAnalyzerResultProvider.cs
```diff
@@ -37,9 +40,9 @@ public class PackageAnalyzerResultProvider : IAnalyzeResultProvider
             _analysisState = null;
         }
 
-        public async Task<bool> IsApplicableAsync(AnalyzeContext analysis, CancellationToken token)
+        public Task<bool> IsApplicableAsync(AnalyzeContext analysis, CancellationToken token)
         {
-            return await Task.FromResult(true);
+            return Task.FromResult(true);
```
sunandabalu, line 45: Why the change to remove async and await from here?

src/components/Microsoft.DotNet.UpgradeAssistant.Analysis/AnalyzeResultWriter.cs
eddynaka: this method would receive the group itself, so, it's already grouped. Check line 53.

With that, all elements have the same ruleId, so, we can just get the  first since all metadata will be the same.

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/WinformsResultProvider.cs
```diff
@@ -16,14 +16,18 @@ namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Windows
 {
     public class WinformsResultProvider : IAnalyzeResultProvider
     {
-        private ILogger Logger { get; }
+        private const string RuleId = "UA102";
+        private const string RuleName = "SomeRuleName";
+        private const string FullDescription = "Some full description";
```
sunandabalu, line 21: Same here too

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageAnalyzerResultProvider.cs
eddynaka: we either need to return the task itself or use configureAwait.
since it's a simple true value, you don't need to add await here since the call is normally awaited.

I can revert if you think the other way was better.

src/components/Microsoft.DotNet.UpgradeAssistant.Analysis/AnalyzeResultWriter.cs
sunandabalu: Have you tested this with scenarios where multiple distinct rules are present like in case of `API Upgradability` section?

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageAnalyzerResultProvider.cs
sunandabalu: Yeah lets put it back the way it was, and focus this PR to be only SARIF log improvements. The `IsApplicableAsync` pattern is used in many areas in UA, change here needs to be reflected everywhere, and I would prefer it to be a separate PR.

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Commands/Analyze/ConsoleAnalyze.cs
sunandabalu: What is the need for changing the variable to be `ToolName`? The Name here is associated with the Provider, which we are repurposing to be the Tool Name in SARIF log, the member name here should not matter from the SARIF perspective. Can you revert this back to Name?

src/components/Microsoft.DotNet.UpgradeAssistant.Analysis/AnalyzeResultWriter.cs
```diff
@@ -49,19 +49,38 @@ await foreach (var ar in analyzeResultDefinitions)
...
+
+            var rule = new ReportingDescriptor
+            {
+                Id = analyzeResult.RuleId,
+                HelpUri = analyzeResult.HelpUri
+            };
+
+            if (string.IsNullOrWhiteSpace(analyzeResult.FullDescription))
+            {
+                rule.FullDescription = new() { Text = analyzeResult.RuleName };
```
sunandabalu, line 73: this if is not setting the rule.Name

eddynaka: it is by design.
Rules generated by Roslyn does not have a specific RuleName, with that, i'm keeping the old behavior.

src/components/Microsoft.DotNet.UpgradeAssistant.Analysis/AnalyzeResultWriter.cs
```diff
@@ -49,19 +49,38 @@ await foreach (var ar in analyzeResultDefinitions)
...
+            }
+            else
+            {
+                rule.Name = analyzeResult.RuleName;
+                rule.FullDescription = new() { Text = analyzeResult.FullDescription };
+            }
+
```
sunandabalu, line 80: instead you can just do `rule.FullDescription = new() { Text = !string.IsNullOrWhiteSpace(analyzeResult.FullDescription) ? analyzeResult.FullDescription : analyzeResult.RuleName };` no need of the `if..else`

sunandabalu (changes requested):

src/components/Microsoft.DotNet.UpgradeAssistant.Analysis/AnalyzeResultWriter.cs
sunandabalu: ah I forgot that we are not updating the roslyn parts here.

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageAnalyzerResultProvider.cs
```diff
@@ -115,8 +120,10 @@ void GetResults<T>(string name, string action, IReadOnlyCollection<Operation<T>>
                             {
                                 results.UnionWith(s.OperationDetails.Details.Select(s => new AnalyzeResult()
                                 {
-                                    RuleId = Id,
-                                    RuleName = this.Name,
+                                    RuleId = RuleId,
+                                    RuleName = RuleName,
+                                    FullDescription = FullDescription,
+                                    HelpUri = _helpUri,
```
sunandabalu, line 126: Maybe we can split each of these operations into a separate rule, that way they will be grouped under the category. For eg: `Packages to be added` is a rule, `Packages to be deleted` is a separate rule so on and so forth. Is that doable here?

sunandabalu: That way each of these rules will have their own custom Rule id, name and description.

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/WinformsResultProvider.cs
```diff
@@ -78,8 +82,10 @@ public async IAsyncEnumerable<AnalyzeResult> AnalyzeAsync(AnalyzeContext analysi
                 {
                     yield return new()
                     {
-                        RuleId = this._id,
-                        RuleName = this.Name,
+                        RuleId = RuleId,
+                        RuleName = RuleName,
+                        HelpUri = _helpUri,
+                        FullDescription = FullDescription,
```
sunandabalu, line 88: The ResultMessage comes from IUpdater for eg - https://github.com/dotnet/upgrade-assistant/blob/main/src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/WinformsDefaultFontUpdater.cs - the title, description and message is defined here. I think these needs to be treated similar to diagnostic analyzers and each of the updater implementation should provide the info on RuleId, RuleName, helpURI and FullDescription

src/components/Microsoft.DotNet.UpgradeAssistant.Analysis/AnalyzeResultWriter.cs
eddynaka: I removed the else, but we will still need the if.
Reason:
- FullDescription can be either the FullDescription or RuleName
- RuleName will only be RuleName if FullDescription is not null/empty (if fullDescription is null/empty, it means it is still using the default RuleName, which can contain multiple parts, spaces, etc.)

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Commands/Analyze/ConsoleAnalyze.cs
eddynaka: sure. just did.

src/components/Microsoft.DotNet.UpgradeAssistant.Analysis/AnalyzeResultWriter.cs
eddynaka: Just updated a test to verify.

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageAnalyzerResultProvider.cs
eddynaka: just updated.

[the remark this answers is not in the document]
eddynaka: just updated.

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/WinformsResultProvider.cs
```diff
@@ -16,14 +16,18 @@ namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Windows
 {
     public class WinformsResultProvider : IAnalyzeResultProvider
     {
-        private ILogger Logger { get; }
+        private const string RuleId = "UA102";
+        private const string RuleName = "SomeRuleName";
+        private const string FullDescription = "Some full description";
 
+        private readonly Uri _helpUri = new("about:blank");
```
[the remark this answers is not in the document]
eddynaka: just updated.

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageAnalyzerResultProvider.cs
eddynaka: SARIF contains one thing that we call SubId.
RuleId = RULE0001
SubId = Rule0001/Update
SubId = Rule0001/Delete
SubId = Rule0001/Create

The same RuleId contains 3 specific formats (update, create, and delete). This is useful when people wants to suppress the specific format (for example, only suppressing the delete option).

If you have 3 different guidance, then we should create 3 separate rules (rule0001, rule0002, rule0003). If all will point to the same guidance, then, we could create just one rule.

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/WinformsResultProvider.cs
eddynaka: I think we should change the `IUpdater` in a separate PR.
Otherwise, we will change many files.

What do you think?

src/components/Microsoft.DotNet.UpgradeAssistant.Analysis/AnalyzeResultWriter.cs
sunandabalu: makes sense. That way when we add FullDescription to all analyzers we can just remove the if here. Could you add a comment here so we will know this is temporary and will be removed in the subsequent PRs?

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/WinformsResultProvider.cs
twsouthwick: Yeah, that'll be a bigger change and would be good in a separate PR

twsouthwick (commented):
This looks like a great change! Thanks for improving the experience here :)

eddynaka:
> This looks like a great change! Thanks for improving the experience here :)

Sure, no problem.
Just moved to ready to review.

Once we merge this, we can focus on the next "type".

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/WinformsResultProvider.cs
sunandabalu: agreed, we can table the changes to IUpdater as part of the changes to Diagnostic Analyzer. Could we make this PR just for the Dependency Analysis section then @eddynaka ? Could we revert back what we had for the `WinformsResultProvider.cs`?

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageAnalyzerResultProvider.cs
sunandabalu: SubId seems like right choice here, is it possible to make the change here for this PR? @eddynaka . Lets sync offline to design this a little better.

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/WinformsResultProvider.cs
eddynaka: sure.

eddynaka: done.

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageAnalyzerResultProvider.cs
```diff
@@ -16,7 +16,10 @@ namespace Microsoft.DotNet.UpgradeAssistant.Steps.Packages
 {
     public class PackageAnalyzerResultProvider : IAnalyzeResultProvider
     {
-        private const string Id = "UA101";
+        private const string RuleId = "UA101";
+        private const string RuleName = "Dependency Analysis";
```
twsouthwick, line 20: Seems a little weird to have the rule name have a space. Are there any recommended (or required) constraints for rule names?

twsouthwick (approved):

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageAnalyzerResultProvider.cs
eddynaka: For now, I'm keeping the same ruleName as we had before. But, for the tools that we own (BinSkim, SARIF-Pattern-Matcher, etc.), the RuleName is only one word: "DoNotUseDeprecatedPackage" for example.

twsouthwick: @sunandabalu we should come up with a "best practices" for defining these. Probably building on what the tools that @eddynaka is referring to does

eddynaka: The only constraint in the spec is:
https://docs.oasis-open.org/sarif/sarif/v2.1.0/csprd01/sarif-v2.1.0-csprd01.html#_Toc10541281

```
A reportingDescriptor object MAY contain a property named name whose value is a localizable string (§3.5.1) containing an identifier that is understandable to an end user. If the name of a rule contains implementation details that change over time, a tool author might alter a rule's name (while leaving the stable id property (§3.49.3) unchanged).

NOTE: A rule name is suitable in contexts where a readable identifier is preferable and where the lack of stability is not a concern.

EXAMPLE: "name": "SpecifyMarshalingForPInvokeStringArguments"
```

eddynaka: Here you can also find some guidance that we created:
https://github.com/microsoft/sarif-tutorials/blob/main/docs/Authoring-rule-metadata-and-result-messages.md#human-readable-identifier

twsouthwick: Created a new issue for us to track this https://github.com/dotnet/upgrade-assistant/issues/943

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageAnalyzerResultProvider.cs
```diff
@@ -90,22 +93,24 @@ private HashSet<AnalyzeResult> ExtractAnalysisResult(string fileLocation, IDepen
...
-
-                void GetResults<T>(string name, string action, IReadOnlyCollection<Operation<T>> collection)
+                GetResults("UA103", "ReferenceShouldBeDeleted", "Reference to ", " needs to be deleted.", analysisState.References.Deletions);
+                GetResults("UA104", "ReferenceShouldBeAdded", "Reference to ", " needs to be added.", analysisState.References.Additions);
+                GetResults("UA105", "PackageShouldBeDeleted", "Package ", " needs to be deleted.", analysisState.Packages.Deletions);
+                GetResults("UA106", "PackageShouldBeAdded", "Package ", " needs to be added.", analysisState.Packages.Additions);
+                GetResults("UA107", "FrameworkShouldBeDeleted", "Framework Reference to ", " needs to be deleted.", analysisState.FrameworkReferences.Deletions);
+                GetResults("UA108", "FrameworkShouldBeAdded", "Framework Reference to ", " needs to be added.", analysisState.FrameworkReferences.Additions);
+
+                void GetResults<T>(string ruleId, string ruleName, string name, string action, IReadOnlyCollection<Operation<T>> collection)
```
eddynaka, line 113: Added ruleId/ruleName.
Let me know what do you think!
Thank you

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageAnalyzerResultProvider.cs
```diff
@@ -90,22 +93,24 @@ private HashSet<AnalyzeResult> ExtractAnalysisResult(string fileLocation, IDepen
...
-                GetResults("Framework Reference to ", " needs to be added ", analysisState.FrameworkReferences.Additions);
-
-                void GetResults<T>(string name, string action, IReadOnlyCollection<Operation<T>> collection)
+                GetResults("UA103", "ReferenceShouldBeDeleted", "Reference to ", " needs to be deleted.", analysisState.References.Deletions);
+                GetResults("UA104", "ReferenceShouldBeAdded", "Reference to ", " needs to be added.", analysisState.References.Additions);
+                GetResults("UA105", "PackageShouldBeDeleted", "Package ", " needs to be deleted.", analysisState.Packages.Deletions);
+                GetResults("UA106", "PackageShouldBeAdded", "Package ", " needs to be added.", analysisState.Packages.Additions);
+                GetResults("UA107", "FrameworkShouldBeDeleted", "Framework Reference to ", " needs to be deleted.", analysisState.FrameworkReferences.Deletions);
+                GetResults("UA108", "FrameworkShouldBeAdded", "Framework Reference to ", " needs to be added.", analysisState.FrameworkReferences.Additions);
```
sunandabalu, line 111: This should be "FrameworkReferenceToBeAdded" and "FrameworkReferencesToBeDeleted" and for the rest to lets make it "ToBe" instead of "ShouldBe". Other than that LGTM

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageAnalyzerResultProvider.cs
sunandabalu: We synced offline and are going with separate ruleids here

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageAnalyzerResultProvider.cs
```diff
@@ -115,8 +120,10 @@ void GetResults<T>(string name, string action, IReadOnlyCollection<Operation<T>>
                             {
                                 results.UnionWith(s.OperationDetails.Details.Select(s => new AnalyzeResult()
                                 {
-                                    RuleId = Id,
-                                    RuleName = this.Name,
+                                    RuleId = ruleId,
+                                    RuleName = ruleName,
+                                    FullDescription = FullDescription,
```
sunandabalu, line 125: Oh the FullDescription should also be passed to the method similar to ruleId and ruleName. For now the helpUrl can just be the common one.

eddynaka: what do u suggest for fulldescription?

sunandabalu: "References that need to be deleted in order to upgrade the project to chosen TFM" similarly for the other areas accordingly "Packages" or "Framework References" and "added" or "deleted" .

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageAnalyzerResultProvider.cs
eddynaka: changed!

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageAnalyzerResultProvider.cs
eddynaka: changed!

sunandabalu (approved):
LGTM, Thanks @eddynaka!

PR #944, opened 2021-11-17
Add search command for extensions

twsouthwick:


src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/UpgradeAssistantHostExtensions.cs
```diff
@@ -183,13 +187,31 @@ private static void AddUserInput(this IServiceCollection services)
                     services.AddSingleton(parseResult);
                     services.AddTransient<IUpgradeStartup, UsedCommandTelemetry>();
                 })
-                .UseSerilog((_, __, loggerConfiguration) => loggerConfiguration
+                .UseSerilog((context, __, loggerConfiguration) => loggerConfiguration
                     .Enrich.FromLogContext()
-                    .MinimumLevel.Is(Serilog.Events.LogEventLevel.Verbose)
+                    .MinimumLevelFromConfiguration(context.Configuration.GetSection("Logging:Loglevel"))
+                    .MinimumLevel.Override("System", LogEventLevel.Warning)
```
sunandabalu, line 193: Why is this needed?

sunandabalu (commented):
Can we add some tests for the search and looks like even PackageCreator

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/UpgradeAssistantHostExtensions.cs
sunandabalu: Is this  to handle something along the lines of what I'm seeing here - https://github.com/dotnet/upgrade-assistant/pull/938/checks?check_run_id=4255975097 -  when I changed the .NET SDK version to `6.0.100`?

twsouthwick: No, that's unrelated.

This is about using HttpClient emits informational stuff about what's going on - we don't need to show that

src/extensions/nuget/Microsoft.DotNet.UpgradeAssistant.Extensions.NuGet/PackageLoader.cs
```diff
@@ -455,10 +455,14 @@ public bool CreateArchive(IExtensionInstance extension, Stream stream)
...
                 Version = NuGetVersion.Parse(extension.Version.ToString()),
-                Description = extension.Description,
-                PackageTypes = new[] { new PackageType("UpgradeAssistantExtension", new Version(1, 0, 0)) }
+                Description = extension.Description
             };
 
+            if (packageType is not null)
+            {
+                builder.PackageTypes = new[] { new PackageType(packageType, new Version(1, 0, 0)) };
+            }
```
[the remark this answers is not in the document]
twsouthwick: No, this is v1 of the package type. We don't want that configureable because that's for us to manage what it means. They can define their own version the same way they always did

PR #950, opened 2021-11-18
Improving SARIF for IUpdater rules

eddynaka:

**PR Title**
Improving SARIF for IUpdater rules

**PR Description**
In this PR I'm proposing some changes to IUpdater rules so we can enhance the emitted SARIF.


src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/WinformsUpdaterResult.cs
```diff
@@ -5,7 +5,13 @@
...
 namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Windows
 {
-    public record WinformsUpdaterResult(bool Result, string Message, IList<string> FileLocations) : IUpdaterResult
+    public record WinformsUpdaterResult(
+        string RuleId,
+        string RuleName,
+        string FullDescription,
+        bool Result,
+        string Message,
+        IList<string> FileLocations) : IUpdaterResult
```
eddynaka, line 14: Are there cases where we don't have a message or FileLocations?
I was thinking in adding everything to the IUpdaterResult and just use the interface properties instead of casting to WindowsUpdaterResult in the WindowsDefaultFontUpdater, for example

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IUpdaterResult.cs
```diff
@@ -5,6 +5,12 @@ namespace Microsoft.DotNet.UpgradeAssistant
 {
     public interface IUpdaterResult
     {
+        string RuleId { get; }
+
+        string RuleName { get; }
+
+        string FullDescription { get; }
```
eddynaka, line 12: I was wondering about the HelpUri. Should I already include it here?

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IUpdaterResult.cs
```diff
@@ -5,6 +5,12 @@ namespace Microsoft.DotNet.UpgradeAssistant
 {
     public interface IUpdaterResult
```
twsouthwick, line 6: @sunandabalu should we just change this to a record? I'm not seeing any behavior that would need to be done through an interface?

eddynaka: and that would facilitate a lot  :)

sunandabalu: IUpdaterResult is implemented differently by updaters other than Winforms Updaters. If we make this a record then does that mean we make separate ones for each of the updaters? I guess the ruleid , name and full description is something that might apply to all updaters as we can pull the data into ResultProviders.

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/WinformsUpdaterResult.cs
sunandabalu: Yeah there are multiple source updaters under Windows extensions and I think they all use the same UpdaterResult and can have FileLocations empty.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IUpdaterResult.cs
eddynaka: Changed a few more places so you can see what would change/happen.
We have to check the ruleId values that you want to use.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IUpdaterResult.cs
twsouthwick: I think that would make sense. Make it nullable and then it can just not be implemented

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/WinformsUpdaterResult.cs
twsouthwick: Does it have to be a list? If so, I think I'd prefer an `IReadOnlyList<>` or `ImmutableArray` (something that cannot be changed)

twsouthwick (approved):
This direction looks fine to me. I'm still thinking we should just get rid of the interface and just have a UpdaterResult with potential nullable fields.

sunandabalu (approved):
Overall, LGTM.

eddynaka:
> Overall, LGTM.

thanks!
now, each rule has its own ruleid/rulename/description.
This will enhance the sarif.

let me know if you have any other concern. if you, feel free to merge it.

PR #955, opened 2021-11-24
Fix TypeUpgradeAnalyzer to not report on non-type syntax or syntax with incorrect FQN

mjrousos:
Addresses #931

### DIFF
--- a/src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/SyntaxNodeExtensions.cs
+++ b/src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/SyntaxNodeExtensions.cs
@@ -303,5 +303,120 @@ namespace Microsoft.DotNet.UpgradeAssistant.Extensions.Default
                 LanguageNames.VisualBasic => StringComparer.OrdinalIgnoreCase,
                 _ => throw new NotImplementedException(Resources.UnknownLanguage),
             };
+
+        /// <summary>
+        /// Attempts to determine if a given syntax node corresponds to a type symbol
+        /// based on the node's usage by its parent. This is a best guess in cases where
+        /// semantic information is not available.
+        /// </summary>
+        /// <remarks>
+        /// Checking the symbol in the semantic model is preferred and this method
+        /// should only be used when not symbol information is available.
+        /// </remarks>
+        /// <param name="node">The syntax node to investigate.</param>
+        /// <returns>True if it appears the syntax node is used in a context where a type
+        /// symbol would usually make sense.</returns>
+        public static bool IsTypeSyntax(this SyntaxNode? node)
+        {
+            var parentNode = node?.Parent;
+
+            if (parentNode is null)
+            {
+                return false;
+            }
+
+            // Object creation
+            if (parentNode.IsKind(CS.SyntaxKind.ObjectCreationExpression) || parentNode.IsKind(VB.SyntaxKind.ObjectCreationExpression))
+            {
+                return true;
+            }
+
+            // Static member access
+            if ((parentNode is CSSyntax.MemberAccessExpressionSyntax csMAE && csMAE.Expression.IsEquivalentTo(node))
+                || (parentNode is VBSyntax.MemberAccessExpressionSyntax vbMAE && vbMAE.Expression.IsEquivalentTo(node)))
+            {
+                return true;
+            }
+
+            // Base type
+            if (parentNode.IsBaseTypeSyntax())
+            {
+                return true;
+            }
+
+            // Generic parameter
+            if (parentNode.IsKind(CS.SyntaxKind.TypeArgumentList) || parentNode.IsKind(VB.SyntaxKind.TypeArgumentList))
+            {
+                return true;
+            }
+
+            // Attribute
+            if (parentNode.IsKind(CS.SyntaxKind.Attribute) || parentNode.IsKind(VB.SyntaxKind.Attribute))
+            {
+                return true;
+            }
+
+            // Array
+            if (parentNode.IsKind(CS.SyntaxKind.ArrayType) || parentNode.IsKind(VB.SyntaxKind.ArrayType))
+            {
+                return true;
+            }
+
+            // Variable type
+            if ((parentNode is CSSyntax.VariableDeclarationSyntax csDecl && csDecl.Type.IsEquivalentTo(node))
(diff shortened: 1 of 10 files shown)


src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/SyntaxNodeExtensions.cs
```diff
@@ -303,5 +303,120 @@ public static StringComparer GetStringComparer(this SyntaxNode? node)
...
+            }
+
+            // Object creation
+            if (parentNode.IsKind(CS.SyntaxKind.ObjectCreationExpression) || parentNode.IsKind(VB.SyntaxKind.ObjectCreationExpression))
+            {
+                return true;
+            }
+
+            // Static member access
+            if ((parentNode is CSSyntax.MemberAccessExpressionSyntax csMAE && csMAE.Expression.IsEquivalentTo(node))
```
twsouthwick, line 335: I think we have some generalized syntax helpers that may simplify this pattern and allow for language agnostic

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/SyntaxNodeExtensions.cs
```diff
@@ -303,5 +303,120 @@ public static StringComparer GetStringComparer(this SyntaxNode? node)
...
+            }
+
+            // VB cast expression
+            if (parentNode is VBSyntax.CastExpressionSyntax vbCastExpression
+                && vbCastExpression.Type.IsEquivalentTo(node))
+            {
+                return true;
+            }
+
+            return false;
```
twsouthwick, line 419: I can't think of any others, but we'll just need to keep an eye out for additional type-like syntaxes :)

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/TypeUpgradeAnalyzer.cs
```diff
@@ -133,14 +133,28 @@ private static void AnalyzeIdentifier(SyntaxNodeAnalysisContext context, IEnumer
...
 
-            // If the identifier resolves to an actual symbol that isn't the old identifier, bail out
-            // Also refrain from adding a diagnostic if the symbol is ambiguous but a candidate symbol matches the
-            // new type. This is useful in cases where the type is referenced as a simple name and the new namespace
-            // has been added but the old namespace hasn't been removed yet.
+            var fullyQualifiedNameNode = context.Node.GetQualifiedName();
+
+            // If the identifier is part of a fully qualified name and the qualified name does not match
+            // the old name, then bail out because it is not referring to that type
+            if ((fullyQualifiedNameNode is CSSyntax.QualifiedNameSyntax || fullyQualifiedNameNode is VBSyntax.QualifiedNameSyntax)
```
twsouthwick, line 140: this would be nice to have in a separate file (whenever I see string comparisons like this, I'm betting we'll do it multiple times somewhere and it would be nice to hide it behind some logical method)

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/TypeUpgradeAnalyzer.cs
```diff
@@ -133,14 +133,28 @@ private static void AnalyzeIdentifier(SyntaxNodeAnalysisContext context, IEnumer
...
+                && !mapping.OldName.EndsWith(fullyQualifiedNameNode.ToString(), StringComparison.Ordinal))
+            {
+                return;
+            }
+
             var symbolInfo = context.SemanticModel.GetSymbolInfo(context.Node);
 
+            // If the identifier resolves to a symbol that isn't a type symbol (property symbol, for example),
+            // then bail out because this analyzer should only flag type usage.
+            if (symbolInfo.Symbol is ISymbol && symbolInfo.Symbol is not INamedTypeSymbol)
```
twsouthwick, line 150:
```suggestion
            if (symbolInfo.Symbol is ISymbol && (!symbolInfo.Symbol is INamedTypeSymbol symbol))
```

would remove the need to do another cast on line 156

twsouthwick (approved):
Seems pretty straight forward. A few comments, but otherwise LGTM

twsouthwick:
btw - I really like the comments. Some of the logic here doesn't seem obvious, so the comments help understand *why* that check is occuring

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers.Common/SyntaxNodeExtensions.cs
mjrousos: Good point. We have `IsMemberAccessExpression` in SyntaxNodeExtensions.Predicates. I'm not sure if I want to use that here, though, because in addition to checking that the parent is an MAE, I'm also checking that the node is the parent's 'expression' (as opposed to its Name) because we want to return true for `Controller.Foo` but not for `Controller.Foo`.

We could add another extension method that looks for that sort of MAE format, but I'm not sure if it's general-purpose enough to be worth extracting it that way.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/TypeUpgradeAnalyzer.cs
mjrousos: Good call. Changed.

src/extensions/default/analyzers/Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers/TypeUpgradeAnalyzer.cs
mjrousos: Agreed. I'll make an extension method in SyntaxNodeExtensions.Predicates for this.

mjrousos: Hah - just went to make the extension method and found it was already there! I'll start using the existing extension method :)

PR #993, opened 2021-12-30
Fixed typo in NuGet extension - ProjectNuGetReferences.cs

banditoth:
**Fixed typo in NuGet extension - ProjectNuGetReferences.cs**

dotnet restore can be accessed with space instead of dash


twsouthwick (approved):
Thanks @banditoth for catching that

PR #995, opened 2022-01-04
Remove Maui install

twsouthwick:
Since the Maui E2E tests aren't running and the Maui toolchain isn't actually required, this will simplify the builds and help with the overall CI run times.


Sweekriti91 (approved):
Yup! removing this should do it, that was the only place I was adding the step to install the maui components.

PR #996, opened 2022-01-05
Add solution wide project file conversion behind feature flag

twsouthwick:
This change enables a user to run conversions to SDK style as a unit of work before continuing on to additional code changes. Currently set behind a feature flag so SOLUTION_WIDE_SDK_CONVERSION must be enabled for the solution wide conversion to be used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,15 @@ When run on a solution in order to analyze dependencies prior to upgrade, the to
 
 Extensions may be managed centrally for a project as described [here](docs/design/Extension_Management.md).
 
+### Experimental features
+
+Feature flags can be used to turn some experimental features on or off. Functionality behind feature flags may or may not be made public and are considered a way to test features that may be unstable. In order to enable a feature, set an environment variable to a semi-colon delimited list of feature names. For example: `$env:UA_FEATURES="FEATURE1;FEATURE2"`.
+
+Current features that are available to try out include:
+
+- `ANALYZE_OUTPUT_FORMAT`: Enables a `--format` flag for the `analyze` command that can take other options. Currently restricted to `sarif` and `html`
+- `SOLUTION_WIDE_SDK_CONVERSION`: Switches project format conversion from old style project files to SDK style to be solution wide first before any other changes to the project files.
+
 ### Optional Features
 
 Some features within Upgrade Assistant may be enabled by adding extensions. Available extensions can be viewed on [here](https://www.nuget.org/packages?packagetype=UpgradeAssistantExtension&sortby=relevance&q=&prerel=True).
--- a/src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildWorkspaceUpgradeContext.cs
+++ b/src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildWorkspaceUpgradeContext.cs
@@ -86,16 +86,16 @@ namespace Microsoft.DotNet.UpgradeAssistant.MSBuild
             ProjectCollection.Dispose();
         }
 
-        public IProject GetOrAddProject(FileInfo path)
+        public IProject GetProject(string path)
         {
-            if (_projectCache.TryGetValue(path.FullName, out var cached))
+            if (_projectCache.TryGetValue(path, out var cached))
             {
                 return cached;
             }
 
-            var project = new MSBuildProject(this, _componentIdentifiers, _factories, _restorer, _comparer, path, _logger);
+            var project = new MSBuildProject(this, _componentIdentifiers, _factories, _restorer, _comparer, new FileInfo(path), _logger);
 
-            _projectCache.Add(path.FullName, project);
+            _projectCache.Add(path, project);
 
             return project;
         }
@@ -111,7 +111,7 @@ namespace Microsoft.DotNet.UpgradeAssistant.MSBuild
 
                 foreach (var path in _entryPointPaths)
                 {
-                    yield return GetOrAddProject(path);
+                    yield return GetProject(path.FullName);
                 }
             }
 
@@ -138,7 +138,7 @@ namespace Microsoft.DotNet.UpgradeAssistant.MSBuild
                     }
                     else
                     {
-                        yield return GetOrAddProject(new FileInfo(project.FilePath));
+                        yield return GetProject(project.FilePath);
                     }
                 }
             }
@@ -223,7 +223,7 @@ namespace Microsoft.DotNet.UpgradeAssistant.MSBuild
                     return null;
                 }
 
-                return GetOrAddProject(_projectPath);
+                return GetProject(_projectPath.FullName);
             }
         }
 
--- a/src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildWorkspaceUpgradeContext.cs
+++ b/src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildWorkspaceUpgradeContext.cs
@@ -88,14 +88,17 @@ namespace Microsoft.DotNet.UpgradeAssistant.MSBuild
 
         public IProject GetProject(string path)
         {
-            if (_projectCache.TryGetValue(path, out var cached))
+            // Normalize to ensure we're always getting the same path
+            var normalizedPath = Path.GetFullPath(path);
+
+            if (_projectCache.TryGetValue(normalizedPath, out var cached))
             {
                 return cached;
             }
 
-            var project = new MSBuildProject(this, _componentIdentifiers, _factories, _restorer, _comparer, new FileInfo(path), _logger);
+            var project = new MSBuildProject(this, _componentIdentifiers, _factories, _restorer, _comparer, new FileInfo(normalizedPath), _logger);
 
-            _projectCache.Add(path, project);
+            _projectCache.Add(normalizedPath, project);
 
             return project;
         }
--- a/src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.ProjectFormat/SdkStyleConversionSolutionWideStep.cs
+++ b/src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.ProjectFormat/SdkStyleConversionSolutionWideStep.cs
@@ -1,0 +1,98 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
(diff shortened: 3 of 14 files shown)


README.md
```diff
@@ -46,6 +46,15 @@ When run on a solution in order to analyze dependencies prior to upgrade, the to
 
 Extensions may be managed centrally for a project as described [here](docs/design/Extension_Management.md).
 
+### Experimental features
+
+Feature flags can be used to turn some experimental features on or off. Functionality behind feature flags may or may not be made public and are considered a way to test features that may be unstable. In order to enable a feature, set an environment variable to a semi-colon delimited list of feature names. For example: `$env:UA_FEATURES="FEATURE1;FEATURE2"`.
+
+Current features that are available to try out include:
+
+- `ANALYZE_OUTPUT_FORMAT`: Enables a `--format` flag for the `analyze` command that can take other options. Currently restricted to `sarif` and `html`
```
twsouthwick, line 55: @bleaphar this renames the feature flag so it's a bit more self-descriptive. Thoughts?

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.ProjectFormat/TryConvertProjectConverterStep.cs
```diff
@@ -36,61 +31,26 @@ public class TryConvertProjectConverterStep : UpgradeStep
...
-            }
-            else
-            {
-                Logger.LogInformation("Project file converted successfully! The project may require additional changes to build successfully against the new .NET target.");
-                return new UpgradeStepApplyResult(UpgradeStepStatus.Complete, "Project file converted successfully");
-            }
-        }
-
-        protected override async Task<UpgradeStepInitializeResult> InitializeImplAsync(IUpgradeContext context, CancellationToken token)
+        protected override Task<UpgradeStepInitializeResult> InitializeImplAsync(IUpgradeContext context, CancellationToken token)
```
bleaphar, line 53: Why is this no longer async?

twsouthwick: It is still async. However, because the method doesn't await anymore, it can just pass on the task from the method it calls

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Solution/Microsoft.DotNet.UpgradeAssistant.Steps.Solution.csproj
```diff
@@ -3,6 +3,7 @@
     <TargetFramework>netstandard2.0</TargetFramework>
   </PropertyGroup>
   <ItemGroup>
+    <ProjectReference Include="..\..\..\common\Microsoft.DotNet.UpgradeAssistant.Abstractions.Internal\Microsoft.DotNet.UpgradeAssistant.Abstractions.Internal.csproj" />
```
mjrousos, line 6: What was this needed for? Feature flags or WorkspaceOptions, I expect, but I didn't see where the solution step used those.

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.ProjectFormat/SdkStyleConversionSolutionWideStep.cs
```diff
@@ -0,0 +1,98 @@
...
+    {
+        private readonly TryConvertRunner _runner;
+
+        private IEnumerable<UpgradeStep>? _substeps;
+
+        public override string Description => $"Use the try-convert tool ({_runner.Path}{_runner.VersionString}) to convert the projects in a solution";
+
+        public override string Title => $"Convert projects in a solution to SDK style";
+
+        public override string Id => WellKnownStepIds.TryConvertProjectConverterStepId;
```
mjrousos, line 24: This is interesting. My first reaction was that this should have a unique ID, but I think sharing is actually better in this case since this step is 'subbing in' for the old one (maybe replacing it eventually?) and other steps that depended on the old one can just depend on this transparently without needing the complication of checking feature flags, etc.

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.ProjectFormat/SdkStyleConversionSolutionWideStep.cs
```diff
@@ -0,0 +1,98 @@
...
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            if (_substeps is null)
+            {
+                _substeps = context.EntryPoints.PostOrderTraversal(t => t.ProjectReferences)
+                    .Select(p => new ProjectSdkConversionStep(_runner, p.FileInfo, Logger))
+                    .ToList();
+
+                return Task.FromResult(new UpgradeStepInitializeResult(UpgradeStepStatus.Incomplete, "Some projects may need to be converted to SDK style.", BuildBreakRisk.High));
```
mjrousos, line 61: This should check the status of the sub-steps. If all the projects are already SDK-style then this is complete and can be skipped.

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.ProjectFormat/SdkStyleConversionSolutionWideStep.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        public SdkStyleConversionSolutionWideStep(
+            TryConvertRunner runner,
+            ILogger<SdkStyleConversionSolutionWideStep> logger)
+            : base(logger)
+        {
+            _runner = runner ?? throw new ArgumentNullException(nameof(runner));
+        }
+
+        protected override Task<bool> IsApplicableImplAsync(IUpgradeContext context, CancellationToken token) => Task.FromResult(true);
```
mjrousos, line 44: This is probably ok, but our other solution-wide steps (project selection, etc.) only return true if `context.CurrentProject` is null so that the step doesn't show up once the user is working on a specific project.

mjrousos (approved):
LGTM; I left a few non-blocking comments in-line.

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Solution/Microsoft.DotNet.UpgradeAssistant.Steps.Solution.csproj
twsouthwick: I moved the graph traversal stuff there

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildWorkspaceUpgradeContext.cs
```diff
@@ -86,16 +86,16 @@ public void Dispose()
...
+            if (_projectCache.TryGetValue(path, out var cached))
             {
                 return cached;
             }
 
-            var project = new MSBuildProject(this, _componentIdentifiers, _factories, _restorer, _comparer, path, _logger);
+            var project = new MSBuildProject(this, _componentIdentifiers, _factories, _restorer, _comparer, new FileInfo(path), _logger);
 
-            _projectCache.Add(path.FullName, project);
+            _projectCache.Add(path, project);
```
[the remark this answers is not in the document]
twsouthwick: that makes sense. I've added a call to `Path.GetFullPath` that seems to be the best API for this. We haven't had issues, but we should probably swap it to something opaque that is consistent like a `ProjectId` that may contain its path....

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.ProjectFormat/TryConvertRunner.cs
```diff
@@ -0,0 +1,119 @@
...
+
+namespace Microsoft.DotNet.UpgradeAssistant.Steps.ProjectFormat
+{
+    public class TryConvertRunner
+    {
+        private readonly ITryConvertTool _runner;
+        private readonly IPackageRestorer _restorer;
+        private readonly ILogger _logger;
+
+        public TryConvertRunner(ITryConvertTool runner, IPackageRestorer restorer, ILogger<TryConvertRunner> logger)
```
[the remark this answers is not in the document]
twsouthwick: done

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.ProjectFormat/TryConvertRunner.cs
```diff
@@ -0,0 +1,119 @@
...
+                        $"Project {projectFile.FilePath} is not an SDK project. Applying this step will convert the project to SDK style.",
+                        components.HasFlag(ProjectComponents.AspNetCore) ? BuildBreakRisk.High : BuildBreakRisk.Medium);
+                }
+                else
+                {
+                    _logger.LogDebug("Project {ProjectPath} already targets SDK {SDK}", projectFile.FilePath, projectFile.Sdk);
+                    return new UpgradeStepInitializeResult(UpgradeStepStatus.Complete, $"Project already targets {projectFile.Sdk} SDK", BuildBreakRisk.None);
+                }
+            }
+#pragma warning disable CA1031 // Do not catch general exception types
```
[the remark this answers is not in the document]
twsouthwick: I don't see any need to catch it here - we can catch anything at a higher level.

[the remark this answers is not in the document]
twsouthwick: removed

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.ProjectFormat/SdkStyleConversionSolutionWideStep.cs
twsouthwick: exactly why I did it this way. It's conceptually the same thing and is mutually exclusive with the other.

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.ProjectFormat/SdkStyleConversionSolutionWideStep.cs
twsouthwick: ok. I've updated to follow that pattern

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.ProjectFormat/SdkStyleConversionSolutionWideStep.cs
twsouthwick: done

PR #1001, opened 2022-01-06
Move transitive reference analyzer to step after other analyzers

twsouthwick:
The transitive dependency analyzer often seems to cause problems with loops. This change moves it so that it occurs after all the analyzers are run by moving it into its own step.


src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Extensions.Default/DefaultExtensionServiceProvider.cs
```diff
@@ -74,7 +74,6 @@ private static void AddPackageReferenceAnalyzers(IServiceCollection services)
         {
             // Add package analyzers (note that the order matters as the analyzers are run in the order registered)
             services.AddTransient<IDependencyAnalyzer, DuplicateReferenceAnalyzer>();
-            services.AddTransient<IDependencyAnalyzer, TransitiveReferenceAnalyzer>();
```
mjrousos, line 77: Instead of removing this, we could just put it last in the list and stop looping in the dependency analyzer step. I'm king of torn which approach is best, but I think just putting this analyzer last solves the problem and doesn't introduce another upgrade step, which is nice.

tests/tool/Integration.Tests/InterceptingKnownPackageLoader.cs
```diff
@@ -33,6 +34,11 @@ public Task<bool> DoesPackageSupportTargetFrameworksAsync(NuGetReference package
 
         public async Task<NuGetReference?> GetLatestVersionAsync(string packageName, IEnumerable<TargetFrameworkMoniker> tfms, PackageSearchOptions options, CancellationToken token)
         {
+            if (packageName == "ControlzEx")
+            {
+                Debugger.Break();
```
mjrousos, line 39: Debugging change that needs removed?

tests/tool/Integration.Tests/InterceptingKnownPackageLoader.cs
```diff
@@ -51,16 +57,20 @@ public Task<bool> DoesPackageSupportTargetFrameworksAsync(NuGetReference package
 
         public async Task<IEnumerable<NuGetReference>> GetNewerVersionsAsync(NuGetReference reference, IEnumerable<TargetFrameworkMoniker> tfms, PackageSearchOptions options, CancellationToken token)
         {
+            if (reference.Name == "ControlzEx")
+            {
+                Debugger.Break();
```
mjrousos, line 62: Same here

mjrousos (commented):
Good change except I wonder if we could just leave the transitive analyzer as a dependency analyzer (instead of a step), make the other changes (to stop looping), and make sure the analyzers are ordered properly. I feel like that might be a simpler solution.

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Extensions.Default/DefaultExtensionServiceProvider.cs
twsouthwick: I agree - but if someone adds an analyzer after, it's last. Maybe we could add a `OrderAttribute` that would allow some sorting....

tests/tool/Integration.Tests/InterceptingKnownPackageLoader.cs
twsouthwick: oops - that was from testing!

tests/tool/Integration.Tests/InterceptingKnownPackageLoader.cs
twsouthwick: same

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Extensions.Default/DefaultExtensionServiceProvider.cs
mjrousos: Yea, I considered that. Adding comments when we register the analyzers would almost be good enough. The challenge would be if more analyzers are added in a different extension. Hm...

I guess the options are either to have some sort of `OrderAttribute` like you suggest (a number?) or to pull it out into a separate step. Neither seems like a great solution as they both introduce some complexity. Either way is fine with me.

mjrousos (approved):

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Extensions.Default/DefaultExtensionServiceProvider.cs
twsouthwick: Added an ordering attribute and associated changes/tests for that

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Analyzers/TransitiveReferenceAnalyzer.cs
```diff
@@ -32,15 +34,34 @@ public async Task AnalyzeAsync(IProject project, IDependencyAnalysisState state,
                 throw new ArgumentNullException(nameof(state));
             }
 
-            // If the package is referenced transitively, mark for removal
-            foreach (var packageReference in state.Packages)
+            bool completed;
+            var count = 0;
+
+            // The maximum iterations should be the number of packages. This is the worst case when all the packages are brought in by referencing a single package.
+            var maxIterations = state.Packages.Count();
```
mjrousos, line 41: It's not clear to me why we need to loop here. There's a `foreach` later to check every package reference, no other analyzers will be making updates after this one, and this analyzer only removes dependencies so nothing should become transitive that wasn't at the beginning of its execution, so running the contents of this `do` loop once seems like it should be sufficient.

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/DependencyAnalyzerRunner.cs
```diff
@@ -14,15 +14,20 @@ public class DependencyAnalyzerRunner : IDependencyAnalyzerRunner
...
             ILogger<DependencyAnalyzerRunner> logger)
         {
+            if (packageAnalyzers is null)
+            {
+                throw new ArgumentNullException(nameof(packageAnalyzers));
+            }
+
             _packageRestorer = packageRestorer ?? throw new ArgumentNullException(nameof(packageRestorer));
-            _packageAnalyzers = packageAnalyzers ?? throw new ArgumentNullException(nameof(packageAnalyzers));
+            _packageAnalyzers = packageAnalyzers.OrderyByPrecedence();
```
mjrousos, line 30: Super nit-pick - `packageAnalyzers?.OrderyByPrecedence() ?? throw new ArgumentNullException(nameof(packageAnalyzers));` would be a smaller change here. Doesn't matter much, though.

mjrousos (commented):
Not sure the transitive reference analyzer needs to loop. Looks good other than that.

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/AutoApplyStep.cs
```diff
@@ -0,0 +1,57 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.DotNet.UpgradeAssistant.Steps.Packages
+{
+    public abstract class AutoApplyStep : UpgradeStep
```
mjrousos, line 11: Maybe call this AutoApplySubStep so that it's clear it's meant to be a base class for sub-steps rather than parents?

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Analyzers/TransitiveReferenceAnalyzer.cs
twsouthwick: removed

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageUpdaterStep.cs
```diff
@@ -55,126 +49,183 @@ public class PackageUpdaterStep : UpgradeStep
...
+            _subSteps = dependencyAnalyzers.OrderyByPrecedence().Select(d => new DependencyAnalyzerStep(this, restorer, d, logger)).ToList();
         }
 
         protected override Task<bool> IsApplicableImplAsync(IUpgradeContext context, CancellationToken token) => Task.FromResult(context?.CurrentProject is not null);
 
-        protected override async Task<UpgradeStepInitializeResult> InitializeImplAsync(IUpgradeContext context, CancellationToken token)
+        protected override Task<UpgradeStepInitializeResult> InitializeImplAsync(IUpgradeContext context, CancellationToken token)
         {
-            if (context is null)
+            if (_subSteps.All(d => d.IsDone))
```
mjrousos, line 64: I think I remember why parent steps often initialize their children. The UpgradeManager will initialize steps (including sub-steps) as it gets to them, but if we wait for that then the steps won't be updated when we do an `IsDone` check in the parent's `InitializeMethod`. So, I suspect that this check isn't going to work the way you expect because the children will all be uninitialized and return false for `IsDone` even if there's no action necessary for them. We'll probably need to initialize the substeps here before doing this check.

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageUpdaterStep.cs
```diff
@@ -55,126 +49,183 @@ public class PackageUpdaterStep : UpgradeStep
...
+                        return new UpgradeStepInitializeResult(UpgradeStepStatus.Complete, "No package updates needed", BuildBreakRisk.None);
+                    }
 
-                        projectFile.RemoveFrameworkReferences(_analysisState.FrameworkReferences.Deletions.Select(i => i.Item));
-                        projectFile.AddFrameworkReferences(_analysisState.FrameworkReferences.Additions.Select(i => i.Item));
+                    var steps = new List<UpgradeStep>();
 
-                        await projectFile.SaveAsync(token).ConfigureAwait(false);
-                        count++;
+                    AddSubsteps(analysis.References.Deletions, "Remove reference '{0}'", static t => t.Name, static (file, op) => file.RemoveReferences(new[] { op.Item }));
```
mjrousos, line 131: It might be simpler to just have this step make all the changes directly instead of having sub-steps for each change. I don't think users will typically need to add or remove packages one at a time instead of just applying all the changes from a given analyzer at once.

mjrousos (commented):
Looking good. A few questions/comments inline.

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageUpdaterStep.cs
twsouthwick: It's about opting into the changes and knowing. This ends up looking like this:

image

There could be a number of changes the package maps end up making and this identifies what those are.

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageUpdaterStep.cs
twsouthwick: I saw that exact behavior initially. I then added a check in the AutoApplyStep to also check the parent's children after initializing itself. Now, it will provide the expected behavior, but will still lazily initialize when the UpgradeStep gets to them and then just complete the parent if everything for the parent is now done after initializing.

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageUpdaterStep.cs
mjrousos: Ok, this seems fine for now. I'm wondering if in cases where there are many packages that need added/removed we'll want to just have them all added/removed once per analyzer. Console output could list the changes that will be made during step initialization. I agree that's not as visible as this, though.

We can try this and change in the future if we find it's too many sub-steps.

mjrousos (approved):

PR #1002, opened 2022-01-08
Fix relative urls in extensibility.md

jcharlesworthuk:
I didn't open an issue for this as it is such a small thing but the relative urls in `extensibility.md` weren't working when viewing the markdown in GitHub - as they had a mix of forward and backward slashes.  

Have updated the markdown to use forward slashes as per https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-readmes#relative-links-and-image-paths-in-readme-files


twsouthwick (approved):
Thanks! Path separators are so annoying :)

PR #1004, opened 2022-01-10
Add command to show feature flags and if they are enabled

twsouthwick:
This allows you to see what features are available. For example:

```
upgrade-assistant features --list
```

outputs:

```
[14:32:33 WRN] Features are subject to change and may be disabled in a subsequent release.
[14:32:33 INF] Features are registered in the environment variable `UA_FEATURES`. Below are available features:
[14:32:33 INF] 'ANALYZE_OUTPUT_FORMAT': False
[14:32:33 INF] 'SOLUTION_WIDE_SDK_CONVERSION': False
```


src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions.Internal/FeatureFlags.cs
```diff
@@ -25,8 +34,10 @@ private static ICollection<string> CreateFeatures()
             return new HashSet<string>(features.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries), StringComparer.OrdinalIgnoreCase);
         }
 
-        public static bool IsAnalyzeFormatEnabled => _features.Contains("ANALYZE_OUTPUT_FORMAT");
+        public static bool IsRegistered(string name) => _features.Contains(name);
+
+        public static bool IsAnalyzeFormatEnabled => _features.Contains(Output_Format);
 
-        public static bool IsSolutionWideSdkConversionEnabled => _features.Contains("SOLUTION_WIDE_SDK_CONVERSION");
+        public static bool IsSolutionWideSdkConversionEnabled => _features.Contains(SolutionWideSdkConversion);
```
sunandabalu, line 41: Instead of tightly coupling the features in here, would it make sense to have a struct for `Feature` with `Name` and `IsEnabled`? That way its easy to get features off of under feature flag as needed.

twsouthwick: Can you give an example of what you mean?

sunandabalu (approved):

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions.Internal/FeatureFlags.cs
```diff
@@ -11,6 +11,17 @@ namespace Microsoft.DotNet.UpgradeAssistant
     /// </summary>
     public static class FeatureFlags
     {
+        private const string Output_Format = "ANALYZE_OUTPUT_FORMAT";
```
sunandabalu, line 14: Nit: naming format can be the same as the other 2 below.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions.Internal/FeatureFlags.cs
sunandabalu: I'm fine with this approach for now, hence approved it. Thinking out loud about how can we make it better to onboard or offboard features, instead of hard-coding each of the features does it make sense to have them as a resource / config file? And then we can just check for `IsRegistered` to see if its enabled, instead of having to create an `Is<Feature>Enabled`.

twsouthwick: The reason I've gone with this approach is to make it so you can easily see when a feature is used to clean up any old code paths ASAP when the flag is no longer needed using the "Find all References"

sunandabalu: Ok, I see your point. I'm fine with this approach.

PR #1007, opened 2022-01-13
Implement a transitive identification service against NuGet APIs

### LINKED ISSUE #1021: Error: NuGet target in project.assets.json is still unavailable after restore. Please verify that the project has been restored.
Problem encountered on https://dotnet.microsoft.com/en-us/platform/upgrade-assistant/tutorial/install-upgrade-assistant
Operating System: windows

I'm trying to upgrade the application developed with .net core 2.0 to .NET 6 with upgrade assistance. 

Getting error while executing `upgrade-assistant upgrade <Path to csproj or sln to upgrade>` command.

Getting following error.

Screenshot 2022-01-20 165802

Tried following things:
1. Tried to restore with dotnet restore and retried to upgrade
2. Cleared nuget cache(s) from visual studio
3. Copied project.assets.json file from obj folder and pasted it at .csproj file's location

twsouthwick:
We have been using information from a lock file to identify the transitive dependencies for a project. This is problematic for our use cases in a couple of ways:

- Requires that the project file be up-to-date with the packages we want to check which may not be the case
- Requires `dotnet restore` to be run and we end up running it fairly often to ensure things are up-to-date

Fixes #1021


tests/tool/Integration.Tests/IntegrationScenarios/WpfSample/csharp/Upgraded/BeanTraderClient/BeanTraderClient.csproj
```diff
@@ -33,7 +33,12 @@
   <ItemGroup>
     <PackageReference Include="Castle.Windsor" Version="4.1.1" />
     <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="2.6.3" />
+    <PackageReference Include="Microsoft.CSharp" Version="4.7.0" />
     <PackageReference Include="Microsoft.Rest.ClientRuntime.Azure" Version="3.3.7" />
+    <PackageReference Include="System.Net.Http" Version="4.3.0" />
+    <PackageReference Include="System.ServiceModel.Duplex" Version="4.8.1" />
```
twsouthwick, line 39: Not sure why these had been removed before. I can't see anything that would have transitively brought them in. Only thing I can think of is when we were doing it before it was still against the previous target framework and may have been deemed unnecessary due to that.

mjrousos: These are all brought in by Microsoft.Windows.Compatibility. Aren't we supposed to add a reference to that since this is has a -windows TFM? After that happens, these would be transitive.

mjrousos (approved):
LGTM

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/Dependencies/TransitiveDependencyExtensions.cs
```diff
@@ -0,0 +1,149 @@
...
+                foreach (var child in current.ProjectReferences)
+                {
+                    if (visited.Add(child))
+                    {
+                        projects.Enqueue(child);
+                    }
+                }
+            }
+            while (projects.Count > 0);
+
```
sunandabalu, line 145: Is there a perf improvement to this approach compared to our previous approach of `restore` every project and its dependencies?

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IProject.cs
```diff
@@ -25,6 +25,11 @@ public interface IProject
 
         INuGetReferences NuGetReferences { get; }
 
+        /// <summary>
+        /// Gets all the PackageReference items in the project including evaluated items.
+        /// </summary>
+        IEnumerable<NuGetReference> PackageReferences { get; }
```
sunandabalu, line 31: 👍

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Microsoft.DotNet.UpgradeAssistant.Steps.Packages.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFramework>net5.0</TargetFramework>
```
sunandabalu, line 3: does it make sense to just move to 6.0 here?

src/extensions/nuget/Microsoft.DotNet.UpgradeAssistant.Extensions.NuGet/NuGetTransitiveDependencyIdentifier.cs
```diff
@@ -0,0 +1,199 @@
...
+            return new(new TargetGraphLookup(graph));
+        }
+
+        private async Task<RestoreTargetGraph?> RestoreProjectAsync(IEnumerable<NuGetReference> packages, IEnumerable<TargetFrameworkMoniker> tfms, CancellationToken token)
+        {
+            var tfmInfo = tfms.Select(tfm => new TargetFrameworkInformation { FrameworkName = NuGetFramework.Parse(tfm.ToFullString()) }).ToList();
+
+            // Create a project in a unique and temporary directory
+            var path = Path.Combine(Path.GetTempPath(), "dotnet-ua", "restores", Guid.NewGuid().ToString(), "project.txt");
+
```
sunandabalu, line 73: Since we are locally doing the restore and cleanup transitive dependencies, I'm expecting this to be much more performant than needing to do `dotnet restore`, is there a cleanup strategy for the temp folder structure created? also will this be an issue with larger projects?

sunandabalu (approved):
Overall like the approach

tests/tool/Integration.Tests/IntegrationScenarios/WpfSample/csharp/Upgraded/BeanTraderClient/BeanTraderClient.csproj
twsouthwick: Hmm not sure why it's not showing up (I mean the Microsoft.Windows.Compat doesn't seem to)

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/Dependencies/TransitiveDependencyExtensions.cs
twsouthwick: Perf and correctness. 

Perf: Not shelling out to a new process
Correctness: We can do stuff in memory instead of using values from the project file that may be out of date

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/Microsoft.DotNet.UpgradeAssistant.Steps.Packages.csproj
twsouthwick: There aren't APIs there we may want to use

src/extensions/nuget/Microsoft.DotNet.UpgradeAssistant.Extensions.NuGet/NuGetTransitiveDependencyIdentifier.cs
twsouthwick: good question. since it's in the temp path I imagine it will just be cleaned up by the os at some point. I'll create an issue to track that.

Issue #550, opened 2021-05-25
As a user, I can run Upgrade Assistant on a mac


twsouthwick:

sshquack:
Please support this for macOS. Here is the error message I get trying to upgrade from a Framework 4.8 to .net 5

```
❯ upgrade-assistant ./TaskService/TaskService.csproj
This tool is not supported on non-Windows platforms due to dependencies on Visual Studio.
```

Environment info
```
❯ dotnet --info
.NET SDK (reflecting any global.json):
 Version:   6.0.101
 Commit:    ef49f6213a

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  12.1
 OS Platform: Darwin
 RID:         osx.12-arm64
 Base Path:   /usr/local/share/dotnet/sdk/6.0.101/

Host (useful for support):
  Version: 6.0.1
  Commit:  3a25a7f1cc

.NET SDKs installed:
  6.0.101 [/usr/local/share/dotnet/sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 6.0.1 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 6.0.1 [/usr/local/share/dotnet/shared/Microsoft.NETCore.App]
```

AndiRudi:
Sadly, I just discovered the same error message. I feel everything `.net` should not rely anymore on Visual Studio.

ramedey:
+1

twsouthwick:
I'm interested in what your use case is. Can you share what project type you're coming from and where you want to go with it? 

The tool was initially designed to move people from ASP.NET to ASP. NET Core which requires VS because that's where the sdk was available. There are other workloads we're exploring that don't require VS, but that is where that limitation comes from.

AndiRudi:
> The tool was initially designed to move people from ASP.NET to ASP. NET Core which requires VS because that's where the sdk was available. There are other workloads we're exploring that don't require VS, but that is where that limitation comes from.

I am not sure what you mean by that. We have some projects written in .net (either .net framework 4.x or .net core 3.1) and I wanted an easy automated way to update them to .net 5 or 6 on my mac.

twsouthwick:
[...]

PR #1022, opened 2022-01-20
Add feature flag to enable running on non-Windows machines

twsouthwick:
This change adds a feature flag that allows someone to opt into running on non-Windows platforms. This will likely cause some problems, but will allow us to gather some information about what issues people are running into.


twsouthwick:
Part of #550

sunandabalu (approved):

PR #1026, opened 2022-01-25
Use .netstandard2.0 for default extension

twsouthwick:
We want to use .NET Standard for the extensions if possible so that we
could potentially run them in Visual Studio at some point. The reason
the default extension was not .NET Standard is because the try-convert
stuff uses MSBuild directly and was forcing .NET 5+ to be the TFM. This
change moves try-convert to be its own extension which allows the
default extension to be .NET Standard.


eng/DependencyValidation/DependencyValidation.modelproj
```diff
@@ -34,6 +34,26 @@
     </Validate>
   </ItemGroup>
   <ItemGroup>
+    <ProjectReference Include="..\..\modules\try-convert\src\MSBuild.Abstractions\MSBuild.Abstractions.csproj">
```
mjrousos, line 37: This may become clearer later, but why are all these additional references needed now?

mjrousos (approved):
Good change; thanks!

eng/DependencyValidation/DependencyValidation.modelproj
twsouthwick: Not really sure. It auto-added them and I'm not really sure how this project file works

twsouthwick: The modelproj format still uses the old style format, so......

mjrousos: Got it.

sunandabalu (approved):
What happens if we take any updates from try-convert, will any of these potentially change?

twsouthwick:
@sunandabalu not that I know of. If it were to update to target .net 6, then we'd need to react to that. The nice thing about the extensions running in their own AssemblyLoadContext is that they are fairly isolated from each other.

PR #1028, opened 2022-01-25
Reduce duplication and coupling of report writing

### LINKED ISSUE #1011: The HTML writer shouldn't be using sarif types.
The HTML writer shouldn't be using sarif types.

_Originally posted by @twsouthwick in https://github.com/dotnet/upgrade-assistant/pull/1008#discussion_r784434519_

twsouthwick:
This removes a direct dependency on sarif writer and has it searched via the provider interface. This reduces duplication of code and removes the direct dependency on it.

Fixes #1011


src/components/Microsoft.DotNet.UpgradeAssistant.Analysis/HtmlAnalyzeResultWriter.cs
```diff
@@ -0,0 +1,101 @@
...
+        {
+            if (provider is null)
+            {
+                throw new ArgumentNullException(nameof(provider));
+            }
+
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            _sarifWriter = new(() =>
+            {
+                if (provider.Value.TryGetWriter(WellKnownFormats.Sarif, out var result))
```
sunandabalu, line 34: Should this not try to get `WellKnownFormats.Html` here?

sunandabalu (approved):
Thanks for making this change, LGTM

PR #1035, opened 2022-01-28
Read evaluated packagereference versions to support CPvM

sunandabalu:
This change makes the Packages used for package analysis to use Evaluated versions in order to support Central Package Version Management using `Microsoft.Build.Evaluation`
Added `MSBuildTests` to test loading a test project with `packages.props` using `Microsoft.Build.Evaluation` and testing on evaluated package versions.


sunandabalu:
@twsouthwick I have addressed your offline feedback, could you PTAL?

tests/components/Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests/MSBuildTestHelper.cs
```diff
@@ -0,0 +1,81 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Build.Evaluation;
+
+namespace Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests
+{
+    public static class MSBuildTestHelper
```
twsouthwick, line 14: Could you make this into a class/struct that implements IDisposable and manages the state? Then cleanup/file additions will be really easy.

I'm thinking something along of the lines of:


```cs
public class MSBuildTestBuild : IDisposable // constructor will create a temp directory
{
  public void Add(string fileName, string contents); // add the contents to the directory

  public Microsoft.Build.Evaluation.Project Build(); // Load up the project

  public void Dispose(); // Clean up directory
}

File: tests/components/Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests/MSBuildTestHelper.cs [hunk 2]
```diff
@@ -0,0 +1,81 @@
...
+            return file;
+        }
+
+        public static Microsoft.Build.Evaluation.Project LoadProject(string path)
+        {
+            ProjectCollection collection = ProjectCollection.GlobalProjectCollection;
+            return collection.LoadProject(path);
+        }
+
+        public static string CreateProject(string folderName)
```
twsouthwick, line 45: How would this be customized for new tests? This is hardcoding a project file that new tests would want to define differently probably.

tests/components/Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests/MSBuildProjectTests.cs
```diff
@@ -0,0 +1,52 @@
...
+
+            // Cleanup
+            MSBuildTestHelper.CleanupProject(_folderName);
+        }
+
+        [InlineData("Newtonsoft.Json", "9.0.1")]
+        [Theory]
+        public void PackageReferenceVersionTest(string packageName, string expectedVersion)
+        {
+            // Act
```
twsouthwick, line 39:
```suggestion
            // Arrange
```

tests/components/Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests/MSBuildProjectTests.cs
```diff
@@ -0,0 +1,52 @@
...
+            var path = MSBuildTestHelper.CreateProject(_folderName);
+            var project = MSBuildTestHelper.LoadProject(path);
+            var packageReference = project.GetItems("PackageReference").FirstOrDefault();
```
twsouthwick, line 42:
```suggestion

            // Act
            var packageReference = project.GetItems("PackageReference").FirstOrDefault();
```

tests/components/Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests/MSBuildTestHelper.cs
```diff
@@ -0,0 +1,81 @@
...
+            {
+                sw.WriteLine(fileContents);
+            }
+
+            return file;
+        }
+
+        public static Microsoft.Build.Evaluation.Project LoadProject(string path)
+        {
+            ProjectCollection collection = ProjectCollection.GlobalProjectCollection;
```
twsouthwick, line 41: Wouldn't this be shared state? If so, that will potentially cause difficult to diagnose test issues.

Can we create a new project collection that isn't shared?

twsouthwick (commented):
A few comments as to the shape and test reliability/debuggability

tests/components/Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests/MSBuildTestHelper.cs
sunandabalu: Changed the method so now new tests can customize how they want the project content to be created.

tests/components/Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests/MSBuildTestHelper.cs
sunandabalu: yeah the GlobalProjectCollection is a singleton, changed to new it up instead. Good catch!

tests/components/Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests/MSBuildProjectTests.cs
```diff
@@ -0,0 +1,99 @@
...
+
+            // Assert
+            Assert.Equal(packageName, packageReference?.EvaluatedInclude);
+            Assert.Equal(expectedVersion, packageReference?.GetMetadataValue("Version"));
+
+            // Cleanup
+            msBuildTestBuild.Dispose();
+        }
+
+        public static string CreateProjectContent(MSBuildTestBuild build, string fileName, string fileContent)
```
twsouthwick, line 94: Not sure of the value of this. Why not just call `build.Add(fileName, fileContent)` at the location?

tests/components/Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests/MSBuildProjectTests.cs
```diff
@@ -0,0 +1,99 @@
...
+</Project> ");
+
+            // Act
+            var project = MSBuildTestBuild.Build(csprojPath);
+
+            // Assert
+            Assert.NotNull(project);
+
+            // Cleanup
+            msBuildTestBuild.Dispose();
```
twsouthwick, line 52: Why not call `using var msBuildTestBuild = ...`?

tests/components/Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests/MSBuildTestBuild.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Threading.Tasks;
+using Microsoft.Build.Evaluation;
+
+namespace Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests
+{
+    public sealed class MSBuildTestBuild : IDisposable
+    {
+        public string TestFolderPath { get; }
+
+        public MSBuildTestBuild(string path)
```
twsouthwick, line 18: Do tests need to care about the path? Can this be an implementation detail?

twsouthwick: It should be something that can be done both thread-safely and process safely (this is neither).

tests/components/Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests/MSBuildTestBuild.cs
```diff
@@ -0,0 +1,52 @@
...
+            if (!Directory.Exists(path))
+            {
+                Directory.CreateDirectory(path);
+            }
+        }
+
+        public string Add(string fileName, string fileContents)
+        {
+            string file = Path.Combine(TestFolderPath, fileName);
+            using (var sw = new StreamWriter(file))
```
twsouthwick, line 30: nit: Simpler to use `File.WriteAllText(...)`

tests/components/Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests/MSBuildTestBuild.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Build.Evaluation;
+
+namespace Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests
+{
+    public sealed class MSBuildTestBuild : IDisposable
```
twsouthwick, line 14: nit: I think `MSBuildTestBuilder` would make more sense since it's the "Builder" pattern it's implementing here.

twsouthwick (approved):
A few additional thoughts, but LGTM

tests/components/Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests/MSBuildTestBuild.cs
```diff
@@ -0,0 +1,52 @@
...
+            {
+                sw.WriteLine(fileContents);
+            }
+
+            return file;
+        }
+
+        public static Microsoft.Build.Evaluation.Project Build(string path)
+        {
+            using var collection = new ProjectCollection();
```
twsouthwick, line 40: not sure if you want to do a `using` here as it will dispose of things in the collection. Better probably to track that in the builder object and dispose of it with the directory (and ensure no double calling the Build method as a sanity check, since you'll only want a single collection to be tracked)

tests/components/Microsoft.DotNet.UpgradeAssistant.MSBuild.Tests/MSBuildProjectTests.cs
```diff
@@ -0,0 +1,88 @@
...
+  </ItemGroup>
+</Project> ");
+
+            msBuildTestBuild.Add("Directory.Build.targets", @"<Project>
+  <Sdk Name=""Microsoft.Build.CentralPackageVersions"" Version=""2.0.1"" />
+</Project> ");
+
+            // Act
+            var project = MSBuildTestBuilder.Build(csprojPath);
+            var packageReference = project.GetItems("PackageReference").FirstOrDefault();
```
twsouthwick, line 81: This doesn't seem to be testing the product, but that MSBuild is working. You could construct an MSBuildProject and test the changes you made

sunandabalu: https://github.com/dotnet/upgrade-assistant/issues/1038 to track adding infra to make MSBuildProject to be testable.

PR #1077, opened 2022-03-30
UWP Support

ujjwalchadha:
This PR moves the current progress being made on UWP -> Windows App SDK conversion support from the fork to an official UA branch. 

The development will continue in the branch and the code is still on prototype stage on some levels. Do not expect production level code/results yet.


sunandabalu (approved):
Approved since this is moving to a branch and not main :)

Issue #1050, opened 2022-02-14
Add more functions to IProject like RemoveItem


ujjwalchadha:
When implementing Updater<IProject>, the collection of project objects does not have enough functionality to perform operations like updating or removing an item from the project file.

I suggest exposing a more flexible object like ProjectRoot for performing operations that the current IProject helper functions do not provide and allow extensions to make use of it.

ujjwalchadha:
@sunandabalu @twsouthwick do we have any update for this?
I am looking for IProjectFile.RemoveItem method which does not seem to be present.

sunandabalu:
@ujjwalchadha Per our offline conversation, feel free to put up a PR with the change we will be happy to review it :)

PR #1100, opened 2022-04-11
Add ability to remove item from IProjectFile

ujjwalchadha:
Addresses #1050 
This PR adds RemoveItem method to IProjectFile and its implementation to MSBuildProject

**Possible breaking changes**: Any class implementing IProjectFIle outside the project (In extensions etc) might break until they implement this new method.

### DIFF
--- a/src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IProjectFile.cs
+++ b/src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IProjectFile.cs
@@ -35,7 +35,7 @@ namespace Microsoft.DotNet.UpgradeAssistant
 
         void AddItem(string name, string path);
 
-        void RemoveItem(ProjectItemType? itemType = null, string? includePath = null, string? excludePath = null, string? elementName = null);
+        bool RemoveItem(ProjectItemType? itemType = null, string? includePath = null, string? excludePath = null, string? elementName = null);
 
         bool ContainsItem(string itemName, ProjectItemType? itemType, CancellationToken token);
 
--- a/src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IProjectFile.cs
+++ b/src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IProjectFile.cs
@@ -35,6 +35,8 @@ namespace Microsoft.DotNet.UpgradeAssistant
 
         void AddItem(string name, string path);
 
+        void RemoveItem(ProjectItemType? itemType = null, string? includePath = null, string? excludePath = null, string? elementName = null);
+
         bool ContainsItem(string itemName, ProjectItemType? itemType, CancellationToken token);
 
         string GetPropertyValue(string propertyName);
--- a/src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.File.cs
+++ b/src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.File.cs
@@ -158,6 +158,39 @@ namespace Microsoft.DotNet.UpgradeAssistant.MSBuild
         public void AddItem(string name, string path)
             => ProjectRoot.AddItem(name, path);
 
+        public void RemoveItem(ProjectItemType? itemType = null, string? includePath = null, string? excludePath = null, string? elementName = null)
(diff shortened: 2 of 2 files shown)


src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IProjectFile.cs
```diff
@@ -35,6 +35,8 @@ public interface IProjectFile
 
         void AddItem(string name, string path);
 
+        bool RemoveItem(ProjectItemType? itemType = null, string? includePath = null, string? excludePath = null, string? elementName = null);
```
sunandabalu, line 38: Should the itemname that needs to be removed not be not nullable here? How else can you check what exactly needs to be removed?

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.File.cs
```diff
@@ -158,6 +158,41 @@ public void RenameFile(string filePath)
...
+                    return false;
+                }
+
+                if (elementName != null && item.ElementName != elementName)
+                {
+                    return false;
+                }
+
+                return true;
+            });
```
sunandabalu, line 186: Can we use `??` check instead of explicit `null` checks? Also why is there the need to return something on every failure? a return for `if items.any()` should suffice, no?

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.File.cs
```diff
@@ -158,6 +158,41 @@ public void RenameFile(string filePath)
...
+
+            foreach (var item in items)
+            {
+                item.Parent.RemoveChild(item);
+            }
+
```
sunandabalu, line 192: The assumption with the `RemoveItem` is to remove a item ideally a leaf node. Have you tested this with removing a nested item ie. x -> y -> z and you are trying to remove y, what happens then?

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IProjectFile.cs
ujjwalchadha: The parameters act like a filter. This method should remove all items that fit the filter.
If I wanna remove all items with a certain IncludePath, I only specify that. If I want a to remove a particular elementName and an ItemType, with that include path, I specify them too.

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.File.cs
ujjwalchadha: Noted. I'll make this change.

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.File.cs
ujjwalchadha: For now this removes everything that fit the parameters including something that potentially has a child. 
Are you recommending that we only delete an item if it is a leaf node?

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IProjectFile.cs
brandonh-msft: Perhaps we could consider a `ItemFilter` _type_ so
1. It conveys that the things you set on the object act like an `and` filter set
2. It avoids changing the method on the interface in the future if we add/remove filter capabilities

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.File.cs
brandonh-msft: Feels like this would be far more readable as a fluent set of `.Where()` clauses instead of one large one in addition to conveying the filters are additive

twsouthwick:
@ujjwalchadha is this PR still needed? The code is contained in the UWP PR as well

ujjwalchadha:
Closed because these changes are included in the uwp-support PR

PR #1104, opened 2022-04-21
Address CG alerts for UA

sunandabalu:


tests/tool/Integration.Tests/IntegrationScenarios/WebLibrary/csharp/Original/packages.config
```diff
@@ -4,5 +4,5 @@
   <package id="Microsoft.AspNet.Razor" version="3.2.7" targetFramework="net472" />
   <package id="Microsoft.AspNet.WebPages" version="3.2.7" targetFramework="net472" />
   <package id="Microsoft.Web.Infrastructure" version="1.0.0.0" targetFramework="net472" />
-  <package id="Newtonsoft.Json" version="6.0.5" targetFramework="net472" />
+  <package id="Newtonsoft.Json" version="12.0.2" targetFramework="net472" />
```
mjrousos, line 7: This is the "before" project for an end-to-end test. It's meant to have out-of-date packages that UA should update. What's the reason we need to change them?

mjrousos (changes requested):

tests/tool/Integration.Tests/IntegrationScenarios/WebLibrary/csharp/Original/packages.config
sunandabalu: CG alerting on this, and we need to upgrade the package versions to address the alert.

mjrousos: Can we upgrade to a lower version? Seeing that UA successfully replaces old versions of packages that don't work with .NET 6 with newer versions that do work is a nice feature of this test.

sunandabalu: Done, picked an older version that is not listed in CG

mjrousos (approved):
Is there any way to mark files as test code or non-shipping or something so that CG ignores them? 

These look fine for now to get things unblocked, but I am concerned that we're removing valuable test cases. One of the things UA should do is replace NuGet packages that don't support .NET 6 (like Newtonsoft v6) with newer versions that do support it. It would be good to find a way to keep old references only in test code for this purpose.

Issue #1110, opened 2022-05-04
Update recommendation for WCF services


cathysull:
Likely just a string and link change to point to CoreWCF doc or repo.

sunandabalu:
This change is merged

PR #1111, opened 2022-05-05
Update WcfServerCheck.cs

samsp-msft:
**PR Description**
Changes text when encountering an SVC file.
- [x] AKA.ms link needs to be updated once https://github.com/microsoft/dotnet-blog/pull/815 has been posted.

Addresses #1110

### DIFF
--- a/src/components/Microsoft.DotNet.UpgradeAssistant/Checks/WcfServerCheck.cs
+++ b/src/components/Microsoft.DotNet.UpgradeAssistant/Checks/WcfServerCheck.cs
@@ -16,9 +16,9 @@ namespace Microsoft.DotNet.UpgradeAssistant.Checks
     /// </summary>
     public class WcfServerCheck : IUpgradeReadyCheck
     {
-        private const string UPGRADE_LINK = "https://aka.ms/migrate-wcf-to-grpc";
+        private const string UPGRADE_LINK = "https://aka.ms/CoreWCF/migration";
         private const string CATEGORY = "WCF Server-side Services";
-        private const string WCF_MESSAGE = $"Support for {CATEGORY} is limited to .NET Full Framework. To learn more please read: {UPGRADE_LINK}";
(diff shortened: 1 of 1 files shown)


brandonh-msft:
I'm curious about the move of the links - it looks like the link we have today directs folks on how to change from WCF -> gRPC, but the one you're proposing is instructional for WCF -> CoreWCF. Is that correct?
The message we give them, then, says that _both_ are actually suitable... Is there a way we could give them **both** links so they can more easily figure out which suits their application? Or is there a reason why we're seemingly switching to prefer CoreWCF over gRPC when we say both are suitable replacements?

sunandabalu:
> I'm curious about the move of the links - it looks like the link we have today directs folks on how to change from WCF -> gRPC, but the one you're proposing is instructional for WCF -> CoreWCF. Is that correct? The message we give them, then, says that _both_ are actually suitable... Is there a way we could give them **both** links so they can more easily figure out which suits their application? Or is there a reason why we're seemingly switching to prefer CoreWCF over gRPC when we say both are suitable replacements?

Agreed, since we are mentioning both CoreWCF and gRPC as viable options, providing both the links make sense.

samsp-msft:
Updated to show both links.

brandonh-msft (approved):

src/components/Microsoft.DotNet.UpgradeAssistant/Checks/WcfServerCheck.cs
```diff
@@ -16,9 +16,10 @@ namespace Microsoft.DotNet.UpgradeAssistant.Checks
     /// </summary>
     public class WcfServerCheck : IUpgradeReadyCheck
     {
-        private const string UPGRADE_LINK = "https://aka.ms/migrate-wcf-to-grpc";
+        private const string WCF_LINK = "https://aka.ms/CoreWCF/migration";
+        private const string gRPC_LINK = "https://aka.ms/migrate-wcf-to-grpc";
```
brandonh-msft, line 20: Build's failing cuz of this
```suggestion
        private const string GRPC_LINK = "https://aka.ms/migrate-wcf-to-grpc";
```

sunandabalu (approved):
LGTM, you might need to change the variable name to be all caps to avoid build breaks.

PR #1118, opened 2022-05-10
Fixing Roslyn RS2002 errors blocking Arcade builds

brandonh-msft:
The RS2002 errors that are causing the arcade PR to fail are due to 

https://github.com/dotnet/roslyn-analyzers/issues/5890?msclkid=b4dd003dd09e11ecbdf974daa92a5f45#issuecomment-1046043775

wherein using the target-typed `new()` declaration causes this to fire on you. "Reverting" to explicit ctor usage fixes the issue and allows the PR branch to build w/o issue:

image


sunandabalu (approved):
Wow, that was a good catch @brandonh-msft

PR #1119, opened 2022-05-11
Adding .NET 6 installation to pipeline tasks to solve for Arcade builds

brandonh-msft:
Arcade builds are failing .NET 6 Integration tests because they are now installing .NET 7 onto boxes instead of .NET 6. So, we need to forcefully (and side-by-side) install .NET 6 so these ITs can still execute.

This change adds a new build task that does this and so should unblock arcade builds.

### DIFF
--- a/azure-pipeline.yml
+++ b/azure-pipeline.yml
@@ -83,7 +83,13 @@ stages:
           displayName: 'Use .NET Core sdk'
           inputs:
             packageType: sdk
-            version: 5.0.400
+            version: 5.0.x
+            installationPath: $(Build.SourcesDirectory)/.dotnet
+        - task: UseDotNet@2
+          displayName: 'Use .NET Core sdk'
+          inputs:
+            packageType: sdk
+            version: 6.0.100
             installationPath: $(Build.SourcesDirectory)/.dotnet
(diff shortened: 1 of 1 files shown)


sunandabalu:
This is fine for now to unblock arcade updates. @brandonh-msft could you file an issue in the backlog to get rid of hard-coded sdk installations in the build.yaml?

azure-pipeline.yml
```diff
@@ -83,7 +83,13 @@ stages:
...
           inputs:
             packageType: sdk
-            version: 5.0.400
+            version: 5.0.x
+            installationPath: $(Build.SourcesDirectory)/.dotnet
+        - task: UseDotNet@2
+          displayName: 'Use .NET Core sdk'
+          inputs:
+            packageType: sdk
+            version: 6.0.100
```
sunandabalu, line 92: maybe we can do a `6.0.x` here so we are not pinning on a very specific version.

sunandabalu: nvm, this might need to be 6.0.100 for now.

brandonh-msft: Yup, tried `.x` and we hit our Integration test problem. Probably going to have to stick here until we figure out what's going on there.

sunandabalu (approved):

PR #1135, opened 2022-05-23
Fix #1127

### LINKED ISSUE #1127: Unable to load projects when TFM platform version is upgraded
The upgrade fails with this error message
```
Could not load project. Please ensure the selected VS instance has the correct workloads installed for your projects. If Upgrade Assistant selected the incorrect VS version, please pass the argument --vs-path with the appropriate path.
```

ujjwalchadha:
This PR fixes #1127 where the project fails to load the project if the platform version is updated.
It adds a fallback option when the project loading fails and flushes out the old obj data before trying again.

### DIFF
--- a/src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.cs
+++ b/src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.cs
@@ -90,9 +90,24 @@ namespace Microsoft.DotNet.UpgradeAssistant.MSBuild
                 {
                     return Context.ProjectCollection.LoadProject(FileInfo.FullName);
                 }
-                catch (InvalidProjectFileException ex)
+                catch (InvalidProjectFileException)
                 {
-                    throw new UpgradeException(LocalizedStrings.InvalidProjectError, ex);
+                    // TODO Delete obj file retry - a change in platform version may cause load to fail for UWP/WinUI apps
+                    var fileDirectory = FileInfo.Directory;
+                    var dirToDelete = fileDirectory?.EnumerateDirectories().Where(directory => directory.Name == "obj").FirstOrDefault();
+                    if (dirToDelete != null)
+                    {
+                        dirToDelete.Delete(true);
+                    }
+
+                    try
+                    {
+                        return Context.ProjectCollection.LoadProject(FileInfo.FullName);
+                    }
+                    catch (InvalidProjectFileException e)
+                    {
+                        throw new UpgradeException(LocalizedStrings.InvalidProjectError, e);
+                    }
                 }
             }
         }


src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.cs
```diff
@@ -90,9 +90,24 @@ public MBuild.Project Project
                 {
                     return Context.ProjectCollection.LoadProject(FileInfo.FullName);
                 }
-                catch (InvalidProjectFileException ex)
+                catch (InvalidProjectFileException)
                 {
-                    throw new UpgradeException(LocalizedStrings.InvalidProjectError, ex);
+                    // TODO Delete obj file retry - a change in platform version may cause load to fail for UWP/WinUI apps
+                    var fileDirectory = FileInfo.Directory;
+                    var dirToDelete = fileDirectory?.EnumerateDirectories().Where(directory => directory.Name == "obj").FirstOrDefault();
```
brandonh-msft, line 97: Is this recursive such that it could unintentionally grab two `obj/` directories?
Also
```suggestion
                    var dirToDelete = fileDirectory?.EnumerateDirectories().FirstOrDefault(directory => directory.Name == "obj");
```

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.cs
```diff
@@ -90,9 +90,24 @@ public MBuild.Project Project
...
+                    if (dirToDelete != null)
```
brandonh-msft, line 98:
```suggestion
                    if (dirToDelete is not null)
```

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.cs
```diff
@@ -90,9 +90,24 @@ public MBuild.Project Project
...
+                    {
+                        dirToDelete.Delete(true);
```
brandonh-msft, line 100: what is `true` here? Add named parameter since it's not obvious to a reader.

brandonh-msft (changes requested):

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.cs
```diff
@@ -90,9 +90,24 @@ public MBuild.Project Project
...
+                    }
+
+                    try
+                    {
+                        return Context.ProjectCollection.LoadProject(FileInfo.FullName);
```
brandonh-msft, line 105: if `dirToDelete is null`, is there a point in re-trying this or should we just throw the exception?

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.cs
```diff
@@ -92,12 +92,10 @@ public MBuild.Project Project
...
                 {
-                    // TODO Delete obj file retry - a change in platform version may cause load to fail for UWP/WinUI apps
-                    var fileDirectory = FileInfo.Directory;
-                    var dirToDelete = fileDirectory?.EnumerateDirectories().Where(directory => directory.Name == "obj").FirstOrDefault();
-                    if (dirToDelete != null)
+                    // TODO Delete obj file and retry - a change in platform version may cause load to fail for UWP/WinUI apps
+                    if (!DeleteObjDirectory())
                     {
-                        dirToDelete.Delete(true);
+                        throw;
```
brandonh-msft, line 98: Do we want to wrap this in our `UpgradeException` as you were doing before?

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.cs
```diff
@@ -90,13 +90,38 @@ public MBuild.Project Project
...
+                    {
+                        throw new UpgradeException(LocalizedStrings.InvalidProjectError, e);
+                    }
                 }
             }
         }
 
+        private bool DeleteObjDirectory()
+        {
+            var dirToDelete = FileInfo.Directory?.EnumerateDirectories().FirstOrDefault(directory => directory.Name == "obj");
```
twsouthwick, line 115: This feels fragile - for instance, this project redirects those directories to a something outside of the project directory.

What's the cause of the load failure? Is there a root cause that should be fixed in msbuild to handle it?

ujjwalchadha: When the `PlatformVersion` is upgraded (In the UpgradeTFM Step), and a dotnet restore is performed on the project, the `Context.ProjectCollection.LoadProject(FileInfo.FullName)` fails to load the project (I believe due to the inconsistencies in the files from older and newer version)

I could use the project property `BaseIntermediateOutputPath` to get the `obj` directory, but then that cannot be fetched until the project is loaded. So, it's kind of a chicken and egg situation and I have to resort to this solution.

If you have any suggestions, I would be happy to consider and implement it :)

ujjwalchadha: I identified specific files that cause the problem. 
Check if the new fix works instead 🙂

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.cs
```diff
@@ -90,13 +90,43 @@ public MBuild.Project Project
...
         }
 
+        private bool DeleteNugetGeneratedFiles()
+        {
+            if (FileInfo.Directory is null)
+            {
+                return false;
+            }
+
+            var filesToDelete = Directory.GetFiles(FileInfo.Directory.FullName, "PhotoLab.csproj.nuget.g.*", SearchOption.AllDirectories);
```
twsouthwick, line 120: So, all projects have a "PhotoLab.csproj.nuget.g.*"? Shouldn't that be `*.csproj.nuget.g.*`?

It still suffers from the same issue that you're expecting it to be in the directory when that is often customized in larger projects

twsouthwick: Also, what about `.vbproj` etc?

twsouthwick: Is this an issue with NuGet that we should be raising there?

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.cs
```diff
@@ -90,13 +90,43 @@ public MBuild.Project Project
...
+            foreach (var filePath in filesToDelete)
+            {
+                File.Delete(filePath);
+                return true;
```
brandonh-msft, line 124: This `return` makes it so only the first file in the list gets deleted. Is that accurate?

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/PackageMaps/UwpPackageMap.json
```diff
@@ -15,7 +15,7 @@
     "NetCorePackages": [
       {
         "Name": "Microsoft.WindowsAppSDK",
-        "Version": "1.0.0"
+        "Version": "1.1.0-preview1"
```
brandonh-msft, line 18: Any chance 1.1.0 works? I'd like to avoid depending on preview versions as they rev so frequently and (generally) aren't guaranteed to have stable APIs.

ujjwalchadha: Windows App SDK 1.1.0 is not out yet.

brandonh-msft: Is this change needed to fix the bug in the associated issue?

ujjwalchadha: 1.1.0 is finally out. I have updated the version

brandonh-msft (approved):

Issue #1137, opened 2022-05-24
Unable to install upgrade-assistant


Manookian:
Problem encountered on https://dotnet.microsoft.com/en-us/platform/upgrade-assistant/tutorial/install-upgrade-assistant
Operating System: windows

Provide details about the problem you're experiencing. Include your operating system version, exact error message, code sample, and anything else that is relevant.

C:\Users\rxmanook\AppData\Local\Temp\0njm4kfv.0we\restore.csproj : error NU1301: Failed to retrieve information about '
upgrade-assistant' from remote source 'https://lsm-na.pkgs.visualstudio.com/_packaging/281ce2c6-41be-41e9-9e45-aed68ad0
e1f0/nuget/v3/flat2/upgrade-assistant/index.json'.
The tool package could not be restored.
Tool 'upgrade-assistant' failed to install. This failure may have been caused by:

* You are attempting to install a preview release and did not use the --version option to specify the version.
* A package by this name was found, but it was not a .NET tool.
* The required NuGet feed cannot be accessed, perhaps because of an Internet connection problem.
* You mistyped the name of the tool.

For more reasons, including package naming enforcement, visit https://aka.ms/failure-installing-tool

twsouthwick:
@Manookian  are you trying to install it from a directory with custom NuGet sources? Please ensure you're doing it in a non-project folder

twsouthwick:
You probably want to try this command:

```
dotnet tool install -g upgrade-assistant --ignore-failed-sources
```

Manookian:
We have a “custom” nuget repo, how do I get it to stop looking there?

twsouthwick:
Run with the `--ignore-failed-sources` command.  If you have NuGet.org disabled, you'll have to either enable it or add the package to your internal feed

Manookian:
Thanks for your help but all has been unsuccessful, not sure how to determine if "NuGet.org disabled"? I have no problems adding nuget packages to my projects so I assume that NuGet.org is enabled as most of the packages we use are NOT in our custom repo but publicly available packages.

twsouthwick:
What's the result if you run with `--ignore-failed-sources`?

This error should not be seen with that:

```
C:\Users\rxmanook\AppData\Local\Temp\0njm4kfv.0we\restore.csproj : error NU1301: Failed to retrieve information about '
[...]

PR #1138, opened 2022-05-24
Add NuGet.org source to install command

twsouthwick:
Some people may have NuGet.org removed as a source from their machine for one reason or another. This updates the install command to be a single one that ensures failed sources are ignored and NuGet.org is used.

See #1137 for an example


sunandabalu (approved):
Should make the issues coming on failed to install UA :)

Issue #1163, opened 2022-06-14
Please do needful -IMPORTANT


AKULASUDHIKSHANA:
![16551991493407363254682874526909](https://user-images.githubusercontent.com/46380106/173545464-568c52b2-0fd9-40b5-b844-f2008bc53f37.jpg)
Problem encountered on https://dotnet.microsoft.com/en-us/platform/upgrade-assistant/tutorial/install-upgrade-assistant
Operating System: windows

Provide details about the problem you're experiencing. Include your operating system version, exact error message, code sample, and anything else that is relevant.

ujjwalchadha:
@AKULASUDHIKSHANA  Can you try `dotnet tool install -g --ignore-failed-sources upgrade-assistant`
If you do not have api.nuget.org as a default feed, you can run `dotnet nuget add source https://api.nuget.org/v3/index.json -n nuget.org` before running the above command.

@brandonh-msft It seems like there is a problem where the switch `--add-source 'url'` tries to add `CurrentFileDirectory + url` as the feed source instead of just `url`. @bgursahani hit the same issue

brandonh-msft:
it seems the cmd line for `dotnet` takes the `'` character as part of the value for the source. @mjrousos could you update the doc with the right fix/value for this? Not sure if char needs to be escaped, if the char that gets c/p'd from the doc is corrupted somehow when pasted in CLI, etc.

brandonh-msft:
I took care of it - @mjrousos or @twsouthwick if you could review & approve as Sunanda's out, that'd be great thanks

PR #1175, opened 2022-06-20
Fixing install command to remove single-quote; #1163

brandonh-msft:


twsouthwick (approved):
Weird - it worked for me like this. I wonder if the parsing has changed. THere are no spaces, so it's fine anyway

brandonh-msft:
> Weird - it worked for me like this. I wonder if the parsing has changed. THere are no spaces, so it's fine anyway

I figured as much but wondered if perhaps you already had nuget.org in your machine-wide sources so it never even tried to parse the arg from the cmd line?

PR #1198, opened 2022-07-01
Fixing integration tests on latest 1ES build pipelines

brandonh-msft:
Trying a few different things to fix our 1ES builds. Namely, don't use the VS Preview when building.
Ideally we _would_ catch if this causes problems, but not in a CI that would potentially gate our releases.

### DIFF
--- a/azure-pipeline.yml
+++ b/azure-pipeline.yml
@@ -46,10 +46,10 @@ stages:
         pool:
           ${{ if or(eq(variables['System.TeamProject'], 'public'), in(variables['Build.Reason'], 'PullRequest')) }}:
             name: NetCore1ESPool-Public
-            demands: ImageOverride -equals build.windows.amd64.vs2022.pre.open
+            demands: ImageOverride -equals build.windows.amd64.vs2022.open
           ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
             name: NetCore1ESPool-Internal
-            demands: ImageOverride -equals build.windows.amd64.vs2022.pre
+            demands: ImageOverride -equals build.windows.amd64.vs2022
         variables:
         # Enable signing for internal, non-PR builds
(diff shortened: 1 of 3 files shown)


azure-pipeline.yml
```diff
@@ -46,10 +46,10 @@ stages:
         pool:
           ${{ if or(eq(variables['System.TeamProject'], 'public'), in(variables['Build.Reason'], 'PullRequest')) }}:
             name: NetCore1ESPool-Public
-            demands: ImageOverride -equals build.windows.amd64.vs2022.pre.open
+            demands: ImageOverride -equals build.windows.amd64.vs2022.open
           ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
             name: NetCore1ESPool-Internal
-            demands: ImageOverride -equals build.windows.amd64.vs2022.pre
+            demands: ImageOverride -equals build.windows.amd64.vs2022
         variables:
```
sunandabalu, line 53: These specific queues were picked to ensure the MAUI workloads were added. @Sweekriti91 worked with the Arcade team to ensure the right queue was used. @Sweekriti91 do you remember if the non-preview queues had the maui workloads?

Sweekriti91: Non-preview queues do not have the maui workloads, at the moment maui workloads are included only in VS2022 Public Preview. I had worked with the Arcade team so they added the components to the Preview images. Is there a problem using VS2022 Previews? Could we maybe do a matrix that uses previews for MAUI and not-preview for other stuff?

Sweekriti91: @brandonh-msft  are the MAUI integration tests causing a problem? Anything I Can help with or provide more information on w.r.t. MAUI?

brandonh-msft: thanks for letting me know the reason for these. The latest preview VS2022 that gets pulled for these is causing issues because it's pulling in the latest MSBuild. We have to, then, pull in the latest Nuget Framework so it's becoming a cat & mouse game. BUT, knowing this is _required_ for MAUI means it's not a viable solution. Appreciate the 👀 !!

PR #1200, opened 2022-07-01
Prerelease NuGet for Integration.Tests

rainersigwald:
Preview versions of Visual Studio come with preview versions of the .NET SDK, which have new MSBuild and NuGet versions. This can cause problems because the version will be newer than the one directly referenced by this test. Change the test project's reference to allow prerelease NuGet packages, so that it pulls the very latest NuGet, which should almost always be a higher version than the one in the SDK used in the tests.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -21,8 +21,9 @@
       - Newtonsoft.Json versions 9.0.1 and 12.0.2 are built by source-build
       - Version 12.0.2 is written to Version.props
       - Arcade needs to use 9.0.1 so we need to override Version.props value here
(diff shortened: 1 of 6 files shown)


brandonh-msft:
@rainersigwald so I don't think this is doing what we intend... seeing this in the logs:
```
        [Debug] Loading NuGet.Frameworks, Version=6.0.2.3, Culture=neutral, PublicKeyToken=31bf3856ad364e35 without pdb from UA_NuGetede67bf3cdaa4d39815b14ab40dbf9c1
```

but we're expecting we should be pulling in _preview_ versions of this pkg now, right?

PR #1201, opened 2022-07-01
Add PCL workload to .vsconfig

rainersigwald:
Without this Visual Studio workload, the PCL upgrade scenario test will fail. Add it to .vsconfig to give a hint as to what might be wrong.


brandonh-msft (approved):
Nice, thanks!

PR #1204, opened 2022-07-07
Prefer to load UA assemblies from the default ALC

mjrousos:
This update makes it so that assemblies with Microsoft.DotNet.UpgradeAssistant prefixes are only loaded in extensions' assembly load contexts if the assemblies aren't able to be loaded in the default context.


brandonh-msft (dismissed):

twsouthwick:
We did this at one point I believe and intentionally changed it. What's this trying to fix?

mjrousos:
@twsouthwick Huh, I don't remember that. Any idea why we changed it? Simona noticed that the UpgradeStepSample wasn't working because it has a project-to-project reference to Microsoft.DotNet.UpgradeAssistant.Abstractions which causes that dll to show up in the sample extension's build output. Because of that, when she tried to use the extension, it was mysteriously failing. Turns out `ExtensionAssemblyLoadContext` as it was prior to this change was preferring the extension-local version of Microsoft.DotNet.UpgradeAssistant.Abstractions to the one in the default ALC resulting in errors while loading the extension's service provider since base types didn't match.

An alternative fix would be to leave this code alone and make sure that extensions never have shared UA binaries next to them (project-to-project references make that a bit more of a pain but most developers will just use NuGet references which can just use private assets or something similar). It seems like an easy mistake to make, though (our samples somehow fell into this problem) and I can't think of any reason not to just always share Microsoft.DotNet.UpgradeAssistant.* libraries if we're able to. If there's some reason we changed this in the past, though, we can make keeping the binaries straight the extension author's responsibility.

twsouthwick:
The guidance for extensions is not to ship the abstractions library - it's removed from the packages when an extension is built. I'd prefer to update the samples to not copy them to the output.

I think the issue was that we had logic as to what to include and what not to. We hit weird issues with versioning, and it seemed better to just assume if an extension shipping binaries, it wanted those binaries. Since the abstractions library is the main contract, we could  potentially exclude that. Extensions shouldn't be using other UA libraries though.

mjrousos:
That makes sense. I've updated this to only exclude the abstractions library, since that's one that extensions will depend on and should always be loaded in the default alc. I've also updated the samples to not copy that dll.

Microsoft.DotNet.UpgradeAssistant.sln
```diff
@@ -1010,7 +1010,7 @@ Global
 		{DA7F5186-9971-4BC0-9BDB-AD5F9058C4C3} = {8B8F1470-B862-4BF8-B41D-F078D43277FA}
 		{8B8F1470-B862-4BF8-B41D-F078D43277FA} = {60D2D4E3-E055-4A9C-BDB1-7328DAA9BD5C}
 		{16141B3F-F785-4CFD-9FD2-67E9D76C4827} = {4E12F429-CEC0-4080-B0FE-39F9B8B9AE4C}
-		{61A3EF4A-68CC-442E-87EC-F05B874A4C98} = {EE464F98-73F3-4858-8B5C-5254EECFA3B3}
```
mjrousos, line 1013: The backup step tests were showing up under the samples node in the solution explorer for some reason, instead of under tests. This fixes that.

twsouthwick:
It's been a while, but the samples could probably use the pattern used by the loose assemblies that gets things referenced right automatically

brandonh-msft:
Tested this locally against one of my sample projects using `--extension` cmd line arg against the `UpgradeStepSample` and _successfully_ got
image

as expected.

brandonh-msft (dismissed):

src/components/Microsoft.DotNet.UpgradeAssistant.Extensions/ExtensionAssemblyLoadContext.cs
```diff
@@ -5,13 +5,15 @@
 using System.IO;
 using System.Reflection;
 using System.Runtime.Loader;
+
```
twsouthwick, line 8:
```suggestion
```

twsouthwick (dismissed):

SimonaLiao (approved):

brandonh-msft:
@mjrousos looks like this change fails our Release build:
```
CSC : error CA1014: Mark assemblies with CLSCompliant [D:\a\_work\1\s\samples\UpgradeStepSample\UpgradeStepSample.csproj]
##[error]CSC(0,0): error CA1014: (NETCORE_ENGINEERING_TELEMETRY=Build) Mark assemblies with CLSCompliant
  Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CodeFixes -> D:\a\_work\1\s\artifacts\bin\Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CodeFixes\Debug\netstandard2.0\Microsoft.DotNet.UpgradeAssistant.Extensions.Default.CodeFixes.dll
  Downloading from "[https://upgradeassistant.blob.core.windows.net/indexes/2021_09_17.index"](https://upgradeassistant.blob.core.windows.net/indexes/2021_09_17.index%22) to "D:\a\_work\1\s\src\extensions\looseassembly\LooseAssembly.NuGet\.data\2021_09_17.index" (245,503,658 bytes).
  Microsoft.DotNet.UpgradeAssistant.Steps.Backup -> D:\a\_work\1\s\artifacts\bin\Microsoft.DotNet.UpgradeAssistant.Steps.Backup\Debug\netstandard2.0\Microsoft.DotNet.UpgradeAssistant.Steps.Backup.dll
  Microsoft.DotNet.UpgradeAssistant.Steps.Razor -> D:\a\_work\1\s\artifacts\bin\Microsoft.DotNet.UpgradeAssistant.Steps.Razor\Debug\netstandard2.0\Microsoft.DotNet.UpgradeAssistant.Steps.Razor.dll
  Microsoft.DotNet.UpgradeAssistant.Steps.Templates -> D:\a\_work\1\s\artifacts\bin\Microsoft.DotNet.UpgradeAssistant.Steps.Templates\Debug\netstandard2.0\Microsoft.DotNet.UpgradeAssistant.Steps.Templates.dll
  Microsoft.DotNet.UpgradeAssistant.Abstractions.Internal -> D:\a\_work\1\s\artifacts\bin\Microsoft.DotNet.UpgradeAssistant.Abstractions.Internal\Debug\netstandard2.0\Microsoft.DotNet.UpgradeAssistant.Abstractions.Internal.dll
CSC : error CA1014: Mark assemblies with CLSCompliant [D:\a\_work\1\s\samples\SourceUpdaterSample\SourceUpdaterSample.csproj]
##[error]CSC(0,0): error CA1014: (NETCORE_ENGINEERING_TELEMETRY=Build) Mark assemblies with CLSCompliant
  Microsoft.DotNet.UpgradeAssistant.Analysis -> D:\a\_work\1\s\artifacts\bin\Microsoft.DotNet.UpgradeAssistant.Analysis\Debug\netstandard2.0\Microsoft.DotNet.UpgradeAssistant.Analysis.dll
  LooseAssembly.NuGet -> D:\a\_work\1\s\artifacts\bin\LooseAssembly.NuGet\Debug\net5.0\Microsoft.DotNet.UpgradeAssistant.Extensions.LooseAssembly.NuGet.dll
CSC : error CA1014: Mark assemblies with CLSCompliant [D:\a\_work\1\s\samples\PackageMapSample\PackageMapSample.csproj]
##[error]CSC(0,0): error CA1014: (NETCORE_ENGINEERING_TELEMETRY=Build) Mark assemblies with CLSCompliant
CSC : error CA1014: Mark assemblies with CLSCompliant [D:\a\_work\1\s\samples\FindReplaceStepSample\FindReplaceStepSample.csproj]
##[error]CSC(0,0): error CA1014: (NETCORE_ENGINEERING_TELEMETRY=Build) Mark assemblies with CLSCompliant
```

mjrousos:
Ah, there's a root Directory.build.targets that's adding a common AssemblyInfo.cs that isn't getting included anymore after my change. I'll add an import to the new Directory.build.targets to import the root one.

brandonh-msft (approved):

PR #1210, opened 2022-07-13
Adding Binary Analysis feature

### LINKED ISSUE #1089: Provide support for API analysis on binaries

brandonh-msft:
This PR adds into Upgrade Assistant a new feature to analyze **binaries** for API compatibility using the .NET API Catalog (aka apisof.net.
Usage is documented in [the MD file included in the PR](https://github.com/dotnet/upgrade-assistant/blob/6f249b66390413e3496fbd619f2cff488259b51e/docs/binary_analysis.md).

#### Example Usage

**Single File**

To analyze a single binary file, simply pass the path to the command

`analyzebinaries c:\bin\myfile.dll`

**Multiple Files**

Multiple files can be passed to the command as separate arguments:

`analyzebinaries c:\bin\file1.dll c:\bin\file2.dll "c:\Program Files\bin\myfile.exe"`

**Directories**

To analyze all files in a given directory, pass the directory as the argument to the command:

`analyzebinaries c:\bin`

**Mixed mode**

Additionally, Binary Analysis supports the ability to specify both directories and files as its input:

`analyzebinaries c:\bin "c:\Program Files\bin\myfile.exe"`

## Interpreting the results

Upon completion, Binary Analysis will highlight zero or more warnings in your binaries based on the target you've chosen. These warnings include:

Rule ID | Name | Notes | Action
-|-|-|-
`UA9000` | InvalidAssembly | The assembly that was analyzed was not a valid .NET Assembly. | None
`UA9010` | RuleName | The cited API is Unavailable in target framework. | You'll need to update your code to use a different API to make it compatible with the target framework.
`UA9011` | ApiAvailableViaExternalPackage | API is available via an external Package (NuGet) | You can bring the binary forward but you'll need to include an external package from NuGet for it to function on the target framework.
`UA9020` | ApiObsoleted | The API is available in the target framework but has been marked as `[Obsolete]` | The obsolete warning in the target framework may/may not break your build and you'll want to follow the instructions presented by it to modify your code accordingly.
`UA9030` | PlatformNotSupported | API is unsupported in target platform. | The platform you're looking to go to (e.g., Windows, Linux) doens't support the API so you'll need to find its equivalent, or remove/`#IFDEF` its usage.

closes #1089


sunandabalu (approved):
Since we already did a full offline review on the content.

brandonh-msft:
Currently blocked by https://github.com/dotnet/arcade/issues/10016

twsouthwick (approved):
LGTM.

Is there an issue to track incorporating this into normal UA analysis? This is two fold:

1. The binary analysis tends to be a challenge because people throw a directory at it with all of their own *and* 3rd party assemblies and will often lead to false positives (i.e. the .NET Framework build of Newtonsoft is not compatible on core, but there is a core version). The UA integration could be smarter about just feeding it the data from a user's own assemblies.
2. A single report would be a lot easier to deal with

twsouthwick:
Here's an issue: https://github.com/dotnet/upgrade-assistant/issues/1217

PR #1212, opened 2022-07-15
Use build.cmd in common to work around publish issue

hoyosjs:


epananth:
We DON'T need that flag ! hurraaayyyy :) thank you

brandonh-msft (approved):

Issue #1208, opened 2022-07-12
Converts to Maui for non-Maui projects


Sweekriti91:
### Discussed in https://github.com/dotnet/upgrade-assistant/discussions/1150

<div type='discussions-op-text'>

<sup>Originally posted by **ERobishaw** June  2, 2022</sup>
If you find the upgrade assistant is trying to make all your projects into Maui projects, even if they have no xamarin code -- it's because you have multi-targeted frameworks...
In your csproj files, change:
`<TargetFrameworks>net48</TargetFrameworks>`  
      to 
`<TargetFramework>net48</TargetFramework>`

And it will no longer do this...
I didn't dive into the code enough to determine the thinking behind this...  but the analysis believes you have a Maui project if it sees any code related to Xamarin.Forms, OR the **PLURAL** version of "TargetFrameworks" targeting. </div>

twsouthwick:
This is because of the following line:

https://github.com/dotnet/upgrade-assistant/blob/6233e2bba56cc29956a93764b33b4475fcd55914/src/extensions/maui/Microsoft.DotNet.UpgradeAssistant.Extensions.Maui/MauiComponentIdentifier.cs#L16

We should not be using `TargetFrameworks` as the flag to convert to Maui - many projects will have this flag. Currently, this stops any conversion.

PR #1216, opened 2022-07-18
Disable Maui component identifier as this breaks any multitargeting library

twsouthwick:


twsouthwick:
This is a stop-gap to enable multi-targeting libraries to succeed related to https://github.com/dotnet/upgrade-assistant/issues/1208

brandonh-msft (approved):

PR #1220, opened 2022-07-19
Fix broken non interactive mode

twsouthwick:
With recent changing to the hosting model, the custom services are registered first and then the default after. This causes things like the upgrade command to register a non-interactive IUSerInput, but it will be overwritten by the default services. Since this behavior may be seen with other services, this change fixes the issue by:

1. Change the registration of IUserInput to be just an attempt in the default services since we don't want to overwrite an existing one
2. Change the ordering of service registration so the default ones are first and then the commands' customizations.


twsouthwick:
Well, either fixes it. The first is more of a correctness as we don't want to register multiple, so if one exists, just use it. The second is to revert back to previous behavior.

brandonh-msft (approved):

PR #1222, opened 2022-07-19
Changing back to prod build pools

brandonh-msft:
Since we disabled the MAUI Integration Tests entirely, we don't need
to be on the staging environment build pools


twsouthwick (approved):
Is this why things are hanging for other builds?

brandonh-msft:
> Is this why things are hanging for other builds?

yup

PR #1243, opened 2022-08-01
Support winrt consumption using CsWinRT reference

ujjwalchadha:
This PR adds support for CsWinRT consumption in UWP converted projects containing any reference to "vcxproj" projects. We add a CsWinRT reference and CsWinRTIncludes property for each vcxproj reference.


ujjwalchadha:
Waiting for [try-convert PR](https://github.com/dotnet/try-convert/pull/447) to get merged.

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/UWPtoWinAppSDKUpgrade/WinUIPropertiesUpdater.cs
```diff
@@ -85,6 +104,20 @@ public async Task<IUpdaterResult> ApplyAsync(IUpgradeContext context, ImmutableA
                 new List<string>());
         }
 
+        private string ParseProjectNameWithExtension(string projectReference, string extension)
+        {
+            var index = projectReference.IndexOf(".vcxproj");
+            index--;
+            StringBuilder sb = new StringBuilder();
```
brandonh-msft, line 111: I was initially confused by what you're doing here, but I _think_ we can avoid the looping, reversing, etc. by simply doing:
```suggestion
            return new string(projectReference.Substring(0, index).Where(c=>char.IsLetterOrDigit(c)));
```

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/UWPtoWinAppSDKUpgrade/WinUIPropertiesUpdater.cs
```diff
@@ -69,6 +75,19 @@ public async Task<IUpdaterResult> ApplyAsync(IUpgradeContext context, ImmutableA
                     }
                 }
 
+                foreach (var projRef in project.AllProjectReferences)
+                {
+                    if (projRef.Contains(".vcxproj"))
+                    {
+                        var projectName = ParseProjectNameWithExtension(projRef, ".vcxproj");
+                        var csWinRTIncludesValue = projectFile.GetPropertyValue("CsWinRTIncludes") ?? string.Empty;
```
brandonh-msft, line 83: Can we make `CsWinRTIncludes` a `const` in here?

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/UWPtoWinAppSDKUpgrade/WinUIPropertiesUpdater.cs
```diff
@@ -18,6 +20,10 @@ public class WinUIPropertiesUpdater : IUpdater<IProject>
     {
         public const string RuleID = "UA302";
 
+        private const string CsWinRTLogMessage = "A CsWinRTIncludes property with value {0} has been added.\n" +
```
brandonh-msft, line 23: nit: if a const is meant to be used with `string.Format` I like to have it end in `Format` so dev knows this.
```suggestion
        private const string CsWinRTLogMessageFormat = "A CsWinRTIncludes property with value {0} has been added.\n" +
```

brandonh-msft (approved):
Overall LGTM just added a few suggestions

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/UWPtoWinAppSDKUpgrade/WinUIPropertiesUpdater.cs
ujjwalchadha: This doesn't solve the purpose. projectReference looks something like this: "`(ProjectId, #82408bc4-5748-4ae3-b1bb-9a6c0bbc3ec2 - ..\\RuntimeComponent1\\RuntimeComponent1.vcxproj)"`
I want to parse the project name which is the part of the path just before `".vcxproj"` which I do character by character until I reach the first non alphanumeric character. Since this is parsed in reverse, I have to reverse the string again.

brandonh-msft: ah what about utilizing `.LastIndexOf(System.IO.Path.PathSeparator)` in this case?

brandonh-msft:
Just merged t-c into `main` via PR #1248 
Fix up the conflicts in this one (the change to subrepo SHA) and make the changes you needed in t-c as part of this PR then let's 🚀

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IProject.cs
```diff
@@ -21,6 +21,8 @@ public interface IProject
 
         IEnumerable<IProject> ProjectReferences { get; }
 
+        public IEnumerable<string> AllProjectReferences { get; }
```
brandonh-msft, line 24: Is there a way we can accommodate the goal of your PR w/o breaking this public interface?
I don't mind if this gets added to `MSBuildProject` but must be _very_ careful about this change as we ship `.Abstractions` as one of our only public nuget pkgs for extending UA. So I'd love to see if there's a way you can do your work w/o modifying this interface

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.cs
```diff
@@ -42,6 +42,8 @@ internal partial class MSBuildProject : IProject
 
         public string Id => Context.SolutionInfo.GetProjectId(FileInfo.FullName);
 
+        public IEnumerable<string> AllProjectReferences => GetRoslynProject().AllProjectReferences.Select(projRef => projRef.ProjectId.ToString());
```
brandonh-msft, line 45: Seems like there's an optimization we could do here where we're not calling `GetRoslynProject()` every time somebody asks for either this or `.ProjectReferences`. Additionally, the LINQ query in both cases ends up getting executed _every_ time somebody Gets either property.

If either/both of these never change, we should optimize both of them with a private backing field or readonly property set at ctor time.

brandonh-msft (changes requested):

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.cs
ujjwalchadha: I think this is an optimization we can do for many of the properties and methods here but being carefully making sure that introducing state would not cause any breaks. For example, this is a possibility in cases where the underlying project changes using the `ProjectFile.SaveAsync` method and thus `GetRoslynProject()` returns a different project.

I would be happy to take this as a separate issue, as it might require some calculated optimizations.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IProject.cs
ujjwalchadha: Since `MSBuildProject` is not available in the windows extension. `IProject` and `IProjectFile` are 2 places where this can reside. Ideally, IProjectFile should have exposed all project properties and items through some iterative property but it does not. So. at the moment there is no good way to access some items like vcxproj project references through any of these interfaces.
I am ok to discuss this and implement this in a better way if you have a suggestion.

ujjwalchadha: One way to do this could be making this change through the try-convert module

brandonh-msft: Since you need this for Windows project only... could we do `WindowsProject : IProject` and add the property there?

brandonh-msft: Have a look [at this commit](https://github.com/dotnet/upgrade-assistant/commit/555cea49ebd00be7e18fddb9b06986b7d23c8f6b)

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/UWPtoWinAppSDKUpgrade/WinUIPropertiesUpdater.cs
```diff
@@ -18,6 +21,12 @@ public class WinUIPropertiesUpdater : IUpdater<IProject>
     {
         public const string RuleID = "UA302";
 
+        private const string CsWinRTLogMessageFormat = "A CsWinRTIncludes property with value {0} has been added.\n" +
```
brandonh-msft, line 24: just a note you can use literal strings (`@" "`) and have line returns right in the string itself instead of concatenation.

brandonh-msft (changes requested):
LGTM! Just blocking until another UWP SME gives it the 👍🏻👍🏻

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/UWPtoWinAppSDKUpgrade/WinUIPropertiesUpdater.cs
```diff
@@ -69,6 +78,19 @@ public async Task<IUpdaterResult> ApplyAsync(IUpgradeContext context, ImmutableA
...
                 }
 
+                foreach (var projRef in project.AllProjectReferences())
+                {
+                    if (projRef.Contains(".vcxproj"))
+                    {
+                        var projectName = ParseProjectNameWithExtension(projRef, ".vcxproj");
+                        var csWinRTIncludesValue = projectFile.GetPropertyValue(CsWinRTIncludesProperty) ?? string.Empty;
+                        var delimiter = csWinRTIncludesValue.Trim().Length == 0 || csWinRTIncludesValue.EndsWith(";") ? string.Empty : ";";
+                        projectFile.SetPropertyValue("CsWinRTIncludes", $"{csWinRTIncludesValue}{delimiter}{projectName}");
```
manodasanW, line 88:
```suggestion
                        projectFile.SetPropertyValue(CsWinRTIncludesProperty, $"{csWinRTIncludesValue}{delimiter}{projectName}");
```

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/UWPtoWinAppSDKUpgrade/WinUIPropertiesUpdater.cs
```diff
@@ -18,6 +21,12 @@ public class WinUIPropertiesUpdater : IUpdater<IProject>
     {
         public const string RuleID = "UA302";
 
+        private const string CsWinRTLogMessageFormat = "A CsWinRTIncludes property with value {0} has been added.\n" +
+                            "If your project assembly name differs from {0}, update this value with the assembly name.\n" +
```
manodasanW, line 25: Is this message printed for the component scenario where the project assembly name comes into play or is it printed for the vcxproj references scenario?

manodasanW (approved):

src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/UWPtoWinAppSDKUpgrade/WinUIPropertiesUpdater.cs
ujjwalchadha: This is printed for all the vcxproj scenarios (There seems to be no good way to parse the assembly name as we are unable to load cpp projects). So, whenever we add a CsWinRTIncludes property with the name same as the vcxproj file name, we show this message on the console.

brandonh-msft (approved):

PR #1246, opened 2022-08-01
Add check for default/package extension

SimonaLiao:
For server-only projects, there is no need to add references to System.ServiceModel packages. I added a check in the PackageUpdaterStep and now it will only add those packages when the project contains client side (containing ChannelFactory or ClientBase).


SimonaLiao:
Currently the integration tests are not passing because this PR will change the output of steps. I will look into those tests and update the expected results tomorrow.

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageUpdaterStep.cs
```diff
@@ -150,6 +155,46 @@ void AddSubsteps<T>(IEnumerable<Operation<T>> items, string messageFormat, Func<
...
+                        {
+                            containsService = true;
+                        }
+                    }
+
+                    if (containsService)
+                    {
+                        return from p in packages.Additions
+                               where !p.Item.Name.Contains("System.ServiceModel")
+                               select p;
```
mconnew, line 183: Would StartsWith be suitable here? It's faster because it can skip most items immediately if they don't start with the letter S, whereas  Contains needs to search at least until there isn't enough space to contain the searched for string. Also should the overload be used which takes a StringComparison object. I don't know whether package names are case sensitive (I believe they aren't but not certain), in which case you should pass StringComparison.OrdinalIgnoreCase.

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageUpdaterStep.cs
```diff
@@ -150,6 +155,46 @@ void AddSubsteps<T>(IEnumerable<Operation<T>> items, string messageFormat, Func<
...
+
+                return packages.Additions;
+            }
+
+            // Checks if the root has descendant nodes that contains id
+            private static bool ContainsIdentifier(SyntaxNode root, string id)
+            {
+                return (from n in root.DescendantNodes().OfType<IdentifierNameSyntax>()
+                        where n.Identifier.ValueText.Contains(id)
+                        select n).Any();
```
mconnew, line 195: I think you want to use StartsWith for the same performance reasons as mentioned above. As class names are case sensitive, you should pass StringComparison.Ordinal.

brandonh-msft: Additionally. Instead of `where` then `Any` you can use lambda syntax an change this whole statement to read:
```csharp
return root.DescendantNodes().OfType<IdentifierNameSyntax>().Any(n => n.Identifier.ValueText.StartsWith(id, StringComparison.Ordinal);
```

mconnew: That's a good point. You still need to use StringComparison.Ordinal otherwise it needs to do more work when doing comparisons as there are some characters considered equivalent in some languages such as accented characters. Ordinal simply means exact same character ignoring any nuances of the current cultures language.

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageUpdaterStep.cs
```diff
@@ -150,6 +155,44 @@ void AddSubsteps<T>(IEnumerable<Operation<T>> items, string messageFormat, Func<
                 return new UpgradeStepInitializeResult(UpgradeStepStatus.Incomplete, $"{_analyzer.Name} has identified some recommended changes", Risk: _risk);
             }
 
+            // For server-only projects, do not need to add System.ServiceModel packages.
+            private static IEnumerable<Operation<NuGetReference>> UpdatePackageAddition(IDependencyCollection<NuGetReference> packages, IProject project)
+            {
+                var files = project.FindFiles(".cs", ProjectItemType.Compile);
+                var containsService = false;
+                if (files.Any())
```
brandonh-msft, line 163: You can remove this `.Any()` check; the `foreach` loop won't execute if the collection is empty so this is basically making it do the work twice.

mconnew (approved):

brandonh-msft (approved):

PR #1248, opened 2022-08-02
Copy relevant try-convert contents to UA repo

### LINKED ISSUE #1242: Copy try-convert into UA repo instead of subrepoing it
This project has a direct dependency on another tool's code: [try-convert](https://github.com/dotnet/try-convert). Because of its [unsupported nature](https://github.com/dotnet/try-convert#support-policy), if/when we need to make changes to that tool it takes a longer-than-acceptable turnaround time and gates us on external parties who, admittedly, don't support the tool directly.

Since we already pin to a version of TC in our build, it makes sense, then, to simply copy the repo as-is from the currently-pinned commit SHA and continue to make the changes we need to it directly in our own repo.

cc @ujjwalchadha who has had insight into this pain/process

brandonh-msft:
Our PRs into the try-convert project are gated on contributors,
while that project is openly stated as unmaintained.

Since we were previously pinning to an exact commit in t-c anyway, 
we have decided to move forward with copying the portions of t-c
which UA utilizes directly into the try-convert extension area of
the project.
This will allow for faster iteration on fixes/extensions that require
updates to try-convert's functionality and decouple us from a
deprecated project.

Fixes #1242


sunandabalu (approved):

brandonh-msft:
Only diffs are expected ones
image

PR #1258, opened 2022-08-15
Post upgrade report design proposal #1

ujjwalchadha:
In design proposal 1, we create an `UpgradeResultWriter` class which is injected through services and made available to each step and substep and has the following capabilities
- Able to add a stream and corresponding format in which the result is to be written
- Use `AnalyzeResultWriter` to write this result
- Expose a function `WriteResultAsync` to allow steps and substeps to write this result

The `ConsoleUpgrade` class, which runs the steps adds a file stream and the user given format to this UpgradeResultWriter. Any extension is also capable of adding their own format and stream, thus making the output available in different formats and streams (file, console etc)

### DIFF
--- a/src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Commands/Upgrade/ConsoleUpgrade.cs
+++ b/src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Commands/Upgrade/ConsoleUpgrade.cs
@@ -22,6 +22,7 @@ namespace Microsoft.DotNet.UpgradeAssistant.Cli
         private readonly ITelemetry _telemetry;
         private readonly IUpgradeStateManager _stateManager;
         private readonly ILogger<ConsoleUpgrade> _logger;
+        private readonly IUpgradeResultWriter _upgradeResultWriter;
 
         public ConsoleUpgrade(
             IUserInput input,
@@ -32,7 +33,8 @@ namespace Microsoft.DotNet.UpgradeAssistant.Cli
             UpgraderManager upgrader,
             ITelemetry telemetry,
             IUpgradeStateManager stateManager,
-            ILogger<ConsoleUpgrade> logger)
+            ILogger<ConsoleUpgrade> logger,
+            IUpgradeResultWriter upgradeResultWriter)
         {
             _input = input ?? throw new ArgumentNullException(nameof(input));
             _context = context;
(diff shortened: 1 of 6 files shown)


src/extensions/windows/Microsoft.DotNet.UpgradeAssistant.Extensions.Windows/UWPtoWinAppSDKUpgrade/WinUIReferenceAnalyzer.cs
```diff
@@ -71,6 +74,10 @@ public async Task AnalyzeAsync(IProject project, IDependencyAnalysisState state,
                     state.Packages.Remove(package, new OperationDetails() { Risk = BuildBreakRisk.Medium, Details = ImmutableList.Create<string>(package.Name) });
                 }
             }
+
+            var result = new Analysis.AnalyzeResult { FileLocation = "fileLocation", FullDescription = "Description", ResultMessage = "result" };
+            var resultDefinition = new Analysis.AnalyzeResultDefinition { Name = "name", AnalysisResults = ImmutableList.Create(result).ToAsyncEnumerable() };
+            await this._upgradeResultWriter.WriteAsync(ImmutableList.Create(resultDefinition).ToAsyncEnumerable(), token);
```
ujjwalchadha, line 80: this is just an example of how the consumers - steps and substeps can use this. Will be removed from this PR

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Commands/Upgrade/ConsoleUpgrade.cs
```diff
@@ -61,6 +64,9 @@ public async Task RunAsync(CancellationToken token)
                 var steps = await _upgrader.InitializeAsync(context, token);
                 var step = await _upgrader.GetNextStepAsync(context, token);
 
+                using var outStream = File.Create("E:\\outa.txt");
```
ujjwalchadha, line 67: Just a test. Will be replaced with a path in the project root and format coming from ConsoleUpgradeOptions

brandonh-msft (commented):
My gut feeling here is this is far less intuitive from a platform & plumbing standpoint than the other PR.

ujjwalchadha:
Closing this PR and proceeding with #1260 instead.

Issue #460, opened 2021-04-26
Post-Upgrade Report


sunandabalu:

ujjwalchadha:
## Design Proposal

### Requirement

The need is to build a post upgrading reporting system that:

Allows extensions to mention each change made by the upgrade assistant including:

- The modified code
    * Associated file and location of the change
    * Associated documentation link for that change (if possible)
    * Action user needs to take to further fix this (if needed)
    * Risk of the change breaking user code
    * Any associated Rule ID
- is capable of generating the report in multiple formats (csv, html, plaintext etc)
- lists any post upgrade actions that user needs to take for a particular extension

### Proposal

The proposal is to add a service to the existing service collection 


```cs

public record ReportData(
    string RuleID,
    string modifiedCode,
    Location location, // file path of modificaion
    string projectFilePath, // path to csproj
    string nextAction, // further action to fix this
    BuildBreakRisk risk,
);

interface IUpgradeReportWriter
{
    // Append the details to a file
    public abstract void WriteReportData(ReportData reportData);
}
```

This can have multiple implementations like:

```cs
class CsvUpgradeReportWriter : IUpgradeReportWriter {...}

class TextUpgradeReportWriter : IUpgradeReportWriter {...}

class HtmlUpgradeReportWriter : IUpgradeReportWriter {...}

```

And we can inject them to existing collection of services (Either when UA is initialized or as a step):

```cs
services.AddSingleton<IPostUpgradeReportWriter, CsvPostUpgradeReportWriter>();
```

So each extension can use it in their upgrade steps like so:

```cs

class Updater: IUpgradeStep 
{
    private IPostUpgradeReportWriter _reportWriter;

    public Updater(IPostUpgradeReportWriter writer)
    {
        this._reportWriter = writer;
    } 

    public void ApplyAsync()
    {
        ... // perform update
        this._reportWriter.WriteReportData(...); // write details of the change
    }
}
```

PR #1260, opened 2022-08-15
Post upgrade report 

ujjwalchadha:
This PR adds a way to create a post upgrade design report by allowing each upgrade step or stub step to add the result to a collection injected as a service. This collection is then used by the ConsoleUpgrade class to write results to file of a particular format using IAnalyzeWriter

Addresses #460


src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IUpgradeResultCollector.cs
```diff
@@ -0,0 +1,21 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.DotNet.UpgradeAssistant.Analysis;
+
+namespace Microsoft.DotNet.UpgradeAssistant
+{
+    public interface IUpgradeResultCollector
```
twsouthwick, line 11: Is this needed? Couldn't we add an `ICollection<AnalyzeResultDefinition> Results { get; }` to the upgrade context?

ujjwalchadha: We can do that, but I did not want to expose all the methods like remove, clear etc to the external extensions which can unknowingly or knowingly hamper the results that will be printed.

The extensions should just be able to add their results and make them accessible to ConsoleUpgrade for writing. Ideally, I want to remove the clear method from this interface and create a way such that only ConsoleUpgrade can reset the collection once it writes the results after each step results are written. This control should not be given to the extensions.

What do you think?

twsouthwick: That makes sense. Mitigation for this could be that each invocation would put a new list backing in, but an uber-list would be maintained separately.

I think my push back here may also stem from it being a fairly general purpose collector, but has been constratined to be just upgrade results.  What about something like:

```csharp
public interface ICollector<T> : IEnumerable<T>
{
  void Add(T item);
  void AddRange(IEnumerable<T> items);
  void Clear(); // Do we need this if we want it to be add-only?
```

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IUpgradeResultCollector.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        public void AddResult(AnalyzeResultDefinition result);
+
+        public void AddResultCollection(IEnumerable<AnalyzeResultDefinition> resultCollection);
```
brandonh-msft, line 15:
```suggestion
        public void AddResults(IEnumerable<AnalyzeResultDefinition> resultCollection);
```

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Commands/Upgrade/ConsoleUpgrade.cs
```diff
@@ -70,6 +81,11 @@ public async Task RunAsync(CancellationToken token)
                     step = await _upgrader.GetNextStepAsync(context, token);
                 }
 
+                if (_writerProvider.TryGetWriter("html", out var writer))
```
brandonh-msft, line 84: Should try and mimic the code in the Analyze command where we pull out the format specified by the user, try and get a writer for that, etc.
https://github.com/dotnet/upgrade-assistant/blob/e1207a2519a12504527f0cfddd4d57e0b3553abd/src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Commands/Analyze/ConsoleAnalyze.cs#L65

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IUpgradeResultCollector.cs
brandonh-msft: I think letting it `: IEnumerable<T>` imbues it w/ capabilities we don't want people picking up on. This is a bag they should **only ever** populate; we take care of iteration and knowing how to iterate.

Additionally - `T` doesn't make sense because inevitably the objects all go into a writer, and a writer doesn't know how to write a `T`, it only knows how to write an `AnalyzeResultDefinition`. We could `where T : AnalyzeResultDefinition` but I'd argue for the usefulness when, again, the writer's only going to look at the properties that exist on an `AnalyzeResultDefinition`

src/components/Microsoft.DotNet.UpgradeAssistant/UpgradeResultCollector.cs
```diff
@@ -0,0 +1,34 @@
...
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using System.Text;
+using Microsoft.DotNet.UpgradeAssistant.Analysis;
+
+namespace Microsoft.DotNet.UpgradeAssistant
+{
+    internal class UpgradeResultCollector : IUpgradeResultCollector
```
brandonh-msft, line 13:
```suggestion
    internal class UpgradeResultCollector : IUpgradeResultCollector, IEnumerable<AnalyzeResultDefinition>
```

and implement `IEnumerable<T>` via `_resultCollection`

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IUpgradeResultCollector.cs
twsouthwick: > T doesn't make sense because inevitably the objects all go into a writer,


Making it generic is just because it seems like something that would be reusable; i.e. a collection that can be added to, but not removed from. I'd assume the actual usage would be something like `ICollector<AnalyzeResultDefinition> IUpgradeContext.Results {get;}`

brandonh-msft: Ah, gotchya.

ujjwalchadha: @twsouthwick can you explain how the `IUpgradeContext` would work here? I tried making it work but is seems like the `IUpgradeContext` that is passed into the steps by the `ConsoleUpgrade` command [here ](https://github.com/dotnet/upgrade-assistant/blob/e1207a2519a12504527f0cfddd4d57e0b3553abd/src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Commands/Upgrade/ConsoleUpgrade.cs#L50) is a different instance than the `IUpgradeContext` instance injected via dependency injection & obtained through constructor parameters in a step (like an Analyzer), thus they are unable to share state.

I want steps to be able to use the injected context rather than the context that is passed as parameter to the ApplyAsync method, because some steps like Roslyn Analyzers and CodeFixers do not have access to the same parameters, but do have access to the injected services.

twsouthwick: > a different instance than the IUpgradeContext instance injected via dependency injection

Hmmm the context is specifically *not* passed via constructor injection, but rather through parameters to methods. Initially because it required async to create, but also because state should not be something passed around via DI.

> some steps like Roslyn Analyzers and CodeFixers do not have access to the same parameters, but do have access to the injected services.

That's actually not true. They may be using them, but if they are that will be a broken experience if used standalone in VS. Analyzers should not be using them, and in VS code fixers use MEF. The analyzers/codefixers should probably not be managing any sort of report stuff. That should be metadata on top of them that gets added (i.e. there's a description field they have we can add to the report if needed).

brandonh-msft:
Regarding build failures here, they are something that will be caught by executing the `.\build.bat` locally; they are effectively Style rules but when we build Release we build with WarnAsError=true which is what is failing the CI.

twsouthwick:
> Regarding build failures here, they are something that will be caught by executing the .\build.bat locally; they are effectively Style rules but when we build Release we build with WarnAsError=true which is what is failing the CI.

Alternatively, check your warning outputs and make sure that's clean in VS

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Commands/Upgrade/ConsoleUpgrade.cs
```diff
@@ -22,6 +26,9 @@ public class ConsoleUpgrade : IAppCommand
         private readonly ITelemetry _telemetry;
         private readonly IUpgradeStateManager _stateManager;
         private readonly ILogger<ConsoleUpgrade> _logger;
+        private readonly IAnalyzeResultWriterProvider _writerProvider;
+        private readonly IUpgradeResultCollector _upgradeResultCollector;
+        private readonly IOptions<AnalysisOptions> _analysisOptions;
```
twsouthwick, line 31: Looks like `AnalysisOptions` should be separated into what is actually options for analysis, and also `OutputOptions` or something for options around output

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IUpgradeContext.cs
```diff
@@ -41,5 +44,7 @@ public interface IUpgradeContext : IDisposable
         ValueTask ReloadWorkspaceAsync(CancellationToken token);
 
         IUpgradeContextProperties Properties { get; }
+
+        ICollector<AnalyzeResultDefinition> UpgradeResults { get; }
```
twsouthwick, line 48:
```suggestion
        ICollector<AnalyzeResultDefinition> Results { get; }
```

It's already in the `IUpgradeContext` so it seems a bit redundant

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IUpgradeContext.cs
```diff
@@ -6,11 +6,14 @@
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
+using Microsoft.DotNet.UpgradeAssistant.Analysis;
 
 namespace Microsoft.DotNet.UpgradeAssistant
 {
     public interface IUpgradeContext : IDisposable
     {
+        int InstanceNumber { get; }
```
twsouthwick, line 15: What's this for?

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IUpgradeContext.cs
ujjwalchadha: Yes. This PR is not yet ready for review. I just wanted to discuss the basic design. This will be cleaned up in subsequent commits

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IUpgradeContext.cs
ujjwalchadha: Same. Will be cleaned up in future commits

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IUpgradeContext.cs
```diff
@@ -41,5 +42,7 @@ public interface IUpgradeContext : IDisposable
         ValueTask ReloadWorkspaceAsync(CancellationToken token);
 
         IUpgradeContextProperties Properties { get; }
+
+        ICollector<AnalyzeResultDefinition> Results { get; }
```
brandonh-msft, line 46: @twsouthwick / @mjrousos how do we feel about this breaking change to the interface? It worries me a bit, but I think it's important enough to warrant and it forces/enables the collection of Upgrade results by default.
I think the only folks it'd break would be those who have forked the repo and implemented their own Contexts...
Alternatively, we could create `IUpgradeContextWithResults : IUpgradeContext` ?

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/ListCollector.cs
```diff
@@ -0,0 +1,35 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.DotNet.UpgradeAssistant
+{
+    public class ListCollector<T> : ICollector<T>
```
brandonh-msft, line 11: I'm not a fan of bleeding the underlying implementation thru by naming. Could we do
```suggestion
    public class Collector<T> : ICollector<T>
```

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildWorkspaceUpgradeContext.cs
```diff
@@ -224,6 +225,8 @@ private void Workspace_WorkspaceFailed(object? sender, WorkspaceDiagnosticEventA
 
         public IUpgradeContextProperties Properties { get; }
 
+        public ICollector<AnalyzeResultDefinition> Results { get; } = new ListCollector<AnalyzeResultDefinition>();
```
brandonh-msft, line 228: I'm surprised - is this _really_ the _only_ place we implement ``IUpgradeContext`?? Feels unnatural that it's an MsBuild Workspace context but this is the thing we pass in/out of upgraders, etc. but I digress

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Commands/Upgrade/ConsoleUpgrade.cs
brandonh-msft: Agreed, I think when I refactored for Binary Analysis I put the `.Format` property on a base class, the Upgrade options could/should now inherit from that one so it gets the `--format` arg available to it and exposed via UpgradeOptions

brandonh-msft (changes requested):
Looks pretty straightforward, just a few comments/fixes - thanks!!

src/components/Microsoft.DotNet.UpgradeAssistant/UpgradeOptions.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
```
brandonh-msft: I think all of these `using`s can be deleted?

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IUpgradeContext.cs
twsouthwick: I think it's probably fine - just put it as a "Breaking Change". Maybe add a comment to the context that we reserve the right to change it and any custom implementation may break in future releases?

brandonh-msft (dismissed):

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Commands/Upgrade/ConsoleUpgrade.cs
```diff
@@ -43,6 +51,8 @@ public class ConsoleUpgrade : IAppCommand
             _telemetry = telemetry ?? throw new ArgumentNullException(nameof(telemetry));
             _stateManager = stateManager ?? throw new ArgumentNullException(nameof(stateManager));
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            _writerProvider = writerProvider;
```
brandonh-msft, line 54: Either have to `?? throw new ArgumentNullException(nameof(writerProvider));` this or check null & ignore down in `WriteUpgradeReport`

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/UpgradeAssistantHostExtensions.cs
```diff
@@ -75,10 +75,11 @@ public static IHostBuilder UseUpgradeAssistant<TApp>(this IHostBuilder host, IUp
 
                     services.TryAddTransient<IUserInput, ConsoleCollectUserInput>();
 
-                    services.AddAnalysis(options =>
+                    services.ConfigureOutputOptions(options =>
                     {
                         options.Format = upgradeOptions.Format ?? context.Configuration["Analysis:DefaultFormat"];
```
brandonh-msft, line 80: Should we change the config section this is in? Perhaps `Output`? e.g.
```suggestion
                        options.Format = upgradeOptions.Format ?? context.Configuration["Output:DefaultFormat"];
```

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/OutputResultDefinition.cs
```diff
@@ -30,6 +30,6 @@ public record AnalyzeResultDefinition
         /// <summary>
         /// Gets results of the analysis type defined above.
         /// </summary>
-        public IAsyncEnumerable<AnalyzeResult> AnalysisResults { get; init; } = AsyncEnumerable.Empty<AnalyzeResult>();
+        public IAsyncEnumerable<OutputResult> AnalysisResults { get; init; } = AsyncEnumerable.Empty<OutputResult>();
```
brandonh-msft, line 33:
```suggestion
        public IAsyncEnumerable<OutputResult> Results { get; init; } = AsyncEnumerable.Empty<OutputResult>();
```

brandonh-msft (changes requested):
Looking good!

just a few more questions/suggestions

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Commands/Upgrade/ConsoleUpgrade.cs
```diff
@@ -76,9 +86,32 @@ public async Task RunAsync(CancellationToken token)
             {
                 // Do not pass the same token as it may have been canceled and we still need to persist this.
                 await _stateManager.SaveStateAsync(context, default);
+                await WriteUpgradeReport(context, token);
```
twsouthwick, line 89: The statement above this is not passing `token` because it may be canceled. Do you only want this to write if in a non-canceled state? (probably fine, just checking). If so, worth adding a comment why it's different than line 88

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Commands/Upgrade/ConsoleUpgrade.cs
```diff
@@ -76,9 +86,32 @@ public async Task RunAsync(CancellationToken token)
...
             }
         }
 
+        private async Task<bool> WriteUpgradeReport(IUpgradeContext context, CancellationToken token)
+        {
+            var outputDirectory = Path.GetDirectoryName(context.InputPath);
+            if (outputDirectory is null || !context.Results.Any())
```
twsouthwick, line 96: I think I'd find value in an empty report if only to say "No more steps"

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Commands/Upgrade/ConsoleUpgrade.cs
```diff
@@ -76,9 +86,32 @@ public async Task RunAsync(CancellationToken token)
...
+            {
+                return false;
+            }
+
+            var outputFileName = $"UpgradeReport.{_options.Value.Format}";
```
twsouthwick, line 101: How to deal with conflicts? I thought there was logic for that in the analysis stuff....

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Commands/Upgrade/ConsoleUpgrade.cs
```diff
@@ -76,9 +86,32 @@ public async Task RunAsync(CancellationToken token)
...
+            var outputFilePath = Path.Combine(outputDirectory, outputFileName);
+            using var stream = File.Create(Path.Combine(outputDirectory, outputFileName));
+            if (_writerProvider.TryGetWriter(_options.Value.Format, out var writer))
```
twsouthwick, line 104: can't we verify this earlier? it would be a shame to start an upgrade, get through it, but not get a report because the format was not available

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/Collector{T}.cs
```diff
@@ -0,0 +1,35 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.DotNet.UpgradeAssistant
+{
+    public class Collector<T> : ICollector<T>
```
twsouthwick, line 11: does this need to be public? maybe put in the internals abstraction?

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IOutputResultWriterProvider.cs
```diff
@@ -9,21 +9,21 @@
 
 namespace Microsoft.DotNet.UpgradeAssistant
 {
-    public interface IAnalyzeResultWriterProvider
+    public interface IOutputResultWriterProvider
```
twsouthwick, line 12: These are breaking changes - make sure to put that on the changelog

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Commands/Upgrade/ConsoleUpgrade.cs
ujjwalchadha: This will need modifying all integration tests with such file. I'll do that anyway as a part of the next PR where I actually use this to add to results. Will take care of this then.

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Commands/Upgrade/ConsoleUpgrade.cs
ujjwalchadha: Good point, but since this change will also affects the analyze command, I'd rather make a new issue out of it and resolve it separately

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Commands/Upgrade/ConsoleUpgrade.cs
brandonh-msft: If we aren't already, we should validate the user's choice for `--format` is one that's been registered w/ the DI container and error out if one doesn't exist. That would get at Taylor's ask here.
I _think_ we're doing this, though.

src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Commands/Upgrade/ConsoleUpgrade.cs
brandonh-msft: > How to deal with conflicts? I thought there was logic for that in the analysis stuff....

I don't believe there is - it stomps over old ones.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/Collector{T}.cs
brandonh-msft: Good catch, agreed.

brandonh-msft (approved):

Issue #849, opened 2021-09-14
Multiple calls to MSBuild ReloadWorkspace causing perf issues


sunandabalu:
For each of the projects being processed there are multiple times reloading of workspace seems to happen even before any actual steps in upgrade / analyze takes place. For a project with a long dependency list (eg., a project with dependency on ~120 projects) all of the projects are loaded several times for each of the operation on any of the projects, in this case the operation was to GetComponentsAsync on the project. 

`Microsoft.DotNet.UpgradeAssistant.MSBuild.dll!Microsoft.DotNet.UpgradeAssistant.MSBuild.MSBuildWorkspaceUpgradeContext.GetMsBuildWorkspaceAsync(System.Threading.CancellationToken token) Line 183	C#
 	Microsoft.DotNet.UpgradeAssistant.MSBuild.dll!Microsoft.DotNet.UpgradeAssistant.MSBuild.MSBuildWorkspaceUpgradeContext.InitializeWorkspace(System.Threading.CancellationToken token) Line 167	C#
 	Microsoft.DotNet.UpgradeAssistant.MSBuild.dll!Microsoft.DotNet.UpgradeAssistant.MSBuild.MSBuildWorkspaceUpgradeContext.ReloadWorkspaceAsync(System.Threading.CancellationToken token) Line 205	C#
 	Microsoft.DotNet.UpgradeAssistant.MSBuild.dll!Microsoft.DotNet.UpgradeAssistant.MSBuild.DotnetRestorePackageRestorer.RestorePackagesAsync(Microsoft.DotNet.UpgradeAssistant.IUpgradeContext context, Microsoft.DotNet.UpgradeAssistant.IProject project, System.Threading.CancellationToken token) Line 55	C#
 	[Resuming Async Method]	
 	[External Code]	
 	Microsoft.DotNet.UpgradeAssistant.Abstractions.dll!Microsoft.DotNet.UpgradeAssistant.ProcessRunner.RunProcessAsync.AnonymousMethod__1(object _, System.EventArgs __) Line 61	C#
 	[External Code]	
 	[Async Call Stack]	
 	[Async] Microsoft.DotNet.UpgradeAssistant.Extensions.NuGet!Microsoft.DotNet.UpgradeAssistant.Extensions.NuGet.ProjectNuGetReferences.<GetAllPackageReferenceDependenciesAsync>d__15.MoveNext() Line 158	C#
 	[External Code]	
[...]

PR #1262, opened 2022-08-16
Improve analyzer/upgrade performance by using a NuGet caching resource provider

brandonh-msft:
Drafting this PR to ensure I'm not missing something and causing some unintended consequences w/ these changes. Results of performance increase in our UT can be found in [this comment of the linked issue](https://github.com/dotnet/upgrade-assistant/issues/849#issuecomment-1216977392).

---

- [x] You've included unit or integration tests for your change, where applicable.
- [x] You've included inline docs for your change, where applicable.
- [x] There's an open issue for the PR that you are making. If you'd like to propose a new feature or change, please open an issue to discuss the change or find an existing issue.

**For PRs which target a specific extension within UA, you _must_ add the appropriate reviewer group _manually_** \
since GH does not support doing this based on repo paths (yet)

|Extension|Reviewer group to request|
--|--
UWP|`dotnet-upgrade-assistant-uwp`
MAUI|`dotnet-upgrade-assistant-maui`
WCF|`dotnet-upgrade-assistant-wcf`

Extension changes won't be approved by a member of the `dotnet-upgrade-assistant-admin` group until a member of one of those has approved, when required.

- [x] You have requested the appropriate reviewer group for your extension code, or it is platform-level code.

This changes our Transitive Dependency analyzer in the following ways:
* When asking for _transitive_ dependencies of a project, compute once & cache result
* When asking for _direct_ dependencies of a project, compute once & cache result
* When doing a nuget restore, use a static Caching Resource Provider for the calls. **This is the bulk of the perf improvement**
* Style changes from for loop and yield returns to simply giving back a LINQ enumerator

Addresses #849


brandonh-msft:
Perf diffs can be found [in the associated issue](https://github.com/dotnet/upgrade-assistant/issues/849#issuecomment-1216977392).

src/extensions/nuget/Microsoft.DotNet.UpgradeAssistant.Extensions.NuGet/NuGetTransitiveDependencyIdentifier.cs
```diff
@@ -98,28 +116,24 @@ public async Task<TransitiveClosureCollection> GetTransitiveDependenciesAsync(IE
...
-            var requestProvider = new DependencyGraphSpecRequestProvider(new RestoreCommandProvidersCache(), dependencyGraphSpec);
+            var requestProvider = new DependencyGraphSpecRequestProvider(_providerCache, dependencyGraphSpec);
 
             var restoreArgs = new RestoreArgs
             {
                 AllowNoOp = true,
                 CacheContext = _context,
-                CachingSourceProvider = new CachingSourceProvider(new PackageSourceProvider(_settings)),
-                Log = new NuGetLogger(_logger),
+                CachingSourceProvider = _cachingSourceProvider,
```
nkolev92, line 125: The caching source provider is supposed to be shared across different restore instances, so caching it is ok.

src/extensions/nuget/Microsoft.DotNet.UpgradeAssistant.Extensions.NuGet/NuGetTransitiveDependencyIdentifier.cs
```diff
@@ -137,39 +151,19 @@ public TargetGraphLookup(RestoreTargetGraph graph)
...
             {
-                foreach (var item in _graph.Flattened)
-                {
-                    if (string.Equals(item.Data.Match.Library.Name, package.Name, StringComparison.OrdinalIgnoreCase))
-                    {
-                        result = new(item.Data);
-                        return true;
-                    }
-                }
+                var foundItem = _graph.Flattened.FirstOrDefault(item => string.Equals(item.Data.Match.Library.Name, package.Name, StringComparison.OrdinalIgnoreCase));
```
nkolev92, line 162: The non-linq version likely allocate less. If this is a hot path, you may save some GC time.

nkolev92 (commented):
NuGet usage looks good to me.

src/extensions/nuget/Microsoft.DotNet.UpgradeAssistant.Extensions.NuGet/NuGetTransitiveDependencyIdentifier.cs
```diff
@@ -25,10 +28,18 @@ namespace Microsoft.DotNet.UpgradeAssistant.Extensions.NuGet
 {
     public class NuGetTransitiveDependencyIdentifier : ITransitiveDependencyIdentifier
     {
+        private const string UniquePathPart1 = "dotnet-ua";
+        private const string UniquePathPart2 = "dotnet-ua";
+        private const string UniquePathFilename = "dotnet-ua";
+
```
sunandabalu, line 34: Assuming its copy-paste error that all 3 strings have the same value?

brandonh-msft: Actually, no - they were used in 3 different ways but all given the same value. I didn't want to assume they _should_ all be the same value so did this. If you're saying they should be, happy to consolidate :)

brandonh-msft:
CI timings replicate my local box 🎉
image

src/extensions/nuget/Microsoft.DotNet.UpgradeAssistant.Extensions.NuGet/NuGetTransitiveDependencyIdentifier.cs
```diff
@@ -69,22 +87,22 @@ public async Task<TransitiveClosureCollection> GetTransitiveDependenciesAsync(IE
             var tfmInfo = tfms.Select(tfm => new TargetFrameworkInformation { FrameworkName = NuGetFramework.Parse(tfm.ToFullString()) }).ToList();
 
             // Create a project in a unique and temporary directory
-            var path = Path.Combine(Path.GetTempPath(), "dotnet-ua", "restores", Guid.NewGuid().ToString(), "project.txt");
+            var path = Path.Combine(Path.GetTempPath(), UniquePathPart1, UniquePathPart2, Guid.NewGuid().ToString(), UniquePathFilename);
```
sunandabalu, line 91: seems like 3 different values here

brandonh-msft: oh wow good catch! maybe the refactor did something unintended here. Thanks!!

src/extensions/nuget/Microsoft.DotNet.UpgradeAssistant.Extensions.NuGet/NuGetTransitiveDependencyIdentifier.cs
twsouthwick: These are all part of creating a scoped project file for restore. Doesn't really matter what the names are, just that the full path is unique for each invocation. Probably worth putting it in a separate method `CreateUniquePath()` or something so it's clearer what it's doing.

src/extensions/nuget/Microsoft.DotNet.UpgradeAssistant.Extensions.NuGet/NuGetTransitiveDependencyIdentifier.cs
```diff
@@ -137,25 +151,11 @@ public TargetGraphLookup(RestoreTargetGraph graph)
...
-            public IEnumerator<NuGetReference> GetEnumerator()
-            {
-                foreach (var item in _graph.Flattened)
-                {
-                    var library = item.Data.Match.Library;
-
-                    yield return new(library.Name, library.Version.ToString());
-                }
-            }
+            public IEnumerator<NuGetReference> GetEnumerator() => _graph.Flattened.Select(item => new NuGetReference(item.Data.Match.Library.Name, item.Data.Match.Library.Version.ToString())).GetEnumerator();
```
twsouthwick, line 154: nit: this seems more difficult to read than the foreach (and probably allocates a bit more if that matters)

twsouthwick (commented):
I've been seeing some perf implications here, glad it's being addressed! Feel free to blame the original author (i.e. me :)

src/extensions/nuget/Microsoft.DotNet.UpgradeAssistant.Extensions.NuGet/NuGetTransitiveDependencyIdentifier.cs
```diff
@@ -25,10 +28,18 @@ namespace Microsoft.DotNet.UpgradeAssistant.Extensions.NuGet
 {
     public class NuGetTransitiveDependencyIdentifier : ITransitiveDependencyIdentifier
     {
+        private const string UniquePathPart1 = "dotnet-ua";
+        private const string UniquePathPart2 = "restores";
+        private const string UniquePathFilename = "project.txt";
+
+        private static readonly ConcurrentDictionary<string, LibraryDependency> _packageDependencies = new();
```
twsouthwick, line 35: I think it may be easier to test/reason about to have a second implementation of `ITransitiveDependencyIdentifier` that provides the caching behavior given an original `ITransitiveDependencyIdentifier`. Then if we figure out a better way to get the data, the caching is still available to be done

That would make the registration something like this:

```csharp
services.AddSingleton(new CachingTransitiveDependencyIdentifier(new NuGetTransitiveDependencyIdentifier));
```

Also, probably shouldn't be `static` - being a singleton would hold the necessary data

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/Dependencies/TransitiveDependencyExtensions.cs
```diff
@@ -120,30 +121,35 @@ public static async Task<bool> IsTransitiveDependencyAsync(this ITransitiveDepen
...
         private static IEnumerable<NuGetReference> GetProjectTranstiveDependencies(this IProject project)
         {
-            var set = new HashSet<NuGetReference>();
-            var visited = new HashSet<IProject>();
-
-            var projects = new Queue<IProject>();
-            projects.Enqueue(project);
-
-            do
+            return _transitiveDependencyCache.GetOrAdd(project, p =>
```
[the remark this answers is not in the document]
brandonh-msft: Yes this is why I wanted y'all to look it over 😁
I wasn't entirely sure if the `Get` here needs to take into account changed dependencies and/or if the `project` object here would be new/diff such that everything would still work.
But I also recognized the crazy-expensive nature of this query and was hoping to see some improvement by doing this.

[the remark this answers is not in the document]
brandonh-msft: One thing I could do, though, is hold `p.CurrentReferences` alongside the `IEnumerable<NugetReference>` value for the project. If the `.CurrentReferences` set differs when we ask for the transitive dependencies for `p`, recompute.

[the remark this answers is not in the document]
twsouthwick: I doubt this specific part is expensive in comparison to actually determining the transitive. This essentially enumerates all the direct packages, and then you pass that to the caching layer. If all the transitive depenendencies are already cached, this will only be the hit of combining stuff and reducing duplicates.

> One thing I could do, though, is hold p.CurrentReferences alongside the IEnumerable<NugetReference> value for the project. 

Maybe. You'd have to enumerate them, and at that point, it may be more complicated than it needs to be.

This makes me think about the `IFeatureCollection.Revision` property so you can identify if there's a difference in the collection. The `List<>` implementation has a property like that, but there's no exposed way to answer the question "Has this collection changed?".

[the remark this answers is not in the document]
twsouthwick: > project object here would be new/diff such that everything would still work.

Last I checked, we cached the actual project instance and it is a mutable object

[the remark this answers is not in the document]
brandonh-msft: > This makes me think about the `IFeatureCollection.Revision` property so you can identify if there's a difference in the collection. The `List<>` implementation has a property like that, but there's no exposed way to answer the question "Has this collection changed?".

Remember when everything had a `.Changed` event? Oh, the days.... 😞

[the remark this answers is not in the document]
brandonh-msft: New way:
image

Original way:
image

I'm comfortable calling this a wash so will revert this change.

src/extensions/nuget/Microsoft.DotNet.UpgradeAssistant.Extensions.NuGet/NuGetTransitiveDependencyIdentifier.cs
brandonh-msft: It occurs to me that the logic using this (line 89) doesn't take into account the version number of the package being looked up as part of its caching strategy. Fixing this now.

To address the injection vs static concern my thought is this: this cache _should_ be static and close to the behavior because we need not compute the dependencies of a particular nuget pkg & version repeatedly; injecting this behavior doesn't really buy us anything and makes debugging harder.

Regarding a new type of `TransitiveDependencyIdentifier` which takes a `NuGetTransitiveDependencyIdentifier` I'm not following the benefit here either as, per Nikolche, the caching provider was purpose-built for this and should be reused across instances. I'm not sure what the actual object model would look like should we build a new Identifier which _takes_ the Nuget one but somehow utilizes Nuget's caching provider under the covers. Creating one that inherits from and overrides some behavior could make some sense, but again the bottom line was that we should've been creating a single provider to begin with.

twsouthwick: I'm referring to the packageDependencies cache (not the NuGet cache thing).

That does open a question as to what benefit is the `_packageDeendencies` doing for us? Do we now have effictively 2 caches? Also, should the NuGet cache instance be shared across *all* usage?

src/extensions/nuget/Microsoft.DotNet.UpgradeAssistant.Extensions.NuGet/NuGetTransitiveDependencyIdentifier.cs
twsouthwick: @brandonh-msft you may want to look across our code base and see if we could see a perf increase by using a shared instance of this provider

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/Dependencies/TransitiveDependencyExtensions.cs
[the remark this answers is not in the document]
twsouthwick: sounds good. Good to know there's a potential benefit here, but caching is hard to get right (https://www.martinfowler.com/bliki/TwoHardThings.html)

src/extensions/nuget/Microsoft.DotNet.UpgradeAssistant.Extensions.NuGet/NuGetTransitiveDependencyIdentifier.cs
```diff
@@ -69,13 +83,16 @@ public async Task<TransitiveClosureCollection> GetTransitiveDependenciesAsync(IE
...
-            var path = Path.Combine(Path.GetTempPath(), "dotnet-ua", "restores", Guid.NewGuid().ToString(), "project.txt");
-
+            var path = CreateUniquePath();
             var spec = new PackageSpec(tfmInfo)
             {
-                Dependencies = packages.Select(i => new LibraryDependency
+                Dependencies = packages.Select(i =>
                 {
-                    LibraryRange = new LibraryRange(i.Name, new VersionRange(i.GetNuGetVersion()), LibraryDependencyTarget.Package),
+                    var lookupKey = $@"{i.Name}|{i.GetNuGetVersion()?.ToString() ?? "null"}";
```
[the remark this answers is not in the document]
twsouthwick: maybe a record struct?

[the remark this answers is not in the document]
brandonh-msft: I didn't want to have to change the key type and deal w/ that :) so thought doing this would be the easiest/most straightforward way

src/extensions/nuget/Microsoft.DotNet.UpgradeAssistant.Extensions.NuGet/NuGetTransitiveDependencyIdentifier.cs
brandonh-msft: I was merely trying to avoid having to `new()` up the exact same information every time we ask for pkg info. When I did profiling, these kinds of objects were _numerous_ so I am trying to trim them down in an effort to optimize.

brandonh-msft: Also, yes, [based on what @nkolev92 said about the caching provider](https://github.com/dotnet/upgrade-assistant/pull/1262#discussion_r951947404) :)

brandonh-msft:
@twsouthwick any more feedback? am I missing answering anything?

src/extensions/nuget/Microsoft.DotNet.UpgradeAssistant.Extensions.NuGet/NuGetTransitiveDependencyIdentifier.cs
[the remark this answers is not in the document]
twsouthwick: a record struct would be easier - and depending how much you'll be doing it, will be much more efficient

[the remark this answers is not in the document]
twsouthwick: although with c# 10, the interpolated stuff may not be as bad....still allocates, but not as much as it used to

twsouthwick (approved):
LGTM!

PR #1281, opened 2022-09-01
Add using System.Threading.Tasks in Constants.cs

SimonaLiao:
This PR adds System.Threading.Tasks to the Constants.cs for the case when user's original code does not include this using directive and the extension modifies void Main() to async Task Main().


mconnew (approved):
LGTM :shipit:

brandonh-msft (approved):

Issue #1265, opened 2022-08-23
Upgrading from Xamarin.Forms to MAUI does not work


holecekp:
- [X] Include the log file `upgrade-assistant.clef` that is produced in the working directory
 
[upgrade-assistant.zip](https://github.com/dotnet/upgrade-assistant/files/9409440/upgrade-assistant.zip)
[AnalysisReport.zip](https://github.com/dotnet/upgrade-assistant/files/9409441/AnalysisReport.zip)

### Describe the bug
The upgrade-assistant is not able to migrate Xamarin.Forms solution to MAUI. It fails even on a blank Xamarin.Forms solution created by the default Visual Studio template (without any user changes)! It modifies only a single file (.csproj for the Android project) and skips the others completely! For example, there are plenty of .cs files with `using Xamarin.Forms` are left unchanged by the upgrade-assistant. It is interesting that running the assistant with the "analyze" command lists the changes to be made, but running it with the "upgrade" command does not perform of these changes.

### To Reproduce
1. Select "Create a new project" in Visual Studio 2022.
2. Choose "Mobile App (Xamarin.Forms)" in the templates list.
3. I have named the solution as "MyApp", selected that it should use the "Flyout" template, and I have checked only Android in the platforms list. Confirm the dialog.
4. It is recommended to create a git repository so that you can track which files where modified by the upgrade-assistant later.
5. Run the upgrade assistant for the solution: `upgrade-assistant upgrade D:\dotNET\MyApp\MyApp.sln`
6. Go through the all steps. I have selected MyApp.Android as the entry point. Next, it asks for the project that should be converted. There is only one choice (MyApp.Android) because the upgrade assistant shows "[Completed]" next to the MyApp project (.NET Standard). So I have selected MyApp.Android again. I have skipped the backup and confirmed all the other steps. The upgrade assistant states that the upgrade has been completed.
7. Now look at Git which files has been changed. Only MyApp.Android.csproj has been modified!!! No other modification has been made. Namely, there are plenty of `using Xamarin.Forms` in the shared .NET Standard (MyApp) project. It seems that this project has been skipped by the assistant completely.
8. I have tried to run the assistant again just for this shared project: 
[...]

PR #1238, opened 2022-07-29
maui bug fixes

Sweekriti91:
Fix for issue : https://github.com/dotnet/upgrade-assistant/issues/1208

- makes MAUI check more robust using the `UseMaui` property value
- Adds the `[ApplicableComponents(ProjectComponents.Maui)]` tag to MAUI extension analyzers so they don't show up for other project types

PR #1288, opened 2022-09-04
Add .NET MAUI support version doc

Tak1wa:
## Description

Added to the requirement that it is available from the Azure DevOps Pipeline version as well as UWP support for the following reasons:

- .NET MAUI support is not available in stable as of today, it appears to be available only from the development version of Azure DevOps Pipeline.
- Functions related to UWP are specified to use the development version version, but there is no description for Xamarin.Forms

Addresses #1265

### DIFF
--- a/docs/maui_support.md
+++ b/docs/maui_support.md
@@ -4,6 +4,7 @@ We are currently building out support for upgrade-assistant to handle Xamarin.Fo
 
 ## Supported Project types and minimum requirements
 
+- The latest version of .NET Upgrade Assistant from the [Azure Devops Pipeline](https://dev.azure.com/dnceng/public/_artifacts/feed/dotnet-tools)
 - Xamarin.Forms app using Xamarin.Forms 5.0 or higher
   - ✅ iOS project
   - ✅ Android project


brandonh-msft (changes requested):
I'm not sure MAUI capability is predicated on using a nightly of UA; if there is some bug fixed recently that hasn't been pushed to the GA version, it will show up in GA when we do our next release, which will be after CoreWCF finalizes its extension.

Can somebody from @dotnet/dotnet-upgrade-assistant-maui  confirm? Thanks

Sweekriti91:
@brandonh-msft  you are correct, some bug fixes were pushed but not released, it should roll into the next GA release. Outside of that, there are no new PRs that are being merged in at the moment.

Sweekriti91:
the last PR I added to address a few of the MAUI related issues was : https://github.com/dotnet/upgrade-assistant/pull/1238 

This is the one I guess would resolve the problem? Although the problem looks like the current GA doesn't have this last PR code in it? the Nightly build does, so like you mentioned, once we do a next release to GA it should resolve most of this.

brandonh-msft:
Thanks - linked & closed

PR #1291, opened 2022-09-07
Improve analysis experience

### LINKED ISSUE #1034: The analyze command doesn't identify NuGet packages that need updated to newer versions
One of the useful features of the `analyze` commands is the ability to see which NuGet dependencies will need to change as part of upgrade.

This doesn't work for packages that need updated to work with the new TFM, though. As an example, I have a project with a dependency on EntityFramework version 6.1. This version of the EF package won't work with .NET 6 so I would expect the `analyze` command to identify the package and suggest updating to version 6.4, instead (since that version works with .NET 6).

The `upgrade` command correctly makes that replacement but the analyze command doesn't list EntityFramework as a package that needs to change.

I haven't debugged through the tool to root cause this yet but perhaps the analyze command is evaluating whether NuGet packages will work based on the current TFM instead of the future one? Or perhaps the target compatibility reference analyzer isn't used by the analyze command?

### LINKED ISSUE #1102: Getting UA0002 with no clear explanation why and what to fix
### Describe the bug
When running upgrade assistant get following:
> Diagnostic UA0002 with the message This type is not supported on .NET Core/.NET 5+ and should be replaced with a modern equivalent. generated

And it is not clear, what type, why. I've scanned through migration guide and could not find anything related.

### To Reproduce

Create new web project:
```
dotnet new web
```
and add class into Program.cs:
```
public class OverrideSuccessResponseCode : ResultFilterAttribute, IApiResponseMetadataProvider
{
    public int StatusCode { get; }
    public Type Type { get; } = typeof(void);

    public OverrideSuccessResponseCode(int statusCode) => StatusCode = statusCode;

    public override void OnResultExecuted(ResultExecutedContext context)
    {
        if (context.HttpContext.Response.StatusCode == (int)HttpStatusCode.OK)
        {
            context.HttpContext.Response.StatusCode = StatusCode;
        }
    }
    void IApiResponseMetadataProvider.SetContentTypes(MediaTypeCollection contentTypes)
    {
    }
}
```
After running
```
upgrade-assistant.exe analyze .\test-upgrade.csproj
```
You should get:
```
PS C:\temp\test-upgrade> upgrade-assistant.exe analyze .\test-upgrade.csproj
-----------------------------------------------------------------------------------------------------------------
Microsoft .NET Upgrade Assistant v0.3.310801+057ffba0656270dc5d83c015748ac56daed55886

We are interested in your feedback! Please use the following link to open a survey: https://aka.ms/DotNetUASurvey
-----------------------------------------------------------------------------------------------------------------

[08:26:03 INF] Loaded 6 extensions
[08:26:03 INF] Using MSBuild from C:\Program Files\dotnet\sdk\6.0.201\
[08:26:03 INF] Using Visual Studio install from C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional [v16]
[08:26:05 INF] Writing output to C:\temp\test-upgrade\AnalysisReport.sarif
[08:26:05 INF] Recommending executable TFM net6.0 because the project builds to an executable
[08:26:06 INF] Reference to .NET Upgrade Assistant analyzer package >(Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers, version 0.3.310801) needs to be added
[08:26:07 INF] Reference to Newtonsoft package (Microsoft.AspNetCore.Mvc.NewtonsoftJson, version 6.0.4) needs to be added
[08:26:07 INF] Running analyzers on test-upgrade
[08:26:08 INF] Identified 1 diagnostics in project test-upgrade
[08:26:08 INF] Diagnostic UA0002 with the message This type is not supported on .NET Core/.NET 5+ and should be replaced with a modern equivalent. generated
```

upgrade-asstant.clef file contains line:
> {"@t":"2022-04-14T05:26:08.3477296Z","@mt":"Diagnostic {Id} with the message {Message} generated","Id":"UA0002","Message":"This type is not supported on .NET Core/.NET 5+ and should be replaced with a modern equivalent.","SourceContext":"Microsoft.DotNet.UpgradeAssistant.Steps.Source.DiagnosticAnalyzerResultProvider"}

AnalysisReport.sarif file contains:
```
{
      "tool": {
        "driver": {
          "name": "API Upgradability",
          "semanticVersion": "0.3.310801",
          "informationUri": "https://docs.microsoft.com/en-us/dotnet/core/porting/upgrade-assistant-overview",
          "rules": [
            {
              "id": "UA0002",
              "fullDescription": {
                "text": "Types should be upgraded"
              },
              "helpUri": "about:blank"
            }
          ]
        }
      },
      "results": [
        {
          "ruleId": "UA0002",
          "message": {
            "text": "This type is not supported on .NET Core/.NET 5+ and should be replaced with a modern equivalent."
          },
          "locations": [
            {
              "physicalLocation": {
                "artifactLocation": {
                  "uri": "file:///C:/temp/test-upgrade/Program.cs"
                },
                "region": {
                  "startLine": 20
                }
              }
            }
          ]
        }
      ],
      "columnKind": "utf16CodeUnits"
    }
```
From this information it is not clear what and how to fix.
Running assistant with upgrade command does not change source in any way.

mjrousos:
A few fixes related to the analysis workflow:

* Display analyzer message in analysis reports rather than description (fixes #1102)
* Fix code for retrieving packages from non-SDK projects which wasn't working (fixes #1034)
* Fix package loader to more thoroughly and correctly check TFM compatibility

### DIFF
--- a/src/extensions/nuget/Microsoft.DotNet.UpgradeAssistant.Extensions.NuGet/PackageConfig.cs
+++ b/src/extensions/nuget/Microsoft.DotNet.UpgradeAssistant.Extensions.NuGet/PackageConfig.cs
@@ -17,7 +17,9 @@ namespace Microsoft.DotNet.UpgradeAssistant.Extensions.NuGet
                 yield break;
             }
 
-            var packages = doc.Root.Element("packages");
+            // Don't use doc.Root.Element because the
+            // packages element likely *is* the root
+            var packages = doc.Element("packages");
 
             if (packages is null)
             {
--- a/src/extensions/nuget/Microsoft.DotNet.UpgradeAssistant.Extensions.NuGet/PackageLoader.cs
+++ b/src/extensions/nuget/Microsoft.DotNet.UpgradeAssistant.Extensions.NuGet/PackageLoader.cs
@@ -8,6 +8,7 @@ using System.IO;
 using System.IO.Compression;
 using System.Linq;
 using System.Net.Http;
+using System.Runtime.CompilerServices;
 using System.Runtime.Versioning;
 using System.Threading;
 using System.Threading.Tasks;
@@ -63,7 +64,7 @@ namespace Microsoft.DotNet.UpgradeAssistant.Extensions.NuGet
             return targetFrameworks.All(tfm => packageFrameworks.Any(f => DefaultCompatibilityProvider.Instance.IsCompatible(NuGetFramework.Parse(tfm.Name), f)));
         }
 
-        public Task<IEnumerable<NuGetReference>> GetNewerVersionsAsync(NuGetReference reference, IEnumerable<TargetFrameworkMoniker> tfms, PackageSearchOptions options, CancellationToken token)
+        public IAsyncEnumerable<NuGetReference> GetNewerVersionsAsync(NuGetReference reference, IEnumerable<TargetFrameworkMoniker> tfms, PackageSearchOptions options, CancellationToken token)
         {
             if (reference is null)
             {
@@ -88,12 +89,12 @@ namespace Microsoft.DotNet.UpgradeAssistant.Extensions.NuGet
                 throw new ArgumentNullException(nameof(options));
             }
 
-            var result = await SearchByNameAsync(packageName, tfms, options, null, sources, token: token).ConfigureAwait(false);
+            var result = SearchByNameAsync(packageName, tfms, options, null, sources, token: token);
 
-            return result.LastOrDefault();
+            return await result.LastOrDefaultAsync(token).ConfigureAwait(false);
         }
 
-        private async Task<IEnumerable<NuGetReference>> SearchByNameAsync(string name, IEnumerable<TargetFrameworkMoniker> tfms, PackageSearchOptions options, NuGetVersion? currentVersion = null, IEnumerable<PackageSource>? sources = null, CancellationToken token = default)
+        private async IAsyncEnumerable<NuGetReference> SearchByNameAsync(string name, IEnumerable<TargetFrameworkMoniker> tfms, PackageSearchOptions options, NuGetVersion? currentVersion = null, IEnumerable<PackageSource>? sources = null, [EnumeratorCancellation] CancellationToken token = default)
         {
             var results = new List<IPackageSearchMetadata>();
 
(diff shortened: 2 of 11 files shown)


src/extensions/nuget/Microsoft.DotNet.UpgradeAssistant.Extensions.NuGet/PackageConfig.cs
```diff
@@ -17,7 +17,9 @@ public static IEnumerable<NuGetReference> GetPackages(string path)
                 yield break;
             }
 
-            var packages = doc.Root.Element("packages");
+            // Don't use doc.Root.Element because the
+            // packages element likely *is* the root
+            var packages = doc.Element("packages");
```
twsouthwick, line 22: weird - is there a test for this?

src/extensions/nuget/Microsoft.DotNet.UpgradeAssistant.Extensions.NuGet/ProjectNuGetReferences.cs
```diff
@@ -54,7 +54,7 @@ public NugetPackageFormat PackageReferenceFormat
             }
         }
 
-        private string? GetPackagesConfigPath() => _project.FindFiles("packages.config", ProjectItemType.Content).FirstOrDefault();
+        private string? GetPackagesConfigPath() => _project.FindFiles("packages.config", ProjectItemType.None).FirstOrDefault();
```
twsouthwick, line 57: Should we just find the file by name without any other constraints?

src/extensions/try-convert/MSBuild.Conversion.Project/MSBuild.Conversion.Project.csproj
```diff
@@ -12,7 +12,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <InternalsVisibleTo Include="Smoke.Tests" />
+    <To Include="Smoke.Tests" />
```
twsouthwick, line 15: ? Is this intentional? What's the `To` `ItemGroup`?

twsouthwick (approved):
LGTM to me. I have a couple of questions, but nothing blocking

src/extensions/nuget/Microsoft.DotNet.UpgradeAssistant.Extensions.NuGet/PackageLoader.cs
```diff
@@ -122,19 +123,24 @@ private async Task<IEnumerable<NuGetReference>> SearchByNameAsync(string name, I
                 }
             }
 
-            return FilterSearchResults(name, results, tfms, currentVersion, options.LatestMinorAndBuildOnly);
+            // Switch to an IEnumerable here because consumers of this API
```
brandonh-msft, line 126: I think this one's breaking the build.
```suggestion
            // Switch to an IEnumerable here because consumers of this API
```

brandonh-msft (changes requested):

src/extensions/nuget/Microsoft.DotNet.UpgradeAssistant.Extensions.NuGet/ProjectNuGetReferences.cs
mjrousos: Yea, I went back and forth on that. I went this route because in all the examples I looked at, packages.config was 'None' but it probably doesn't hurt to find it regardless of item type. I'll do that instead.

brandonh-msft (approved):

PR #1292, opened 2022-09-08
Add report data to steps

ujjwalchadha:


src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/WellKnownDocumentationUrls.cs
```diff
@@ -0,0 +1,14 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.DotNet.UpgradeAssistant
+{
+    public static class WellKnownDocumentationUrls
+    {
+        public static readonly Uri UpgradeAssistantUsageDocumentationLink = new Uri("https://github.com/dotnet/upgrade-assistant#usage");
```
brandonh-msft, line 12: If this is in `DocumentationUrls` class, probably don't need `DocumentationLink` at the end of the name
```suggestion
        public static readonly Uri UpgradeAssistantUsage = new Uri("https://github.com/dotnet/upgrade-assistant#usage");
```

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Backup/BackupStep.cs
```diff
@@ -89,6 +92,33 @@ protected override Task<UpgradeStepInitializeResult> InitializeImplAsync(IUpgrad
...
+            {
+                FileLocation = backupLocation ?? "Not specified",
+                RuleId = Id,
+                ResultMessage = status switch
+                {
+                    UpgradeStepStatus.Skipped => "Backup Skipped",
+                    UpgradeStepStatus.Failed => "Backup Failed",
+                    UpgradeStepStatus.Complete => "Backup Complete",
+                    UpgradeStepStatus.Incomplete => "Backup Incomplete",
+                    _ => throw new Exception("Invalid UpgradeStepStatus")
```
brandonh-msft, line 107:
```suggestion
                    _ => throw new ArgumentException("Invalid UpgradeStepStatus", nameof(status))
```

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Packages/PackageUpdaterStep.cs
```diff
@@ -262,9 +263,27 @@ protected override Task<UpgradeStepApplyResult> ApplyImplAsync(IUpgradeContext c
...
+            private void AddResultToContext(IUpgradeContext context)
+            {
+                var result = new OutputResult()
+                {
+                    FileLocation = context.CurrentProject?.GetFile()?.FilePath ?? "Project not loaded",
+                    RuleId = Id,
+                    ResultMessage = Title
+                };
+
+                context.Results.Add(new OutputResultDefinition()
```
brandonh-msft, line 279: This chunk seems common - could we add a constructor or `.Create()` method on the type to just take `Name` but prepopulate `InformationUri` and `Results` as we continue to do in this PR?

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Source/CodeFixerStep.cs
```diff
@@ -155,6 +163,26 @@ protected override async Task<UpgradeStepApplyResult> ApplyImplAsync(IUpgradeCon
             return new UpgradeStepApplyResult(UpgradeStepStatus.Complete, $"No instances of {DiagnosticId} need fixed");
         }
 
+        private void AddResultToContext(IUpgradeContext context, string diagnosticId, string location, string status, string resultMessage)
```
brandonh-msft, line 166: is it appropriate for `status` to be `string` or should it follow an enum (perhaps already defined?)

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Source/CodeFixerStep.cs
```diff
@@ -129,7 +135,9 @@ protected override async Task<UpgradeStepApplyResult> ApplyImplAsync(IUpgradeCon
                 var newDiagnosticCount = Diagnostics.Count();
                 if (diagnosticCount == newDiagnosticCount)
                 {
-                    Logger.LogWarning("Diagnostic {DiagnosticId} was not fixed as expected. This may be caused by the project being in a bad state (did NuGet packages restore correctly?) or by errors in analyzers or code fix providers related to this diagnostic.", DiagnosticId);
+                    var description = $"Diagnostic {DiagnosticId} was not fixed as expected. This may be caused by the project being in a bad state (did NuGet packages restore correctly?) or by errors in analyzers or code fix providers related to this diagnostic.";
+                    Logger.LogWarning(description);
+                    AddResultToContext(context, DiagnosticId, string.Empty, "Success", description);
```
brandonh-msft, line 140: based on the description, this doesn't seem like Success...

src/extensions/try-convert/Microsoft.DotNet.UpgradeAssistant.Extensions.TryConvert/TryConvertRunner.cs
```diff
@@ -83,6 +85,26 @@ public async Task<UpgradeStepInitializeResult> InitializeAsync(IProject project,
             }
         }
 
+        private void AddResultToContext(IUpgradeContext context, string status, string resultMessage)
```
brandonh-msft, line 88: at this point I'd love to see `.AddResult()` as an extension method on `IUpgradeContext` which takes the necessary parameters OR a delegate to populate things in a more custom manner. This code seems to have become boilerplate for every Step

src/extensions/try-convert/Microsoft.DotNet.UpgradeAssistant.Extensions.TryConvert/TryConvertRunner.cs
```diff
@@ -95,12 +117,16 @@ private async Task<UpgradeStepApplyResult> RunTryConvertAsync(IUpgradeContext co
 
             if (!result)
             {
-                _logger.LogCritical("Conversion with try-convert failed.");
-                return new UpgradeStepApplyResult(UpgradeStepStatus.Failed, $"Conversion with try-convert failed.");
+                var description = "Conversion with try-convert failed.";
+                _logger.LogCritical(description);
```
brandonh-msft, line 121: why Critical here vs Debug as was done everywhere else?

src/extensions/try-convert/Microsoft.DotNet.UpgradeAssistant.Extensions.TryConvert/TryConvertRunner.cs
```diff
@@ -95,12 +117,16 @@ private async Task<UpgradeStepApplyResult> RunTryConvertAsync(IUpgradeContext co
...
+                AddResultToContext(context, "Failed", description);
+                return new UpgradeStepApplyResult(UpgradeStepStatus.Failed, description);
             }
             else
             {
-                _logger.LogInformation("Project file converted successfully! The project may require additional changes to build successfully against the new .NET target.");
+                var description = "Project file converted successfully! The project may require additional changes to build successfully against the new .NET target.";
```
brandonh-msft, line 127:
```suggestion
                var description = "Project file converted! The project may require additional changes to build successfully against the new .NET target.";
```

tests/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Backup.Tests/BackupStepTests.cs
```diff
@@ -144,6 +145,7 @@ private static IUpgradeContext GetContext(AutoMock mock, string inputPath, strin
             context.Setup(c => c.InputPath).Returns(inputPath);
             context.Setup(c => c.InputIsSolution).Returns(Path.GetExtension(inputPath).Equals(".sln", StringComparison.OrdinalIgnoreCase));
             context.Setup(c => c.Properties).Returns(mock.Container.Resolve<IUpgradeContextProperties>());
+            context.Setup(c => c.Results).Returns(new Mock<ICollector<OutputResultDefinition>>().Object);
```
brandonh-msft, line 149: Could we also add tests that check to make sure Results are populated as we anticipate w/ the other changes in this PR?

brandonh-msft (changes requested):

brandonh-msft:
as an aside: I was _trying_ to get `CODEOWNERS` to automagically add uwp & admin groups for a PR like this but see you had to add the UWP one manually :/ need to do some more work here I guess.

brandonh-msft (changes requested):
re-approval hoping to satisfy the `-admin` group

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Source/CodeFixerStep.cs
ujjwalchadha: Good catch. My bad.

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Source/CodeFixerStep.cs
ujjwalchadha: Sure if we delegate this to an extension method, an enum would make more sense

src/extensions/try-convert/Microsoft.DotNet.UpgradeAssistant.Extensions.TryConvert/TryConvertRunner.cs
ujjwalchadha: This is existing code. I have just removed the description out of it. Do you think it should be changed?

tests/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Backup.Tests/BackupStepTests.cs
ujjwalchadha: We should be able to test this by fixing the broken integration tests (which should now expect the upgrade report)

src/extensions/try-convert/Microsoft.DotNet.UpgradeAssistant.Extensions.TryConvert/TryConvertRunner.cs
brandonh-msft: ah if was already like this, probably fine then.

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Backup/BackupStep.cs
```diff
@@ -89,6 +92,11 @@ protected override Task<UpgradeStepInitializeResult> InitializeImplAsync(IUpgrad
             }
         }
 
+        private void AddResultToContext(IUpgradeContext context, string backupLocation, UpgradeStepStatus status, string description)
+        {
+            context.AddResult("Backup Step", backupLocation, Id, status, description);
```
brandonh-msft, line 97: why not use `.Title` here? Perhaps even add an overload to the new `.AddResult` that takes an `UpgradeStep` obj and does this automatically? Just trying to clean up the code and make usage of this new capability as consistent and affordable as possible

brandonh-msft (approved):

src/extensions/default/Microsoft.DotNet.UpgradeAssistant.Steps.Backup/BackupStep.cs
ujjwalchadha: Not sure if adding UpgradeStep obj would add much value. Most of the information here cannot be directly deduced from the step object but is local to the particular class or is a local variable. We can only obtain Id and maybe title from the step. But then we have cases like try-convert step where the upgrade is not performed in the step class but a separate class - TryConvertRunner - which doesn't have access to the step object, the Id or the Title properties. 

I believe we might end up over engineering. What do you think?

manodasanW (approved):

brandonh-msft (approved):
🚢

thanks Ujjwal!

Issue #1297, opened 2022-09-16
Version 0.4.346201 throws unhandled exception parsing command line arguments


mrahl:
- [x] Include the log file `upgrade-assistant.clef` that is produced in the working directory

`{"@t":"2022-09-16T12:45:04.6502249Z","@mt":"Hosting starting","@l":"Debug","EventId":{"Id":1,"Name":"Starting"},"SourceContext":"Microsoft.Extensions.Hosting.Internal.Host"}`

### Describe the bug
After updating upgrade assistant to the latest version (0.4.346201) and trying to launch it with the `--extension` switch it throws an unhandled exception.

- Going back to the previous version (0.4.336902), it does not have this issue. It works fine with my extension.
- If I omit the --extension argument I also don't run into the issue.
- If I give the extension argument a path to a nonexisting or empty folder, I still get the same exception. So it does not seem to relate to the contents of my extension.
-  Similarly, if I give the path to a csproj file that does not exist, I still get the same exception.
 
This, together with the stack trace of the exception (see below), seems to indicate that it is the argument parsing itself that fails.

### To Reproduce

The contents of the csproj or the extension does not seem to matter, it doesn't even have to exist and the behavior is still the same.

Update the tool to version 0.4.346201 and launch it with the --extension argument.

```
upgrade-assistant upgrade "Project.csproj" --extension "C:\Dev\my-extension"
```

### Exceptions (if any)

```
Unhandled exception. System.ArgumentException: Type System.Collections.Generic.IReadOnlyCollection`1[System.String] cannot be created without a custom binder.
   at System.CommandLine.Binding.ArgumentConverter.CreateEnumerable(Type type, Type itemType, Int32 capacity)
   at System.CommandLine.Binding.ArgumentConverter.ConvertTokens(Argument argument, Type type, IReadOnlyList`1 tokens, LocalizationResources localizationResources, ArgumentResult argumentResult)
   at System.CommandLine.Binding.ArgumentConverter.TryConvertArgument(ArgumentResult argumentResult, Object& value)
   at System.CommandLine.Parsing.ArgumentResult.Convert(Argument argument)
   at System.CommandLine.Parsing.ArgumentResult.GetArgumentConversionResult()
   at System.CommandLine.Parsing.ParseResultVisitor.ValidateAndConvertArgumentResult(ArgumentResult argumentResult)
[...]

PR #1330, opened 2022-10-12
Fixing #1297

brandonh-msft:
System.Commandline no longer supports IReadOnlyCollection, it seems. Pivoted to ICollection<T> but still assigning to an IReadOnlyCollection which solves the parsing, but still protects the collection from modification

cc @twsouthwick @mjrousos @sunandabalu for review


lutzroeder (approved):

lutzroeder:
@mgoertz-msft

mjrousos (approved):
Would be good to get tests for this, too, to prevent future regressions.

PR #1350, opened 2022-11-03
Adding STS to the list of TFM selection

vijayrkn:

## Description
Detail 1
Detail 2


vijayrkn:
@twsouthwick @mjrousos @abpiskunov @brandonh-msft 

@OliaG helped validate the changes (locally) in this draft PR and this helped unblock her on using the new STS version by default. The change here is pretty much replacing current with STS and updating the TFM value. I haven't updated the tests to match the newer versions. I saw that the tests were not updated for NET6 as well. 

Any concerns with going with this interim approach for facilitating the NET70 release?

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IUpgradeAssistantOptions.cs
```diff
@@ -20,6 +20,6 @@ public class BaseUpgradeAssistantOptions : IUpgradeAssistantOptions
 
         public string? Format { get; set; }
 
-        public UpgradeTarget TargetTfmSupport { get; set; } = UpgradeTarget.LTS;
+        public UpgradeTarget TargetTfmSupport { get; set; } = UpgradeTarget.STS;
```
mjrousos, line 23: Not sure we want this to be the default. Particularly for upgrade scenarios, customers are more likely to need longer support than new net7 features.

src/components/Microsoft.DotNet.UpgradeAssistant/TargetFramework/DefaultTfmOptions.cs
```diff
@@ -8,22 +8,22 @@ namespace Microsoft.DotNet.UpgradeAssistant
...
+        public UpgradeTarget TargetTfmSupport { get; set; } = UpgradeTarget.STS;
 
         public string DetermineTargetTfmValue() => this.TargetTfmSupport switch
         {
-            UpgradeTarget.Current => this.Current,
+            UpgradeTarget.STS => this.STS,
             UpgradeTarget.Preview => this.Preview,
             UpgradeTarget.LTS => this.LTS,
-            _ => this.LTS,
+            _ => this.STS,
```
mjrousos, line 26: Again, I'd leave LTS as the default.

src/components/Microsoft.DotNet.UpgradeAssistant/TargetFramework/TargetFrameworkSelector.cs
```diff
@@ -38,14 +38,13 @@ public async ValueTask<TargetFrameworkMoniker> SelectTargetFrameworkAsync(IProje
 
             var appBase = _selectorOptions.TargetTfmSupport switch
             {
-                UpgradeTarget.Current => _selectorOptions.Current,
+                UpgradeTarget.STS => _selectorOptions.STS,
                 UpgradeTarget.Preview => _selectorOptions.Preview,
                 UpgradeTarget.LTS => _selectorOptions.LTS,
-                _ => _selectorOptions.LTS,
+                _ => _selectorOptions.STS,
```
mjrousos, line 44: LTS

tests/Microsoft.DotNet.UpgradeAssistant.TestHelpers/TargetFrameworkMonikerParser.cs
```diff
@@ -38,7 +38,7 @@ public static class TargetFrameworkMonikerParser
         public const string Net60_Android = "net6.0-android";
         public const string Net60_iOS = "net6.0-ios";
 
-        public const string Current = Net50;
+        public const string STS = Net50;
```
mjrousos, line 41: As long as we're in here, we probably should update the tests to match STS/LTS/Preview TFM values (or maybe just change them to totally different "test" values). Not high priority, though.

mjrousos (commented):
I think we should make sure LTS is still the default TFM if users don't specify one. Other than that, this looks good.

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IUpgradeAssistantOptions.cs
vijayrkn: @OliaG has specifically asked to change the default to the latest instead of LTS. That is the reason I updated all the defaults. 

@OliaG - Is this decision finalized?

src/components/Microsoft.DotNet.UpgradeAssistant/TargetFramework/DefaultTfmOptions.cs
vijayrkn: same reasoning as above.

src/components/Microsoft.DotNet.UpgradeAssistant/TargetFramework/TargetFrameworkSelector.cs
vijayrkn: same reasoning as above.

tests/Microsoft.DotNet.UpgradeAssistant.TestHelpers/TargetFrameworkMonikerParser.cs
vijayrkn: Agreed. I didn't want to do it in the same PR. I will take a follow-up item to fix up all the tests with the right values.

mjrousos: Sounds good. Thanks!

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IUpgradeAssistantOptions.cs
OliaG: As the whole modernization initiative, we want users on the latest released version and encourage them to "Stay current" meaning updating every November. For some workloads like MAUI the difference is very significant.

Plus, those users who want LTS are usually aware of our support policy, terminology etc.  and can specify this parameter explicitly. But by default, users would assume they are getting the latest .NET version.

Happy to chat about it more and loop in more people if you think it's needed. For now, I would keep it STS, the .NETConf is all about porting to .NET 7 :)

mjrousos: Yea, I think it makes sense for new project templates to target the latest .NET but my experience with customers *upgrading* is that they're more likely to want LTS (after we've explained the difference - as @twsouthwick points out, they usually don't know how the support policy works). We can chat more offline.

OliaG (approved):
LGTM

src/common/Microsoft.DotNet.UpgradeAssistant.Abstractions/IUpgradeAssistantOptions.cs
vijayrkn: There will be additional discussions on this to finalize the long-term plan. For now, will be going ahead with the current change.

Issue #1359, opened 2022-11-11
Upgrading MAUI netstandard2.1 library doesn't work


filipnavara:
### Describe the bug

The check in the code is too restrictive:
https://github.com/dotnet/upgrade-assistant/blob/1eaed1d7ed79d5c3af34ba72dbc7897aa8f3942a/src/extensions/maui/Microsoft.DotNet.UpgradeAssistant.Extensions.Maui/MauiComponentIdentifier.cs#L41

It only allows `netstandard2.0`, we use `netstandard2.1`.

### Further technical details

> upgrade-assistant --version
0.4.355802+b2aeae2c0e41fbfed35df6ab2e88b82a0c11be2b

filipnavara:
Workaround: Change the target framework to `netstandard2.0` before running the upgrade assistant. The TFM gets upgraded anyway.

mgoertz-msft:
Should work in current versions.

Issue #988, opened 2021-12-22
Missing file path in backupName!


shahramshahram:
https://github.com/dotnet/upgrade-assistant/blob/efb3cc30a24e699e6e4337f263a2b9478f04f618/src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.File.cs#L135

something like that 
            var fileName = Path.GetFileName(filePath);
            var backupName = $"{Path.GetFileNameWithoutExtension(fileName)}.old{Path.GetExtension(fileName)}";
            var dirPath = Path.GetDirectoryName(filePath);
            var backupPath = Path.Combine(dirPath!, backupName);
            var counter = 0;

            while (File.Exists(backupPath))
            {
                backupName = $"{Path.GetFileNameWithoutExtension(fileName)}.old.{counter++}{Path.GetExtension(fileName)}";
                backupPath = Path.Combine(dirPath!, backupName);
            }

sunandabalu:
@shahramshahram Are you seeing an error due to a missing file path ? The filepath is used to move the backup file there [here](https://github.com/dotnet/upgrade-assistant/blob/main/src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.File.cs#L155) , the filepath is provided by users at runtime.

shahramshahram:
I mean in here   
https://github.com/dotnet/upgrade-assistant/blob/622bd58fdc036ea706e4804534836361cf1db08c/src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/MSBuildProject.File.cs#L135     
the File.Exists should look on the whole path not the name of file.

PR #1403, opened 2023-01-11
Fixes for MAUI, logging, failed step repetitions, etc.

mgoertz-msft:

## Description
* Allow any version of .NetStandard to be migrated to MAUI
* Avoid unnecessary TFM merges for MAUI TFM SelectorFilter
* Always show 'dotnet --info' in case of workload install failure
* Added more information to log messages, such as project names
* Added MAUI step results to SARIF output
* Prevent endless failed step repetitions by adding Failed to IsDone
* Added failure level info to SARIF output using Note as default
* Added additional optional failure details to SARIF output
* Removed about:blank default which caused empty browser windows on click
* Fixed a few grammar and spelling issues

Addresses #1359
Addresses #988

### DIFF
--- a/src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Commands/Upgrade/ConsoleUpgrade.cs
+++ b/src/cli/Microsoft.DotNet.UpgradeAssistant.Cli/Commands/Upgrade/ConsoleUpgrade.cs
@@ -127,7 +127,7 @@ namespace Microsoft.DotNet.UpgradeAssistant.Cli
             if (!await ExecuteAndTimeCommand(context, step, command, token))
             {
                 Console.ForegroundColor = ConsoleColor.Yellow;
-                await _io.Output.WriteAsync($"Command ({command.CommandText}) did not succeed");
+                await _io.Output.WriteLineAsync($"Command ({command.CommandText}) did not succeed");
                 Console.ResetColor();
             }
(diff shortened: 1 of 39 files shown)


mgoertz-msft:
Looking at test failures...

lutzroeder (dismissed):

mgoertz-msft:
@twsouthwick Looks like you may have authored the `FailedStepsAreEnumerated` test, which is failing for me now with this change (same problem in `ConfigUpdaterSubStepTests.ApplyTests`). Could you please explain the reasoning behind this test and behavior?

In my experience a failed step gets repeated endlessly in non-interactive mode, so I needed to change this behavior to fix that. Unless the step status gets reset, due to a `CurrentProject` change in the context for example, there's really no point to reselect a failed step in `GetNextStep`, is there?

mgoertz-msft:
We discussed the failed step behavior this morning and it looks like this can be useful for interactive scenarios, so we decided to keep it as is for interactive and change it for non-interactive scenarios. I will see how/where we can best make that distinction and update the PR.

mgoertz-msft:
@dotnet/dotnet-upgrade-assistant-admin and @dotnet/dotnet-upgrade-assistant-uwp Please review your areas of this PR. Thank you!

jstedfast (approved):

abpiskunov (approved):
:shipit:

PR #1425, opened 2023-02-27
Fixed memory corruption in VisualStudioFinder

jstedfast:
VisualStudioFinder.GetLatestPath() uses COM to iterate over the install locations for Visual Studio but somehow, once the method returns to the caller (GetLatestVisualStudioPath), none of the methods can be relied on to return sane values.

Therefore, the easiest solution is to track the values we care about and just return them to our caller who can then avoid needing to re-make the same COM invocations a second time.

- [ ] You've read the [Contributor Guide](../CONTRIBUTING.md) and [Code of Conduct](../CODE-OF-CONDUCT.md).

## Description
Detail 1
Detail 2

### DIFF
--- a/src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/VisualStudioFinder.cs
+++ b/src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/VisualStudioFinder.cs
@@ -44,35 +44,34 @@ namespace Microsoft.DotNet.UpgradeAssistant.MSBuild
 
         private (string? Path, int? Version) GetLatestVisualStudioPath(string? suppliedPath)
         {
-            var latest = GetLatestPath(suppliedPath);
+            var latest = GetLatestVisualStudio(suppliedPath);
 
-            if (latest is null)
+            if (latest.InstallPath is null)
             {
                 _logger.LogWarning("Did not find a Visual Studio instance");
                 return default;
             }
 
-            var version = Version.Parse(latest.GetInstallationVersion());
-            var installation = latest.GetInstallationPath();
+            _logger.LogError("InstallationPath:{InstallPath}  Version:{Version}", latest.InstallPath, latest.Version);
 
-            if (Directory.Exists(installation))
+            if (Directory.Exists(latest.InstallPath))
             {
-                _logger.LogDebug("Using Visual Studio v{VsVersion} [{VsPath}]", version, installation);
+                _logger.LogDebug("Using Visual Studio v{VsVersion} [{VsPath}]", latest.Version, latest.InstallPath);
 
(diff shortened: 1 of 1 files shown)


src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/VisualStudioFinder.cs
```diff
@@ -35,35 +35,34 @@ public void Configure(WorkspaceOptions options)
 
         private (string? Path, int? Version) GetLatestVisualStudioPath(string? suppliedPath)
         {
-            var latest = GetLatestPath(suppliedPath);
+            var latest = GetLatestVisualStudio(suppliedPath);
 
-            if (latest is null)
+            if (latest.InstallPath is null)
```
mgoertz-msft, line 40: Seems like latest could be null so this should probably be latest?.InstallPath...

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/VisualStudioFinder.cs
```diff
@@ -35,35 +35,34 @@ public void Configure(WorkspaceOptions options)
...
             {
                 _logger.LogWarning("Did not find a Visual Studio instance");
                 return default;
             }
 
-            var version = Version.Parse(latest.GetInstallationVersion());
-            var installation = latest.GetInstallationPath();
+            _logger.LogError("InstallationPath:{InstallPath}  Version:{Version}", latest.InstallPath, latest.Version);
```
DianaSoltani, line 46: Should this be removed? There's no error here and you log this information later as well

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/VisualStudioFinder.cs
jstedfast: it cannot be null anymore - it now always returns a non-null object. Only latest.InstallPath can be null.

src/components/Microsoft.DotNet.UpgradeAssistant.MSBuild/VisualStudioFinder.cs
jstedfast: d'oh, yes. Thanks.

DianaSoltani (approved):

jstedfast:
I think this was merged as part of one of @DianaSoltani 's PRs so we can close this.

jstedfast:
@abpiskunov can I get a review on this?

mgoertz-msft (approved):

abpiskunov (approved):

PR #1576, opened 2024-06-25
Timestamp NuGet packages

jstedfast:
Several people have been bitten while testing the UpgradeAssistant with local builds of the Mappings nuget package because every build produces the same versioned package and if they forget to delete the cached version in the NuGet cache directory, then their changes will appear to not take affect because UA is using an old version of their changes.

Also modified the Nerdbank.GitVersioning version.json file to use a base of "1.0" instead of "1.0.0".

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <GenerateNuspecDependsOn>$(GenerateNuspecDependsOn);TimestampNugetPackage</GenerateNuspecDependsOn>
+
+    <!-- Enable package timestamping by default -->
+    <TimestampPackage Condition=" '$(TimestampPackage)' == ''">true</TimestampPackage>
+  </PropertyGroup>
+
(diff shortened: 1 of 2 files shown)


Directory.Build.targets
```diff
@@ -0,0 +1,17 @@
...
+    <GenerateNuspecDependsOn>$(GenerateNuspecDependsOn);TimestampNugetPackage</GenerateNuspecDependsOn>
+
+    <!-- Enable package timestamping by default -->
+    <TimestampPackage Condition=" '$(TimestampPackage)' == ''">true</TimestampPackage>
+  </PropertyGroup>
+
+  <Target Name="TimestampNugetPackage">
+    <PropertyGroup>
+      <CurrentDate>$([System.DateTime]::Now.ToString(yyyyMMdd-HHmm))</CurrentDate>
+      <PackageVersion Condition="'$(TimestampPackage)' == 'true'">$(PackageVersion)-CI-$(CurrentDate)</PackageVersion>
```
mgoertz-msft, line 13: Since this is also used for local build, do we want to remove "-CI" from the package version?

mgoertz-msft (approved):

Directory.Build.targets
jstedfast: Yea, maybe we should

PR #1709, opened 2026-04-28
Update modernize-dotnet plugin to 1.0.1063-preview1

anvillan:
Automated update of modernize-dotnet plugin from UpgradeAssistant build main-20260428.3 (version 1.0.1063-preview1).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,17 @@
 
 This changelog applies to the custom agent available in this repo for Copilot CLI and Copilot Coding Agent as well as in the Visual Studio Code extension
 
+## 1.0.1063-preview1
+
+### Added
+- User can now specify a folder or a list of projects to run the upgrade process
+
+### Changed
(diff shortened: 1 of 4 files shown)


SanyamSavla (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,17 @@
...
+## 1.0.1063-preview1
+
+### Added
+- User can now specify a folder or a list of projects to run the upgrade process
+
+### Changed
+- Relaxed branch selection instructions to allow users to choose staying on current branch
+
+### Security
+- Bumped Microsoft.Build.Tasks.Core from 16.10.0-preview to 16.11.6
```
tlmii: This is really a VS-only change (and build-time only, for that matter) if I remember correctly, so we should probably exclude it from the changelog as it doesn't have customer impact

anvillan: Removed that entry.

(Did not know that security _patch_ is only applicable for VS)

tlmii (approved):

